PR #26, opened 2022-06-04
Add OpenApi generation

toddbaert:
This PR adds generation of an OpenFeature-compliant REST API from the OpenAPI spec [here](https://github.com/open-feature/spec/pull/92). It includes models, validation, and parsing.

The Chi framework was used since it has no dependencies, and is compatible with native go `net/http` handlers. 

Run `make generate` (which is now a dependency of `make build`) after downloading the require go modules to generate the code. The generated code is .gitignored.

Then, you can perform flag evaluations:

```
curl -X POST localhost:8080/flags/some-flag/resolve/boolean?default-value=true
// {"reason":"DEFAULT","value":true}

curl -X POST localhost:8080/flags/some-flag/resolve/string?default-value=hi
// {"reason":"DEFAULT","value":"hi"}

curl -X POST localhost:8080/flags/some-flag/resolve/number?default-value=100
// {"reason":"DEFAULT","value":100}

curl -X POST localhost:8080/flags/some-flag/resolve/object?default-value=foo,bar
// {"reason":"DEFAULT","value":{"foo":"bar"}}
```

I've yet to get all the abstractions working together after this change, but I'll work on that next. 

Closes https://github.com/open-feature/flagd/issues/27


open_api_gen_config.yml
```diff
@@ -0,0 +1,5 @@
+output: ./pkg/generated/server.gen.go
+package: gen
+generate:
+  chi-server: true
+  models: true
```
toddbaert, line 5: Basic configuration for the code generation.

openapi/provider.yml
```diff
@@ -0,0 +1,234 @@
+openapi: "3.0.2"
```
toddbaert, line 1: I copied this file from the spec repo, but we might want to make it a git submodule.

pkg/service/http_service.go
```diff
@@ -1,60 +1,63 @@
...
 
-type HttpServiceResponse struct {
-	rawPayload string
-}
+var defaultReason = "DEFAULT"
 
-func (h *HttpServiceResponse) GetPayload() string {
-	return h.rawPayload
+// implement the generated ServerInterface.
+type Server struct {}
```
toddbaert, line 23: The code generator defines an interface that we must implement to handle our request. It also generates middlewares that do the parsing and validation, which is in the generated code that should not be manually modified (`pk/generated/server.gen.go`

pkg/service/http_service.go
```diff
@@ -1,60 +1,63 @@
...
+
+func (s Server) ResolveBoolean(w http.ResponseWriter, r *http.Request, flagKey gen.FlagKey, params gen.ResolveBooleanParams) {
+	json.NewEncoder(w).Encode(gen.ResolutionDetailsBoolean{
+		Value: &params.DefaultValue,
```
toddbaert, line 27: All the functions are currently stubbed and simply return the default.

Issue #23, opened 2022-06-03
OpenAPI schema


beeme1mr:
Relates to: https://github.com/open-feature/spec/pull/92

PR #28, opened 2022-06-08
Feat/rest api and file provider

toddbaert:
This PR adds:

- generation of an OpenFeature-compliant REST API from a OpenAPI spec including models, validation, and parsing.
- a "evaluator" concept and package, with an `IEvaluator` interface. This is the "brain" doing flag evaluation. It knows hot to take a string from the `ISync`'s `Fetch` method, and convert it to a particular schema, and then perform flag evaluations. The only implementation so far is the `JsonEvaluator`, which uses the schema [here](https://github.com/open-feature/schemas/blob/main/json-schema/flagd-definitions.json) (included as a submodule)

The Chi framework was used since it has no dependencies, and is compatible with native go `net/http` handlers. 

To test:

- pull in the schema submodule with `git submodule update --init --recursive`
- `go mod download`
- run `make build` (which is now includes of `make generate`). The generated code is .gitignored.

Then, you can perform flag evaluations after running `./flagd start  --uri ./examples/provider_flags.json`

```
curl -X POST localhost:8080/flags/myBoolTest/resolve/boolean?default-value=true
// {"reason":"DEFAULT","value":true}

curl -X POST localhost:8080/flags/myStringTest/resolve/string?default-value=hi
// {"reason":"DEFAULT","value":"red"}

curl -X POST localhost:8080/flags/myNumericTest/resolve/number?default-value=123
// {"reason":"DEFAULT","value":1}

curl -X POST localhost:8080/flags/myObjectTest/resolve/object?default-value=foo,bar
// {"reason":"DEFAULT","value":{"color":"blue"}}
```
Closes: https://github.com/open-feature/flagd/issues/23
Closes: https://github.com/open-feature/flagd/issues/27


AlexsJones:
Hey great stuff, Can we talk this one through on a zoom call @toddbaert ?

Makefile
```diff
@@ -1,9 +1,11 @@
 IMG=flagd:latest
 PHONY: .docker-build .build .run
 PREFIX=/usr/local
+generate:
```
toddbaert, line 4: new step generate is prereq for build

pkg/eval/json_evaluator.go
```diff
@@ -0,0 +1,50 @@
+package eval
```
toddbaert, line 1: The only current IEvaluator implementation. It maintains the flag state and knows how to intelligently parse it to evaluate flags as per OpenFeature spec.

pkg/runtime/runtime.go
```diff
@@ -66,12 +62,5 @@ func Start(syncr sync.ISync, server service.IService, ctx context.Context) {
...
 
-	go server.Serve(
-		func(ir service.IServiceRequest) service.IServiceResponse {
-			if ir.GetRequestType() == service.SERVICE_REQUEST_ALL_FLAGS {
-				return ir.GenerateServiceResponse(getSyncPayload())
-			}
-			return nil
-		})
-
+	go server.Serve(ev)
```
toddbaert, line 65: Server has handlers of some shape, which are run with the `.Serve()` method, which takes the `IEvaluation` for the handlers to call.

pkg/sync/isync.go
```diff
@@ -1,13 +1,9 @@
...
-	SYNC_STRATEGY_MERGE   = 1
-)
-
+/*
+ISync implementations watch for changes in the flag source (HTTP backend, local file, s3 bucket), and fetch the latest values.
+*/
 type ISync interface {
 	Fetch() (string, error)
-	Notify(chan INotify)
+	Notify(chan<- INotify)
```
toddbaert, line 8: Restricted this channel direction.

pkg/eval/ievaluator.go
```diff
@@ -0,0 +1,14 @@
+package eval
+
+/*
+IEvaluator implementations store the state of the flags, do parsing and validation of the flag state and evaluate flags in response to handlers.
+*/
+type IEvaluator interface {
+	// TODO: add context param when rule evaluator is implemented
+	GetState() (string, error)
+	SetState(state string) error
```
toddbaert, line 9: With this abstraction, it's actually possible to add IService impls that don't have well-defined transports as well... you could simply use GetState() to return the entire flag data, and only call that in whatever handler you implement.

Makefile
```diff
@@ -1,9 +1,11 @@
...
+	${GOPATH}/bin/oapi-codegen --config=./config/open_api_gen_config.yml ./schemas/openapi/provider.yml
 docker-build:
 	docker buildx build --platform="linux/ppc64le,linux/s390x,linux/amd64,linux/arm64" -t ${IMG} .
-build:
+build: generate
 	go build -o flagd
 run:
```
AlexsJones, line 10: We might also need to add generate here too?

cmd/start.go
```diff
@@ -118,6 +139,7 @@ func init() {
 	startCmd.Flags().StringVarP(&socketServicePath, "socketpath", "d", "/tmp/flagd.sock", "flagd socket path")
 	startCmd.Flags().StringVarP(&serviceProvider, "service-provider", "s", "http", "Set a serve provider e.g. http or socket")
 	startCmd.Flags().StringVarP(&syncProvider, "sync-provider", "y", "filepath", "Set a sync provider e.g. filepath or remote")
+	startCmd.Flags().StringVarP(&evaluator, "evaluator", "e", "json", "Set and evaluator e.g. json")
```
AlexsJones, line 142:
```suggestion
startCmd.Flags().StringVarP(&evaluator, "evaluator", "e", "json", "Set an evaluator e.g. json")
```

AlexsJones (changes requested):

Makefile
```diff
@@ -1,9 +1,11 @@
 IMG=flagd:latest
 PHONY: .docker-build .build .run
 PREFIX=/usr/local
+generate:
+	${GOPATH}/bin/oapi-codegen --config=./config/open_api_gen_config.yml ./schemas/openapi/provider.yml
 docker-build:
```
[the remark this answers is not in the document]
toddbaert: :100:

pkg/runtime/runtime.go
```diff
@@ -5,37 +5,8 @@ import (
...
-		case <-requestBuffer:
-			data, err := syncr.Fetch()
-			if err != nil {
-				log.Warn(err.Error())
-			}
-			messageBuffer <- data
-		}
-	}
+func Start(server service.IService, provider sync.ISync, ctx context.Context) {
+	go server.Serve(provider)
```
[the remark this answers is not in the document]
toddbaert: Good catch. Go newbie issues.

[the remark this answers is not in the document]
toddbaert: It seems like you commented on a sale commit here, possibly from the first PR before I rebased? I have added a context arg to resolve this comment, but you may want to take a look at this file again to make sure you saw the latest.

PR #30, opened 2022-06-08
Feature/watchers

AlexsJones:
Adds simple watcher implementation for filesystem and remote.

I will remove the debug_bin in the squash


examples/example_flags.json
```diff
@@ -1,6 +1,6 @@
 {
     "newWelcomeMessage": {
-      "state": "disabled"
+      "state": "disable"
```
thschue, line 3: I would keep "disabled" personally ...

thschue (approved):
lgtm

examples/example_flags.json
AlexsJones: Fixed

PR #40, opened 2022-06-17
Update to latest schema, add errors/reasons

toddbaert:
This PR:

- updates to the latest flag definition schema version without the ugly type categories
- updates the `Flags/Flag` model to be an `interface{}` since flags can have any value (if conversion/type assertion fails, we indicate a type error which is a well-defined concept in OpenFeature providers) 
- shifts evaluation reason/error logic into eval package. The HTTP layer now really only concerns itself with HTTP error codes, not the `reasons`/`errorCode` fields which are general flagd semantics


beeme1mr (approved):

AlexsJones:
I think we should chat about those vars next time @toddbaert

toddbaert:
> I think we should chat about those vars next time @toddbaert

Yep, apologies... I'd have liked to have a bit more discussion on this one before merge, but we really wanted to get a more robust demo working that had an actual web UI (based on a conversation earlier with @thschue and @beeme1mr ). I normally wouldn't be so quick with the merge button :sweat_smile: ...

I think your const changes of course make sense, thanks! If you have concerns about the values conceptually, we can talk more about them... Nothing here set in stone.

Issue #42, opened 2022-06-20
Remove FSNotify Events Stop File Watcher


beeme1mr:
When using a mounted ConfigMap, Kubernetes removes and readds the file. This causes Flagd to disconnect and no longer watch the file for changes.

Here is the output from Flagd after a ConfigMap update:

```
time="2022-06-20T14:48:50Z" level=info msg="Starting filepath sync notifier"
time="2022-06-20T14:48:50Z" level=info msg="Notifying filepath: /etc/flagd/config.json"
time="2022-06-20T14:49:57Z" level=info msg="Filepath notifier event: /etc/flagd/config.json CHMOD"
time="2022-06-20T14:49:57Z" level=info msg="Filepath notifier event sent"
time="2022-06-20T14:49:57Z" level=info msg="Filepath notifier event: /etc/flagd/config.json REMOVE"
time="2022-06-20T14:49:57Z" level=info msg="New configuration created"
time="2022-06-20T14:49:57Z" level=info msg="Configuration deleted"
time="2022-06-20T14:49:57Z" level=info msg="Filepath notifier event sent"
```

toddbaert:
Confirmed. I saw this ealier.

This is because K8s actually exposes configmaps as symlinks in the volume. The symlink update is what causes the remove events.

PR #43, opened 2022-06-20
Restore watcher on remove

toddbaert:
Closes: https://github.com/open-feature/flagd/issues/42

@AlexsJones I found this issue which seems to have to do with the way K8s exposes configmaps when treated as volumes. The file is actually a symlink which is moved when the configmap is changed. This fires the remove event, which removes our watcher.

The solution seems to be to simply add the watcher back. This resolves the issue in my local cluster. @beeme1mr will confirm on his GKE cluster as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-IMG=flagd:latest
+IMG ?= flagd:latest
 PHONY: .docker-build .build .run
 PREFIX=/usr/local
 guard-%:
@@ -13,6 +13,8 @@ generate: guard-GOPATH
 	${GOPATH}/bin/oapi-codegen --config=./config/open_api_gen_config.yml ./schemas/openapi/provider.yml
 docker-build: generate
 	docker buildx build --platform="linux/ppc64le,linux/s390x,linux/amd64,linux/arm64" -t ${IMG} .
+docker-push: generate
+	docker buildx build --push --platform="linux/ppc64le,linux/s390x,linux/amd64,linux/arm64" -t ${IMG} .
 build: generate
 	go build -o flagd
 run: generate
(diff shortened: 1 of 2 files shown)


pkg/sync/filepath_sync.go
```diff
@@ -50,6 +50,13 @@ func (fs *FilePathSync) Notify(w chan<- INotify) {
 				case fsnotify.Write:
 					evtType = E_EVENT_TYPE_MODIFY
 				case fsnotify.Remove:
+					// K8s exposes config maps as symlinks.
+					// Updates cause a remove event, we need to re-add the watcher in this case.
+					err = watcher.Add(fs.URI)
+					if (err != nil) {
+						log.Println("Error restoring watcher:", err)
+						log.Fatal(err)
```
toddbaert, line 58: I made this `Fatal`... I think that's the right choice here, since swallowing this error will probably mean nothing is ever updated again.

Makefile
```diff
@@ -1,4 +1,4 @@
-IMG=flagd:latest
+IMG ?= flagd:latest
```
AlexsJones, line 1: I don't think this is necessary as it's outside of the target and can be overridden. Is there an issue with this?

pkg/sync/filepath_sync.go
```diff
@@ -50,6 +50,13 @@ func (fs *FilePathSync) Notify(w chan<- INotify) {
 				case fsnotify.Write:
 					evtType = E_EVENT_TYPE_MODIFY
 				case fsnotify.Remove:
+					// K8s exposes config maps as symlinks.
+					// Updates cause a remove event, we need to re-add the watcher in this case.
+					err = watcher.Add(fs.URI)
+					if (err != nil) {
+						log.Println("Error restoring watcher:", err)
```
AlexsJones, line 57: We don't need two error messages after each other

pkg/sync/filepath_sync.go
AlexsJones: Sounds about right yeah

AlexsJones (commented):
Have a few open questions

PR #46, opened 2022-06-20
Fix CI in PRs from forks, scan local image

toddbaert:
This PR does trivy scanning with a local image instead of from a pushed image.

It also removes the scanning in main, because if this is merged nothing that hasn't been scanned should get that far.

### DIFF
--- a/.github/workflows/build-push-oci.yaml
+++ b/.github/workflows/build-push-oci.yaml
@@ -1,10 +1,7 @@
-name: build-oci
+name: build-push-oci
 
 on:
   push:
-    branches:
-      - main
-  pull_request:
     branches:
       - main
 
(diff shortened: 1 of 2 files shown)


.github/workflows/local-build-oci.yaml
```diff
@@ -0,0 +1,47 @@
...
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          outputs: type=docker,dest=${{ github.workspace }}/flagd-local.tar
+          tags: flagd-local:test
+
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@49e970d7ac7539dfcc6ddba87b22ddd9b7377540
+        with:
+          # This doesn't work, but should. Looks like a recent bug: https://github.com/aquasecurity/trivy-action/issues/120
```
toddbaert, line 37: Very unfortunate

AlexsJones (approved):

PR #48, opened 2022-06-21
Add json-logic rule evaluation

toddbaert:
This PR adds rule evaluation (aka targeting) via [json logic](https://jsonlogic.com/), as well as related tests:

Basically:

- the `json_evaluator` houses this logic. It remains the only component aware of the json-schema and how to parse it
- a targeting rule must return a defined variant. If it doesn't, the default is used

Still do do:
- on startup/file-change, validate rules using `jsonlogic.IsValid()` on each flag that has targeting defined.
- consider upgrading to golang 1.18 to remove some duplication via generics

To test, call the newest demo flag, which has a rule, and supply a context payload in the body:

new rule:

```
"isColorYellow": {
      "state": "ENABLED",
      "variants": {
        "on": true,
        "off": false
      },
      "defaultVariant": "off",
      "targeting": {
        "if": [
          {
            "==": [
              {
                "var": [
                  "color"
                ]
              },
              "yellow"
            ]
          },
          "on",
          "off"
        ]
      }
    }
```

Example call:

```
$ curl -X POST localhost:8080/flags/isColorYellow/resolve/boolean?default-value=true \
-d '{"color": "yellow"}'
// {"reason":"TARGETING_MATCH","value":true}
```

Closes: https://github.com/open-feature/flagd/issues/49


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "schemas"]
 	path = schemas
 	url = https://github.com/open-feature/schemas
+	branch = feat/rules
```
toddbaert, line 4: Temporarily using this branch.

pkg/eval/json_evaluator.model.go
```diff
@@ -1,12 +1,17 @@
...
 
 // we could make this more type-safe with generics when we upgrade to 1.18.
 type Flag struct {
-	State string `json:"state"`
-	DefaultVariant string	`json:"defaultVariant"`
-	Variants map[string]interface{} `json:"variants"`
+	State 					string									`json:"state"`
+	DefaultVariant	string									`json:"defaultVariant"`
+	Variants 				map[string]interface{}	`json:"variants"`
+	Rules 					json.RawMessage					`json:"rules"`
```
toddbaert, line 16: Used `json.RawMessage` as the type here, since we are just going to convert it to bytes again for use in json-logic evaluation anyway.

pkg/service/http_service.go
```diff
@@ -26,17 +27,14 @@ type Server struct {
 
 // implement the generated ServerInterface.
 // TODO: might be able to simplify some of this with generics.
-// TODO: add improved, more RESTful error handling, we should inspect the returned ErrorCode and respond with a matching HTTP status code. 
 func (s Server) ResolveBoolean(w http.ResponseWriter, r *http.Request, flagKey gen.FlagKey, params gen.ResolveBooleanParams) {
-	result, reason, err := s.eval.ResolveBooleanValue(flagKey, params.DefaultValue)
+
+	var contextObj gen.Context
+	json.NewDecoder(r.Body).Decode(&contextObj)
```
AlexsJones, line 33: Might want to put some guards around this

toddbaert:
@thomaspoignant I should add I'm not married to this implementation. I'm open to alternatives. I was kinda excited about getting this done in less than a day... I was going to do something basic and ended up getting a entire engine working fairly quickly.

toddbaert:
If https://github.com/open-feature/flagd/pull/51 is merged, I will rebase on it and add the testing make file phase to the workflow.

toddbaert:
Rebased to absorb new lint rules. Thanks again @thomaspoignant

Makefile
```diff
@@ -7,7 +7,7 @@ guard-%:
         exit 1; \
     fi
 generate: guard-GOPATH
-	git submodule update --init --recursive
+	git submodule update --init --recursive --remote
```
toddbaert, line 10: This just allows us to use branch refs, which is especially nice for PRs.

.github/workflows/code-checks.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v1
         with:
-          go-version: '^1.16.0'
+          go-version: "^1.17.0"
       - run: make generate
       - run: make lint
+      - run: make test
```
thomaspoignant, line 25: I would suggest having 2 jobs instead of one.
It will allow `test` and `lint` to run at the same time and it will allow a quick feedback loop.

Something like this:
```yaml
  Lint:
    runs-on: ubuntu-latest
    env:
      GOPATH: /home/runner/work/open-feature/flagd
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2
      - name: Setup go
        uses: actions/setup-go@v1
        with:
          go-version: "^1.17.0"
      - run: make generate
      - run: make lint

  Test:
    runs-on: ubuntu-latest
    env:
      GOPATH: /home/runner/work/open-feature/flagd
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2
      - name: Setup go
        uses: actions/setup-go@v1
        with:
          go-version: "^1.17.0"
      - run: make generate
      - run: make test
```

pkg/eval/json_evaluator_test.go
```diff
@@ -0,0 +1,342 @@
+package eval
```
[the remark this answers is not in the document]
thomaspoignant: I am a heavy user of `github.com/stretchr/testify/assert` so makes totally sense to me 👍

.github/workflows/code-checks.yaml
toddbaert: Great idea!

pkg/eval/json_evaluator_test.go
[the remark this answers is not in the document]
thomaspoignant: I did not go into details in your tests but in **golang** there is heavy use of **table-driven style** to write tests.

You can check this example https://gobyexample.com/testing to see what I mean.
It makes tests way more readable since you have one function only for everything you want to test.

You can check this [file](https://github.com/thomaspoignant/go-feature-flag/blob/main/internal/flagv1/rollout_test.go) in one of my repo to see how it look like.

I will go more into what you are doing inside the tests later.

thomaspoignant, line 1:
```suggestion
package eval_test
```

A good practice is to have a different package for test, it forces you to test only your public function.

.github/workflows/code-checks.yaml
```diff
@@ -19,6 +19,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: /home/runner/work/open-feature/flagd
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup go
+        uses: actions/setup-go@v1
+        with:
+          go-version: "^1.17.0"
```
AlexsJones, line 36:
```suggestion
          go-version: "^1.17.11"
```

Makefile
AlexsJones, line 10: > Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch. 

This is a very dangerous change to add and breaks immutability of the build process @toddbaert

.github/workflows/code-checks.yaml
```diff
@@ -19,6 +19,20 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@v1
         with:
-          go-version: '^1.16.0'
+          go-version: "^1.17.0"
       - run: make generate
       - run: make lint
+
+  Test:
```
AlexsJones, line 26:
```suggestion
  test:
```

pkg/eval/ievaluator.go
```diff
@@ -1,15 +1,19 @@
...
 */
 type IEvaluator interface {
-	// TODO: add context param when rule evaluator is implemented
 	GetState() (string, error)
 	SetState(state string) error
-	ResolveBooleanValue(flagKey string, defaultValue bool) (bool, string, error)
-	ResolveStringValue(flagKey string, defaultValue string) (string, string, error)
-	ResolveNumberValue(flagKey string, defaultValue float32) (float32, string, error)
-	ResolveObjectValue(flagKey string, defaultValue map[string]interface{}) (map[string]interface{}, string, error)
+	ResolveBooleanValue(flagKey string, defaultValue bool, context gen.Context) (value bool, reason string, err error)
```
AlexsJones, line 12: More of a general observation, but the error message can contain a reason inside of it, what would the value be of having a separate error to a reason?

AlexsJones (changes requested):

Makefile
toddbaert: Yes, this makes sense. I didn't think about the nuance of this change. Thank you, I will revert this.

pkg/eval/ievaluator.go
toddbaert: There's a semantic difference between a reason and an error within the OpenFeature spec. The reason denotes why a flag evaluated the way it did, even in cases of normal execution. For example, a reason could denote that specific user targeting occurred, or a random split occurred, or a static value was used, etc. In abnormal execution, the reason should denote an error occurred, but the details of that error are carried by the error code itself.

https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#requirement-146

pkg/eval/json_evaluator_test.go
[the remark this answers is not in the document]
toddbaert: @thomaspoignant I was wrong. I made the tests for the resolver methods tabular, and I like it much better. It also increased test coverage.

AlexsJones (approved):

PR #51, opened 2022-06-22
Add linters + pre-commit

thomaspoignant:
This pull request adds the support of `golangci-lint` which is a pretty standard linter for golang.
It will help to have a uniform way of writing go code in the repository.

I have also fixed all the issue that were raised by the linters.
The only one I have ignore is the one related to the hashing solution because I guess you have used `sha1` for a reason.

`golangci-lint` will also be run with github action.

There is also a `pre-commit` configuration to be used with https://pre-commit.com/ that will run `golangci-lint` and `gofumpt` at each commit.
Note that you have to install the pre-commit hook with the command `pre-commit install` to have this feature.

### DIFF
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -1,4 +1,4 @@
-name: ci
+name: build-oci
 
 on:
   push:
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -13,19 +13,6 @@ env:
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
-  Lint:
-    runs-on: ubuntu-latest
-    env:
-      GOPATH: /home/runner/work/open-feature/flagd
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-      - name: Setup go
-        uses: actions/setup-go@v1
-        with:
-          go-version: '^1.16.0'
-      - run: make generate
-      - run: make lint
   docker:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -7,12 +7,25 @@ on:
   pull_request:
     branches:
     - main
-   
+
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
 
 jobs:
+  Lint:
+    runs-on: ubuntu-latest
+    env:
+      GOPATH: /home/runner/work/open-feature/flagd
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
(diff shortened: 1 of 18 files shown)


toddbaert (commented):
Thank you! this is a good idea. As far as the sha1, we're just using it in the http sync to ignore no-op changes (body has same sha1 => ignore). Please correct me if I'm wrong @AlexsJones . Anyway, because that's not security related, I think it's fine, though upgrading to a better hash alg probably wouldn't have much of a performance impact either in this case.

Anyway, the rest looks good to me, seems like mostly casing/spacing changes. I would like @AlexsJones 's approval before we merge.

.github/workflows/build-oci.yaml
```diff
@@ -7,12 +7,25 @@ on:
...
+      GOPATH: /home/runner/work/open-feature/flagd
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup go
+        uses: actions/setup-go@v1
+        with:
+          go-version: '^1.16.0'
+      - run: make generate
+      - run: make lint
```
toddbaert, line 28: I think I would rather see this in a new workflow file. In my other PR, I've added a testing phase too, so I will include it here as well.

I think we should keep this file mostly about building the image.

thomaspoignant: I was confused by the name of the workflow `ci` that's why I have put it in the same file.
I will rework this PR to create a new workflow or to add in the testing one tomorrow.

toddbaert: > I was confused by the name of the workflow ci that's why I have put it in the same file.

I think it's "oci" as in "OCI Image Format"

> I will rework this PR to create a new workflow or to add in the testing one tomorrow.

:+1: Yep, if you just add the lint stage, I will add testing with my PR after rebasing on yours.

thomaspoignant: > I think it's "oci" as in "OCI Image Format"

What confused me was this line, maybe we should change this to a different name at some point.
https://github.com/open-feature/flagd/blob/317c60a84c2e9168e08442948f75ff849f6eb829/.github/workflows/build-oci.yaml#L1

toddbaert:
@thomaspoignant can you rebase on main? Or merge into this? That should fix the build.

thomaspoignant:
> @thomaspoignant can you rebase on main? Or merge into this? That should fix the build.

Done ✅ CI all green 🙌

toddbaert (approved):
@AlexsJones said he'd be a bit scarce this week, and we can always change lint rules, so I'm going to merge this!

.github/workflows/code-checks.yaml
```diff
@@ -0,0 +1,24 @@
+name: code-checks
```
toddbaert, line 1: I like this much better.

Issue #52, opened 2022-06-23
Build push is failing


AlexsJones:
```
error: failed to solve: unexpected status: 403 Forbidden
Error: buildx failed with: error: failed to solve: unexpected status: 403 Forbidden
```

I am out some of this week, but could you take a look at this before we approve any more PR's?

toddbaert:
I mentioned this in the channel a couple times, but an issue is a good idea.

This is because of the changes made [here](https://github.com/open-feature/flagd/commit/ccbc8708708df5b904e778b5dcd1b44107dee96f) to push images for tivy scanning. The $GITHUB_TOKEN can't be used to push images in PRs from forks, hence the 403.

Merging works fine from origin branches.

I've been experimenting with a workaround, but it's blocked by a bug in the trivy action: https://github.com/open-feature/flagd/pull/46

I don't want to stop all PR approvals because of this issue. Perhaps we can just not use forks for now. The other option would be to disable trivy scanning on PRs from forks. I can easily do that.

toddbaert:
Because this is a known issue in trivy, for now I will disable scanning in PRs from forks. I will recommend people create PRs from the origin. I have a PR to scan a local trivy image, and as soon as the trivy issue is fixed, we can use that to scan from PRs as well.

PR #53, opened 2022-06-23
Disable pushing in forks

toddbaert:
This is a temporary stop-gap to not run the trivy scan in PRs from forks, for now, since attempting a push from a fork causes a 403... Once the issue with the trivy action described [here](https://github.com/open-feature/flagd/pull/46) is fixed, we can change the build to do tarball scanning instead of scanning published images.

Closes: https://github.com/open-feature/flagd/issues/52

### DIFF
--- a/.github/workflows/build-oci.yaml
+++ b/.github/workflows/build-oci.yaml
@@ -2,12 +2,12 @@ name: ci
 
 on:
   push:
-   branches:
-   - main
+    branches:
+      - main
   pull_request:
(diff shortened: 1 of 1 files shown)


AlexsJones (approved):
I spoke to aqua and escalated this issue

toddbaert:
> I spoke to aqua and escalated this issue

Thanks! I think this wasn't their fault, hopefully a solution comes soon.

PR #55, opened 2022-06-24
go install linter

toddbaert:
Just a small change here to install the linter bin via `go install` to the `GOPATH` instead of piping curl to the shell, which always makes me nervous.


thomaspoignant (approved):
It looks even better yes 👍

PR #62, opened 2022-07-06
simplified pipelines

AlexsJones:
Our pipelines are doing two things 
- becoming hard to read
- pushing on main builds, we should be moving away from this and pinning to a release OCI in the OFO


thschue (approved):
lgtm

thomaspoignant:
In this comment, https://github.com/open-feature/flagd/pull/51#discussion_r904154880 @toddbaert mentioned that he would prefer to have a new workflow for test/lint.

Are we all agree that this should be in the same file ?

AlexsJones:
I didn't see the comment @thomaspoignant happy to revert if we have sound reasoning for distributing across files

toddbaert:
@AlexsJones @thomaspoignant I'm fine with this how it is. I just wanted to keep the scaning / testing / linting separate from any pushing, which is no issue anymore since we are scanning a local image.

PR #71, opened 2022-07-12
JSON merge function for multiple files 

### LINKED ISSUE #25: [Feature] JSON merge function for multiple files
In order to combine multiple feature flag json files ( in the case of aggregation etc )
Then we should have a function to merge multiple json files....

e.g. 

```
MergeJSON(jsonPayload [][]byte) ([]byte, error)
```

moshebe:
Close #25

### DIFF
--- a/pkg/eval/json_evaluator_model.go
+++ b/pkg/eval/json_evaluator_model.go
@@ -10,10 +10,10 @@ type Flags struct {
 
 func (f Flags) Merge(ff Flags) Flags {
 	result := Flags{Flags: make(map[string]Flag)}
-	for k, v := range ff.Flags {
+	for k, v := range f.Flags {
 		result.Flags[k] = v
 	}
-	for k, v := range f.Flags {
+	for k, v := range ff.Flags {
 		result.Flags[k] = v
 	}
 	return result
--- a/pkg/eval/json_evaluator_model.go
+++ b/pkg/eval/json_evaluator_model.go
@@ -8,6 +8,17 @@ type Flags struct {
 	Flags map[string]Flag `json:"flags"`
 }
 
+func (f Flags) Merge(ff Flags) Flags {
+	result := Flags{Flags: make(map[string]Flag)}
+	for k, v := range ff.Flags {
+		result.Flags[k] = v
+	}
+	for k, v := range f.Flags {
+		result.Flags[k] = v
+	}
+	return result
+}
+
 // we could make this more type-safe with generics when we upgrade to 1.18.
 type Flag struct {
 	State          string                 `json:"state"`
--- a/pkg/eval/json_evaluator_test.go
+++ b/pkg/eval/json_evaluator_test.go
@@ -462,7 +462,7 @@ func TestMergeFlags(t *testing.T) {
 			}},
 		},
 		{
-			name: "ignore override",
+			name: "override",
 			current: eval.Flags{Flags: map[string]eval.Flag{
 				"waka": {DefaultVariant: "off"},
 			}},
@@ -471,7 +471,7 @@ func TestMergeFlags(t *testing.T) {
 				"paka": {DefaultVariant: "on"},
 			}},
 			want: eval.Flags{Flags: map[string]eval.Flag{
-				"waka": {DefaultVariant: "off"},
+				"waka": {DefaultVariant: "on"},
 				"paka": {DefaultVariant: "on"},
 			}},
 		},
--- a/pkg/eval/json_evaluator_test.go
+++ b/pkg/eval/json_evaluator_test.go
@@ -416,68 +416,86 @@ func TestResolveObjectValue(t *testing.T) {
 	}
 }
 
-func TestMergeJSON(t *testing.T) {
+func TestMergeFlags(t *testing.T) {
+	t.Parallel()
 	tests := []struct {
-		name  string
-		left  string
-		right string
-		want  string
+		name    string
+		current eval.Flags
+		new     eval.Flags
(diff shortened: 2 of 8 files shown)


pkg/eval/json_evaluator_model.go
```diff
@@ -8,6 +8,17 @@ type Flags struct {
 	Flags map[string]Flag `json:"flags"`
 }
 
+func (f Flags) Merge(ff Flags) Flags {
```
James-Milligan, line 11: might be worth mentioning this behaviour in the README / -f flag definition.
I believe the only reference to this being first come first serve is in the original issue.

beeme1mr: Agreed, good catch @James-Milligan . First come, first server makes the sense to me but that behavior should be documented. Thanks for the PR @moshebe!

toddbaert: Additionally, could you add an example of starting flagd with multiple filepath syncs?

pkg/eval/json_evaluator_model.go
```diff
@@ -8,6 +8,17 @@ type Flags struct {
...
+	result := Flags{Flags: make(map[string]Flag)}
+	for k, v := range ff.Flags {
+		result.Flags[k] = v
+	}
+	for k, v := range f.Flags {
+		result.Flags[k] = v
+	}
+	return result
```
toddbaert, line 19: I think it would be nice to have a logged warning that a flag has been overwritten when it happens, what do you folks think?

pkg/eval/json_evaluator_model.go
moshebe: Thanks for the feedback @James-Milligan, @beeme1mr and @toddbaert.
Added an example to the README and describing the option within the start command (`--help`)

Please review the texts 🙏

pkg/eval/json_evaluator_test.go
```diff
@@ -414,3 +415,87 @@ func TestResolveObjectValue(t *testing.T) {
...
+			new:     eval.Flags{Flags: nil},
+			want:    eval.Flags{Flags: map[string]eval.Flag{}},
+		},
+		{
+			name:    "both empty flags",
+			current: eval.Flags{Flags: map[string]eval.Flag{}},
+			new:     eval.Flags{Flags: map[string]eval.Flag{}},
+			want:    eval.Flags{Flags: map[string]eval.Flag{}},
+		},
+
```
toddbaert, line 439:
```suggestion
```

pkg/eval/json_evaluator_test.go
```diff
@@ -414,3 +415,87 @@ func TestResolveObjectValue(t *testing.T) {
...
+			}},
+			new: eval.Flags{Flags: map[string]eval.Flag{
+				"current": {DefaultVariant: "on"},
+				"new":     {DefaultVariant: "on"},
+			}},
+			want: eval.Flags{Flags: map[string]eval.Flag{
+				"current": {DefaultVariant: "off"},
+				"new":     {DefaultVariant: "on"},
+			}},
+		},
```
toddbaert, line 478: Am I misunderstanding, or is this the opposite of what we decided?

Based on the names, it seems like the newest flag does not win, the oldest does.

moshebe: I agree @James-Milligan, will rename it.

moshebe: Done.

toddbaert: @moshebe I'm sorry to trouble you more, this is a great PR, the code and tests look excellent. I think we probably should do "last wins", instead of "first wins", like we seemed to agree upon in the [issue](https://github.com/open-feature/flagd/issues/25#issuecomment-1181941174). I think this will be especially important in the K8s CRD use-case I give [here](https://github.com/open-feature/flagd/issues/25#issuecomment-1181965242).

I think listing flags with increasing specificity will be the intuitive pattern, and in that case having the last win also seems intuitive:

```
annotations:
        openfeature.dev: "enabled"
        openfeature.dev/featureflagconfiguration: ["common-flags", "specific-flags"]   # we want "specific-flags" to win here.
```

Please let me know if you think first wins is better for a specific reason!

pkg/eval/json_evaluator_test.go
```diff
@@ -414,3 +415,86 @@ func TestResolveObjectValue(t *testing.T) {
...
+				"waka": {DefaultVariant: "off"},
+			}},
+			new: eval.Flags{Flags: map[string]eval.Flag{
+				"waka": {DefaultVariant: "on"},
+				"paka": {DefaultVariant: "on"},
+			}},
+			want: eval.Flags{Flags: map[string]eval.Flag{
+				"waka": {DefaultVariant: "on"},
+				"paka": {DefaultVariant: "on"},
+			}},
```
toddbaert, line 476: Seems better to me, thank you!

pkg/eval/json_evaluator_test.go
moshebe: @toddbaert nothing specific with this methods, just wanted to set a deterministic behavior.
Thanks for your feedback, glad to contribute!

Made the changes you suggested and updated the README and flag help description.
Please let me know what do you think.

toddbaert:
Merging.

@moshebe if you are interested in contributing more, I can invite you to the org. You may also want to add yourself to the [interested parties](https://github.com/open-feature/community/blob/main/interested-parties.md)

moshebe:
> 

Cool @toddbaert it would be great

Issue #69, opened 2022-07-11
Remove required query-param "default-value"


toddbaert:
There's no need to pass this value. The provider calling flagd already has the default, and can default the value whenever appropriate. It only over-complicates the API and the validation.

beeme1mr:
Please make sure to update the readme when the default param is no longer required.

Issue #29, opened 2022-06-08
[Feature] Create specific loggers for sync/services


AlexsJones:

toddbaert:
@AlexsJones can you elaborate a little bit on the intent of this issue?

AlexsJones:
This is to pass a sub logger from logrus per service, it's a convenience that prefixes log entries

beeme1mr:
@james-milligan can this issue be closed?

james-milligan:
Not as of yet, currently I believe sync/services are using the default log from logrus

AlexsJones:
I can look at this or someone else, basically we need to branch out sub loggers into the components, it's not a huge deal of work just a bit of plumbing

AlexsJones:
https://github.com/open-feature/flagd/pull/92

PR #73, opened 2022-07-13
http service refactor to use grpc-gateway + gRPC service

james-milligan:
The http service is now based on the following model:
Screenshot 2022-07-13 at 16 47 55

This allows for a single source of truth or both the rest and grpc interfaces. Both services are available through the same port and are routed accordingly, however it is likely that further development will see more features added to the grpc service and therefore the addition of grpc_service.go.

These changes also remove the default_value parameter:
https://github.com/open-feature/flagd/issues/69

The variant field of the response is now also populated.


pkg/service/http_service.go
```diff
@@ -19,153 +26,169 @@ type HTTPServiceConfiguration struct {
...
-	if err = json.NewEncoder(w).Encode(gen.ResolutionDetailsWithError{
-		ErrorCode: &message,
-		Reason:    &reason,
-	}); err != nil {
-		log.Errorf("Error encoding response: %s", err.Error())
-	}
+	return status.Error(statusCode, message)
+}
+
+// TODO: could be replaced with a logging client
```
beeme1mr, line 191: Is there a issue for this? If not, please create one so we don't forget about it.

James-Milligan: https://github.com/open-feature/flagd/issues/29

I think it falls under this issue, and will add a link to this comment so that it is not missed

.gitignore
```diff
@@ -5,5 +5,4 @@ flagd
 dist/
 
 # ignore generated code
-pkg/generated/*.gen.*
```
toddbaert, line 8: @James-Milligan thanks for this, this doesn't delete the file, but it at least makes it appear in my pending changes so I know to delete it!

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "schemas"]
 	path = schemas
 	url = https://github.com/open-feature/schemas
+	branch = main
```
toddbaert, line 4: This is safe because we aren't using `git submodule update --remote` in any automation, but if we did, it would introduce some nondeterminism into the build since every time such a command would run it would get the current head of `main`, instead of a specific SHA:

https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt---remote

It is nice for making it obvious to people how to reference a different branch of the submodule for local work, so I think it makes sense to keep it.

README.md
```diff
@@ -18,20 +18,20 @@ Flagd is a simple command line tool for fetching and presenting feature flags to
...
 
-$ curl -X POST "localhost:8080/flags/myObjectFlag/resolve/object?default-value=foo,bar" -d {}
-// {"reason":"STATIC","value":{"color":"blue"}}
+$ curl -X POST "localhost:8080/flags/myObjectFlag/resolve/object" -d {}
+// {"value":{"key":"val"},"reason":"STATIC","variant":"object1"}
 
-$ curl -X POST "localhost:8080/flags/isColorYellow/resolve/boolean?default-value=true" -d '{"color": "yellow"}'
-// {"reason":"TARGETING_MATCH","value":true}
+$ curl -X POST "localhost:8080/flags/isColorYellow/resolve/boolean" -d '{"color": "yellow"}'
+// {"value":true,"reason":"TARGETING_MATCH","variant":"on"}
```
toddbaert, line 34: Thanks for doing doc updates!

README.md
```diff
@@ -18,20 +18,20 @@ Flagd is a simple command line tool for fetching and presenting feature flags to
...
 ```
```
toddbaert, line 35: You mentioned you might add an error example? You could easily do one with a missing flag:

`curl -X POST "localhost:8080/flags/aMissingFlag/resolve/object" -d {}`

James-Milligan: added :)

pkg/service/http_service.go
```diff
@@ -19,153 +24,84 @@ type HTTPServiceConfiguration struct {
...
+	case st.Code() == codes.NotFound:
+		w.WriteHeader(http.StatusNotFound)
+	default:
+		w.WriteHeader(http.StatusInternalServerError)
 	}
-
-	result, reason, err := s.eval.ResolveObjectValue(flagKey, params.DefaultValue.AdditionalProperties, contextObj)
-	if err != nil {
-		handleEvaluationError(err, reason, w)
+	details := st.Details()
```
toddbaert, line 88: details[0] is an `ErrorResponse`, which doesn't have a `reason`... The end result of this is that `"reason": "ERROR"` isn't returned in the HTTP response, which is probably fine. Similar to the `default-value` and the `flag-key`, the provider using the http client will know this is an error and can set the field accordingly

For reference: https://github.com/open-feature/spec/blob/main/specification/flag-evaluation.md#requirement-148

README.md
```diff
@@ -18,20 +18,20 @@ Flagd is a simple command line tool for fetching and presenting feature flags to
...
+// {"value":"val1","reason":"STATIC","variant":"key1"}
 
-$ curl -X POST "localhost:8080/flags/myNumberFlag/resolve/number?default-value=13" -d {}
-// {"reason":"STATIC","value":1}
+$ curl -X POST "localhost:8080/flags/myNumberFlag/resolve/number" -d {}
+// {"value":1,"reason":"STATIC","variant":"one"}
 
-$ curl -X POST "localhost:8080/flags/myObjectFlag/resolve/object?default-value=foo,bar" -d {}
-// {"reason":"STATIC","value":{"color":"blue"}}
+$ curl -X POST "localhost:8080/flags/myObjectFlag/resolve/object" -d {}
```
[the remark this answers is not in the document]
beeme1mr: Could the example be updated to remove the empty body?

pkg/service/http_service.go
james-milligan: will add, schema update PR found here
https://github.com/open-feature/schemas/pull/11

README.md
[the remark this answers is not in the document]
james-milligan: done

toddbaert (approved):
Looks good! I will open up an issue to resolve the strange coverage thing, which will also include the coverage bot.

Nice work!

pkg/service/grpc_service.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type GRPCService struct {
+	GRPCServiceConfiguration *GRPCServiceConfiguration
+	eval                     eval.IEvaluator
+	gen.UnimplementedServiceServer
+}
+
+func (s *GRPCService) SetEval(eval eval.IEvaluator) {
+	s.eval = eval
```
AlexsJones, line 29: Why not set this on a field, even public is more idiomatic than this?

james-milligan: opened a new PR for this change 

https://github.com/open-feature/flagd/pull/75

PR #75, opened 2022-07-17
removed setter for GRPCService eval field

james-milligan:
Removed the SetEval method. Moved unit tests into package to allow for field access without setter function.

Signed-off-by: James-Milligan <james@omnant.co.uk>

### DIFF
--- a/pkg/service/grpc_service.go
+++ b/pkg/service/grpc_service.go
@@ -25,10 +25,6 @@ type GRPCService struct {
 	gen.UnimplementedServiceServer
 }
 
-func (s *GRPCService) SetEval(eval eval.IEvaluator) {
-	s.eval = eval
-}
-
 func (s GRPCService) Serve(ctx context.Context, eval eval.IEvaluator) error {
 	s.eval = eval
 
--- a/pkg/service/grpc_service_test.go
+++ b/pkg/service/grpc_service_test.go
@@ -1,4 +1,4 @@
-package tests
+package service
 
 import (
 	"context"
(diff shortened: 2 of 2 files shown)


pkg/service/grpc_service_test.go
```diff
@@ -1,4 +1,4 @@
-package tests
+package service
```
toddbaert, line 1: Any opinions on tests being a separate package?

On the one hand, the same package might make things easier... on the other, using a different package forces some "blackbox" testing since we lose access to the packages private identifiers.

Just thinking aloud...

toddbaert (approved):
LGTM, I think this resolves @AlexsJones 's concern... He's on PTO so I'll leave it up to you to choose to merge or not without his input.

pkg/service/grpc_service.go
```diff
@@ -25,10 +25,6 @@ type GRPCService struct {
 	gen.UnimplementedServiceServer
 }
 
-func (s *GRPCService) SetEval(eval eval.IEvaluator) {
-	s.eval = eval
-}
-
 func (s GRPCService) Serve(ctx context.Context, eval eval.IEvaluator) error {
 	s.eval = eval
```
AlexsJones, line 29: I never got time to finish reviewing before this got merged, why is the assignment being done this way?

pkg/service/grpc_service_test.go
AlexsJones: It's not just that, it also gets access to private variables

pkg/service/grpc_service.go
james-milligan: Assignment was previously being handled within the Serve method, for putting together tests access was needed to set a mocked eval. 
As per your comment I have moved the tests into the same package to allow access to the un-exported field, preventing the need for a setter, alternatively we can make eval available as an exported field and put the tests back in their own package.

AlexsJones:
Thanks this is far more idiomatic and readable

AlexsJones (approved):

PR #76, opened 2022-07-22
JSON evaluator using generics

moshebe:
- Replaced `interface{}` with `any` (go 1.18)
- Changed variable name (`bytes` -> `data`) due to lint warning about collision with imported package name
- Re-use the evaluate logic to reduce code duplication in resolveX functions

BTW, saw we are not handling errors "the go way" (defining hard coded strings instead of typed errors)
Didn't tackle it within this PR but would like to hear if there is any specific reason I'm missing of choosing this option.

### DIFF
--- a/pkg/eval/ievaluator.go
+++ b/pkg/eval/ievaluator.go
@@ -21,5 +21,5 @@ type IEvaluator interface {
 		context *structpb.Struct) (value float32, variant string, reason string, err error)
 	ResolveObjectValue(
 		flagKey string,
-		context *structpb.Struct) (value map[string]interface{}, variant string, reasons string, err error)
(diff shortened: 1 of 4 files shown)


AlexsJones (approved):

toddbaert:
> BTW, saw we are not handling errors "the go way" (defining hard coded strings instead of typed errors)
Didn't tackle it within this PR but would like to hear if there is any specific reason I'm missing of choosing this option.

Yep, I'm aware of this too. I agree it should be done. We just have to be sure to return the appropriate error string in the REST API and GPRC transport.

toddbaert (approved):
Looks good to me. It would be great if we could actually consolidate all the `ResolveX` functions into one taking a generic (not 100% sure if this is possible given the protobuf generated stuff) but this is a good start.

PR #78, opened 2022-07-22
Update schema and generate commands

beeme1mr:


toddbaert (approved):
Looks like we had to change imports due to schema changes?

...anyway LGTM

PR #85, opened 2022-07-26
Sync tests

skyerus:


pkg/sync/filepath_sync_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		inotifyChan := make(chan sync.INotify)
+		go func() {
+			fpSync.Notify(ctx, inotifyChan)
+		}()
+
+		// enough time for Notify to be ready to receive file watcher events.
+		// time.Sleep in tests is bad practice. A rework to avoid this would be to pass a ready channel to Notify
+		// to read from to determine readiness.
+		time.Sleep(10 * time.Millisecond)
```
skyerus, line 33: I've put this `time.Sleep` in place for the reason described by the comment above. I'm against this kind of non-deterministic approach, I'd prefer to rework the `Notify` func to accept a `ready chan <- struct{}` for us to determine when the notifier is at a ready state. Where do you guys stand on this? @toddbaert @james-milligan @beeme1mr

james-milligan: I think this is 100% going to be worthwhile to introduce the ```ready chan <- struct{}```, especially if we intend to implement the ```PROVIDER_NOT_READY``` error code as we will need to evaluate this in the check.

In terms of the ```time.Sleep()``` - I find it worrying that there may be a race condition in the tests, although I'm sure that this will be enough time

skyerus: I think so too, in that case I will include the rework as part of this.

Rest assured that introducing the `ready chan<- struct{}` signal would remove the need for `time.Sleep()`.

Makefile
```diff
@@ -36,3 +36,5 @@ uninstall:
 lint:
 	go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@latest
 	${GOPATH}/bin/golangci-lint run --deadline=3m --timeout=3m ./... # Run linters
+mockgen:
+	mockgen -source=pkg/sync/http_sync.go -destination=pkg/sync/mock/http.go -package=syncmock
```
AlexsJones, line 40: Where does this binary come from?

pkg/runtime/runtime.go
```diff
@@ -31,7 +31,9 @@ func startSyncer(ctx context.Context, notifier chan sync.INotify, syncr sync.ISy
 		log.Error(err)
 	}
 
-	go syncr.Notify(ctx, notifier)
+	ready := make(chan struct{})
```
AlexsJones, line 34: Can you outline the purpose of this? It's blocking the notifiers

skyerus: Added comments clarifying what this does: 
- https://github.com/open-feature/flagd/pull/85/commits/7975e4a3d371a96c56107094043cb6cf3acf94d7#diff-e56f9bb947923c6f8bb4c995bf4b9469a0ed117276223d16b177dcb04399a4bcR36
- https://github.com/open-feature/flagd/pull/85/commits/7975e4a3d371a96c56107094043cb6cf3acf94d7#diff-3dd5e32262f4cd7ce7746cc313911ec3bd30069d8a44ddcf4c2b1a7ae4ebc009R11

The runtime synchronise mechanism doesn't care about whether `Notify` is ready to emit events or not but without this the testing is forced to `time.Sleep` to wait for `Notify` to be ready. More info here: https://github.com/open-feature/flagd/pull/85#discussion_r930155185

Makefile
skyerus: Added a comment above this line for clarification

AlexsJones: Screenshot 2022-07-29 at 14 58 08

AlexsJones: Can you add this as a target to auto install? We prefer to have dependencies carried in the Makefile rather than outside the project.

pkg/runtime/runtime.go
AlexsJones: I would be open to having a return channel on the `Notify` method rather than you having to add this into the runtime. WDYT?

Makefile
toddbaert: Definitely agree with this.

skyerus: Good shout, done :)

pkg/runtime/runtime.go
skyerus: I'm not following, are you suggesting changing `Notify` to return a channel?

AlexsJones: Yes that's what I'm thinking, or as a parameter.

skyerus: `ready chan<- struct{}` has been included as a parameter of `Notify` as a part of this PR.
Returning the channel instead wouldn't work as we need the channel prior to `Notify` returning.

AlexsJones: Why create this new channel rather than using the existing channel?

E.g. modifying the struct
```
w <- &Notifier{
					Event: Event[DefaultEventType]{
						Ready: true,
					},
				}
```

What do you think, wouldn't that keep it a bit cleaner?

skyerus: I've reworked the implementation to include a Ready type event on the notifier channel. What do you think?

AlexsJones (approved):
Nicely done, thank you

PR #88, opened 2022-07-28
Default variant validation on SetState

skyerus:


beeme1mr:
Please add a test for this.

AlexsJones (approved):

AlexsJones:
Thank you for your contribution @skyerus

pkg/runtime/runtime.go
```diff
@@ -18,10 +19,13 @@ var (
...
-		return err
+		return fmt.Errorf("fetch: %w", err)
 	}
 	mu.Lock()
-	_ = ev.SetState(msg)
+	err = ev.SetState(msg)
+	if err != nil {
+		return fmt.Errorf("set state: %w", err)
+	}
 	mu.Unlock()
```
AlexsJones, line 29: This mutex will never unlock, can we defer it before the set state?

AlexsJones: @skyerus ^

skyerus: Nice spot, done. I'm surprised this wasn't picked up by the linter

AlexsJones (approved):

PR #89, opened 2022-07-28
Proposal for an issue template

thomaspoignant:
A label called `Needs Triage` will be added to all new issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yaml
+++ b/.github/ISSUE_TEMPLATE/feature.yaml
@@ -1,0 +1,11 @@
+name: Feature
+description: Add new functionality to the project.
+title: "[FEATURE] "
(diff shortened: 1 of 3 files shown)


beeme1mr:
The `Needs Triage` has been created in this repo and at the org level.

beeme1mr (approved):

.github/ISSUE_TEMPLATE/feature.yaml
```diff
@@ -0,0 +1,11 @@
+name: Feature
```
AlexsJones, line 1: Should we mention something about OFEPs for features that go beyond the scope of FlagD?

beeme1mr: I think that's a good idea. I'll open a PR soon.

PR #92, opened 2022-07-29
split out loggers

AlexsJones:
Signed-off-by: Alex Jones <alexsimonjones@gmail.com>

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -10,6 +10,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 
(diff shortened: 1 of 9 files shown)


james-milligan:
Looks good to me, just wondering if any `log.Fatal` calls should be through the service specific loggers? Although I assume they have been left out for a reason

AlexsJones:
> Looks good to me, just wondering if any `log.Fatal` calls should be through the service specific loggers? Although I assume they have been left out for a reason If not it may be worth removing the logger from the HTTPService as it seems to be unused

I have intentionally avoided using fatals to call sys.exits in the program just yet, this was more a like-for-like drop in replacement. 

We can now go back and refine behaviours.

james-milligan (approved):

PR #93, opened 2022-07-29
Targeting rules documentation

skyerus:


beeme1mr:
This is looks great! I think you should mention that the else section that it can be nil instead of a variant name. When you do that, it would use the default variant.

skyerus:
> This is looks great! I think you should mention that the else section that it can be nil instead of a variant name. When you do that, it would use the default variant.

Thanks for the suggestion, done!

AlexsJones (approved):

toddbaert (approved):

PR #99, opened 2022-08-04
support for float64 and int64 flag values

james-milligan:
Integrates schema changes to introduce /int and /float endpoints to http interface, and introduces the respective rpcs.

when using int64 the http interface will return a string value, this is due to the [proto3 json mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) within grpc-gateway. These changes will be reflected in the [golang-sdk-contribs flagd provider](https://github.com/open-feature/golang-sdk-contrib/pull/4).


pkg/eval/json_evaluator.go
```diff
@@ -103,15 +103,25 @@ func (je *JSONEvaluator) ResolveStringValue(flagKey string, context *structpb.St
...
+func (je *JSONEvaluator) ResolveIntValue(flagKey string, context *structpb.Struct) (
+	value int64,
 	variant string,
 	reason string,
 	err error,
 ) {
 	var val float64
 	val, variant, reason, err = resolve[float64](flagKey, context, je.evaluateVariant, je.state.Flags[flagKey].Variants)
-	value = float32(val)
+	value = int64(val)
```
skyerus, line 124: Is this safe? Why are we going to int64 via float64?

james-milligan: Its my understanding that the returned number values from the json flag definitions are all of type float64, so the cast into int64 is required. (Although I would love to be wrong on this one as I'm not happy with the approach either).

As for the type conversion, technically it is unsafe as the conversion may not be lossless, however in these cases the float64 version should be used anyway - what are your thoughts?

AlexsJones (approved):

skyerus (approved):

thomaspoignant (approved):

beeme1mr (approved):

PR #101, opened 2022-08-04
feat: Implemented reusable targeting evaluators

skyerus:


beeme1mr (approved):

config/samples/example_flags.json
```diff
@@ -67,6 +67,30 @@
...
+        "memo": "memo",
+        "loop": "loop",
+        "binet": "binet"
+      },
+      "defaultVariant": "recursive",
+      "state": "ENABLED",
+      "targeting": {
+        "if": [
+          {
+            "$ref": "#/$evaluators/emailWithFaas"
```
toddbaert, line 83: I think the way this is implemented might imply some functionality that doesn't exist.

`#/$evaluators/...` implies that some kind of pathing is supported, when it's really not.


I would recommend we just do something more like: 

```
"$evaluator": "emailWithFaas"
```

That's enough to imply it has some special treatment, but not that we look at some specified path, ie: `#/some/other/path`

skyerus: Thanks for the suggestion, done

toddbaert (approved):

Issue #128, opened 2022-08-10
[FEATURE] Implement uniform ctx cancellation for all services 


james-milligan:
### Requirements

currently context cancellation is not implemented in all services of flagd, and I'm some instances the method of context cancellation is not the same.

All relevant services of flagd should implement context cancellation for a graceful shutdown and the approaches used should all match.

PR #131, opened 2022-08-11
refactor: Context handling updates

james-milligan:
Changes to introduce proper context handling for graceful shutdown

https://github.com/open-feature/flagd/issues/128

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -39,29 +39,27 @@ func startSyncer(ctx context.Context, notifier chan sync.INotify, syncr sync.ISy
 
 	go func() {
 		for {
-			select {
-			case <-ctx.Done():
-				return
-			case w := <-notifier:
-				switch w.GetEvent().EventType {
-				case sync.DefaultEventTypeCreate:
-					logger.Info("New configuration created")
-					if err := updateState(ctx, syncr); err != nil {
-						log.Error(err)
-					}
-				case sync.DefaultEventTypeModify:
-					logger.Info("Configuration modified")
-					if err := updateState(ctx, syncr); err != nil {
-						log.Error(err)
-					}
-				case sync.DefaultEventTypeDelete:
-					logger.Info("Configuration deleted")
-				case sync.DefaultEventTypeReady:
-					logger.Info("Notifier ready")
+			w := <-notifier
+			switch w.GetEvent().EventType {
+			case sync.DefaultEventTypeCreate:
+				logger.Info("New configuration created")
+				if err := updateState(ctx, syncr); err != nil {
+					log.Error(err)
 				}
+			case sync.DefaultEventTypeModify:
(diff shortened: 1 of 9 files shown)


pkg/runtime/runtime.go
```diff
@@ -39,29 +39,27 @@ func startSyncer(ctx context.Context, notifier chan sync.INotify, syncr sync.ISy
...
-				case sync.DefaultEventTypeModify:
-					logger.Info("Configuration modified")
-					if err := updateState(ctx, syncr); err != nil {
-						log.Error(err)
-					}
-				case sync.DefaultEventTypeDelete:
-					logger.Info("Configuration deleted")
-				case sync.DefaultEventTypeReady:
-					logger.Info("Notifier ready")
+			w := <-notifier
```
skyerus, line 42: What's the reason for getting rid of the `<-ctx.Done()` cleanup?

pkg/sync/http_sync.go
```diff
@@ -51,7 +51,7 @@ func (fs *HTTPSync) fetchBodyFromURL(ctx context.Context, url string) ([]byte, e
 	}
 	defer func() { _ = resp.Body.Close() }()
 
-	body, err := ioutil.ReadAll(resp.Body)
+	body, err := io.ReadAll(resp.Body)
```
AlexsJones, line 54: What is the benefit/difference here?

pkg/service/http_service.go
```diff
@@ -111,14 +115,31 @@ func (s *HTTPService) Serve(ctx context.Context, eval eval.IEvaluator) error {
...
 
-	go func() { handleServiceError(httpServer.Serve(httpl)) }() // HTTP
-
-	go func() { handleServiceError(httpServer.Serve(httpsl)) }() // HTTPS
-
-	go func() { handleServiceError(grpcServer.Serve(gorpcl)) }() // GRPC
-	go func() { handleServiceError(tlsm.Serve()) }()
-	go func() { handleServiceError(tcpm.Serve()) }()
-	<-ctx.Done()
+	g.Go(func() error {
```
AlexsJones, line 118: How do errors work here? 
If the first routine exits immediately does the error return get waited on until all return?
Or does the exit cause all other routines to exit?

pkg/runtime/runtime.go
james-milligan: Following our discussion I'm going to roll this back

pkg/service/http_service.go
james-milligan: any error returned by one of the goroutines in the `errgroup` will cancel the group context will cancel the group context and `<-gCtx.Done()` will no longer block. The original error will be passed out from `g.Wait()`

james-milligan:
closed as changes are required for [this PR](https://github.com/open-feature/flagd/pull/135), if those changes are rejected I will reopen this PR

PR #135, opened 2022-08-12
refactor: reduced complexity of /cmd

james-milligan:
Proposed refactor to reduce the complexity of `/cmd`
My intentions with these changes are to make the code a little easier to follow for new contributors
`/cmd` is more lightweight, takes the arguments and passes them to the `RuntimeFromConfig()` method, then calls the cobra command specific method on `Runtime`, e.g. `Start()` or `StartDebug()`


AlexsJones:
@thschue what do you think of this?

pkg/runtime/start.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"golang.org/x/sync/errgroup"
+)
+
+func (r *Runtime) Start() {
```
skyerus, line 12: This func could return an error and the caller can handle it instead of logging at this level.
This is subjective though so I'll leave you to make the call.

pkg/runtime/start.go
```diff
@@ -0,0 +1,45 @@
...
+	if r.Service == nil {
+		r.Logger.Error("no Service set")
+		return
+	}
+	if r.SyncImpl == nil || len(r.SyncImpl) == 0 {
```
skyerus, line 17:
```suggestion
	if len(r.SyncImpl) == 0 {
```
nil slices have length 0

pkg/runtime/start.go
james-milligan: I was thinking about this, the error would then need to be handled at the `/cmd` level which is why I took this approach. I would be keen to see if anyone feels strongly either way

skyerus: I suppose given the naming of `cmd` vs `runtime` I think handling the errors at the level you've chosen makes sense

skyerus (approved):
Nice work :)

thschue (approved):
lgtm

AlexsJones:
The people have spoken. 
Will give it another review once conflicts solved

thschue (approved):

skyerus (approved):

AlexsJones (approved):

toddbaert:
I reviewed this only after it was merged, but I can say I agree it's an improvement.

Issue #138, opened 2022-08-16
[FEATURE] Version CLI Flag


beeme1mr:
## Overview

Add a flag to the CLI that returns the version.

## Example

input:
`flagd --version`
output:
`v0.0.8`

AlexsJones:
How do you feel about changing to `flagd version` it's slightly more golang idiomatic?

AlexsJones:
https://github.com/open-feature/flagd/pull/141

beeme1mr:
I have no strong preference either way as long as it's documented.

beeme1mr:
There's some inconsistencies with the version depending on how the binary is installed:

## Release binary
```shell
wget https://github.com/open-feature/flagd/releases/download/v0.0.9/flagd_0.0.9_Linux_x86_64.tar.gz
tar -xvf flagd_0.0.9_Linux_x86_64.tar.gz
./flagd version
```
### Returns
```shell
flagd 0.0.9 (bd9aed0c5d9d2e0498f45a54aeb845f51ea04fae) built at 2022-08-17T20:17:08Z
```

## Docker:
```shell
docker run -it ghcr.io/open-feature/flagd:v0.0.9 version
```

### Returns
```shell
flagd ghcr.io/open-feature/flagd:v0.0.9 (bd9aed0c5d9d2e0498f45a54aeb845f51ea04fae) built at 2022-08-17
```

## Go install
```shell
go install github.com/open-feature/flagd@v0.0.9
flagd version
```

### Returns
```shell
flagd dev (HEAD) built at unknown
```

AlexsJones:
Good spot, though the last one looks as expected, it's not a compiled release it's pulling and building the binary before downloading? I don't know how that works @james-milligan any pointers?

The other two are using different date formatters, which I can sort out @beeme1mr

AlexsJones:
https://github.com/open-feature/flagd/pull/145

james-milligan:
Had a look into it and we can 'combo' `debug.ReadBuildInfo()` [docs](https://pkg.go.dev/runtime/debug#ReadBuildInfo) with our current approach, when we use `go install` the version field is populated on build, this can be seen by running `go version -m flagd` in your `go/bin` which includes `mod	github.com/open-feature/flagd	v0.0.9` in the return.
So if we check the variables affected by the `ld` flags and they are the default, then we can pull from `debug.ReadBuildInfo()` and see if the version is there.

james-milligan:
```
[...]

PR #141, opened 2022-08-16
feat: Version

AlexsJones:


Makefile
```diff
@@ -1,15 +1,16 @@
 IMG ?= flagd:latest
 PHONY: .docker-build .build .run .mockgen
 PREFIX=/usr/local
+VERSION=0.0.8
```
beeme1mr, line 4: We can have release please automatically bump this version 😄

toddbaert: For example: https://github.com/open-feature/node-sdk-contrib/blob/3bc87990f3db5dae27969b5eaa9c1ee77dbb7ef3/libs/hooks/open-telemetry/src/lib/open-telemetry-hook.ts#L23

beeme1mr (approved):

main.go
```diff
@@ -1,7 +1,26 @@
...
+
+var (
+	version = "dev"
+	commit  = "HEAD"
+	date    = "unknown"
+)
 
 func main() {
-	cmd.Execute()
+	showVersion := flag.Bool("version", false, "show version")
```
james-milligan, line 18: this could be added in as a cobra command to keep things consistent

```
// versionCmd represents the version command
var versionCmd = &cobra.Command{
	Use:   "version",
	Short: "Get Flagd version",
	Long:  ``,
	Run: func(cmd *cobra.Command, args []string) {
		fmt.Printf("flagd %s (%s) built at %s\n", version, commit, date)
		os.Exit(0)
	},
}

// added into either `Execute()` or `init()`
rootCmd.AddCommand(versionCmd)
```

then can be run with `flagd version`

Makefile
beeme1mr: Will this need to be maintained manually?

AlexsJones: I could have it increment in the release job if you want me to add that into this PR?

main.go
AlexsJones: Will see what he thinks in https://github.com/open-feature/flagd/issues/138

Makefile
beeme1mr: I think we need to automate this or we'll inevitably forget about it in the future.

AlexsJones:
Screenshot 2022-08-17 at 16 54 31

Makefile
AlexsJones: Well, it's also the same with OFO. 
I would like to do them at the same time, I'll open an issue after this to track

AlexsJones: Or we could base the version off the current /ref/tag which probably would give the same outcome.

AlexsJones: automated this @beeme1mr

PR #142, opened 2022-08-16
feat: added support for environment variables

skyerus:


cmd/start.go
```diff
@@ -7,37 +7,39 @@ import (
...
 )
 
 func findService(name string) (service.IService, error) {
+	serverCertPath := viper.GetString(serverCertPathFlagName)
+	serverKeyPath := viper.GetString(serverKeyPathFlagName)
 	registeredServices := map[string]service.IService{
 		"http": &service.HTTPService{
 			HTTPServiceConfiguration: &service.HTTPServiceConfiguration{
-				Port:           servicePort,
+				Port:           viper.GetInt32(portFlagName),
```
skyerus, line 42: Forced to go for this `viper` retrieval approach: https://github.com/spf13/viper/issues/85

beeme1mr:
I think we should set the prefix `FLAGD` for the environment variables. It looks like you can use `SetEnvPrefix` in [Viper](https://github.com/spf13/viper#working-with-environment-variables).

skyerus:
> I think we should set the prefix `FLAGD` for the environment variables. It looks like you can use `SetEnvPrefix` in [Viper](https://github.com/spf13/viper#working-with-environment-variables).

Done.

docs/configuration.md
```diff
@@ -0,0 +1,24 @@
...
+  -c, --server-cert-path string   Server side tls certificate path
+  -k, --server-key-path string    Server side tls key path
+  -s, --service-provider string   Set a service provider e.g. http or grpc (default "http")
+  -d, --socketpath string         flagd socket path (default "/tmp/flagd.sock")
+  -y, --sync-provider string      Set a sync provider e.g. filepath or remote (default "filepath")
+  -f, --uri strings               Set a sync provider uri to read data from this can be a filepath or url. Using multiple providers is supported where collisions between flags with the same key, the later will be used.
+```
+
+Environment variable keys are uppercased, prefixed with `FLAGD_` and all `-` are replaced with `_`. For example,
+`sync-provider` in environment variable form is `FLAGD_SYNC_PROVIDER`.
```
toddbaert, line 22: Love this.

toddbaert:
This looks good to me. It should make configuration less error-prone in K8s.

toddbaert (approved):

beeme1mr (approved):

AlexsJones:
This will impact OFO

skyerus:
> This will impact OFO

How so? The existing behaviour hasn't been affected, any config previously setup ought to work as before unless I've missed something.

AlexsJones:
I would think we should harmonise on using envs through the pod spec.

Issue #124, opened 2022-08-09
[FEATURE] Unix socket support


james-milligan:
### Requirements

[OFEP](https://github.com/open-feature/research/blob/main/006-OFEP-flagd-sockets.md)

Requirements:

- [x] Create a socket path parameter within flagD
- [x] Implement uniform ctx cancellation for all services
- [x] `grpc_service.go` to support unix sockets through a new method that returns a net connection
- [x] Modifiy the existing `http_service` code to leverage a similar net connection start
- [x] Write tests

james-milligan:
@AlexsJones is socket support for the `http_sevice` still part of the plan? Happy to take a look

AlexsJones:
Go for it, should be quite quick to implement !

PR #146, opened 2022-08-19
feat: Socket support

james-milligan:
https://github.com/open-feature/flagd/issues/124

Provides socket support to the gRPC service.
When the `--socket-path` / `-d` flag is set host and port are overridden and a unix socket service is opened instead.

Required provider updates for existing flagd providers: [node](https://github.com/open-feature/node-sdk-contrib/tree/main/libs/providers/flagd) and [go](https://github.com/open-feature/golang-sdk-contrib/pull/11)


AlexsJones (approved):
LOVELY!

pkg/service/grpc_service.go
```diff
@@ -49,9 +52,16 @@ func (s *GRPCService) Serve(ctx context.Context, eval eval.IEvaluator) error {
 	grpcServer := grpc.NewServer(serverOpts...)
 	gen.RegisterServiceServer(grpcServer, s)
 
-	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", s.GRPCServiceConfiguration.Port))
-	if err != nil {
-		return err
+	if s.GRPCServiceConfiguration.ServerSocketPath != "" {
+		lis, err = net.Listen("unix", s.GRPCServiceConfiguration.ServerSocketPath)
+		if err != nil {
```
skyerus, line 57: These 2 error checks could be a single error check after the outer `if {} else {}` statement

skyerus (approved):
Nice work!

PR #148, opened 2022-08-19
feat: added benchmarks to GitHub Actions

james-milligan:
Nightly cron to populate a gh-pages benchmarks page to track historical data
Benchmark job on PRs to compare to historical data and flag any impacts >5%


toddbaert (approved):

.github/workflows/build.yaml
```diff
@@ -11,6 +11,31 @@ on:
...
+       - uses: actions/checkout@v2
+       - name: Setup go
+         uses: actions/setup-go@v1
+         with:
+           go-version: '1.18.1' 
+       - name: Run benchmark
+         run: go test -bench . -benchmem ./... | tee output.txt
+
+       - name: Store benchmark result
+         uses: benchmark-action/github-action-benchmark@v1
```
toddbaert, line 27: The fact an action exists for this makes me feel a bit better, but I'd still like to see what performance guarantees github provides in actions.

@james-milligan did you happen to find any info on this?

beeme1mr (approved):

.github/workflows/build.yaml
beeme1mr: I think this is the best we can do without a paid service. It at least provides a decent baseline for us to use.

PR #157, opened 2022-08-26
feat: kubernetes sync

AlexsJones:
https://github.com/open-feature/research/pull/31
Screenshot 2022-08-30 at 15 54 16


pkg/sync/kubernetes/featureflagconfiguration/featureflagconfiguration.go
```diff
@@ -0,0 +1,74 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+)
+
+type Interface interface {
+	List(opts metav1.ListOptions) (*v1alpha1.FeatureFlagConfigurationList, error)
+	Get(name string, options metav1.GetOptions) (*v1alpha1.FeatureFlagConfiguration, error)
+	Create(*v1alpha1.FeatureFlagConfiguration) (*v1alpha1.FeatureFlagConfiguration, error)
+	Watch(opts metav1.ListOptions) (watch.Interface, error)
+	// ...
```
beeme1mr, line 18: Remove comment

beeme1mr (commented):
Please don't forget to update the [configuration docs](https://github.com/open-feature/flagd/blob/main/docs/configuration.md).

README.md
```diff
@@ -252,6 +252,17 @@ returns
...
 
+### Customising sync providers
+
+Custom sync providers can be used to provide flag evaluation logic.
+The Kubernetes provider allows flagD to connect to a Kubernetes cluster and evaluate flags against a specified FeatureFlagConfiguration resource as defined within the [open-feature-operator](https://github.com/open-feature/open-feature-operator/blob/main/apis/core/v1alpha1/featureflagconfiguration_types.go) spec.
+
+To use an existing FeatureFlagConfiguration custom resource, start flagD with the following command:
+
+```shell
+flagd start --sync-provider=kubernetes --sync-provider-args=featureflagconfiguration=my-example --sync-provider-args=featureflagconfigurationnamespace=default
```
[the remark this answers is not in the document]
AlexsJones: There isn't.. provider args is a new concept and I just went with the first I thought of
```
featureflagconfiguration
featureflagconfigurationnamespace
```

We could move to  `featureflagconfigurationname` to make it more idiomatic or add spaces, I don't mind

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,117 @@
...
+				// It is important to now assertain if it is an actual flag configuration change
+				updateFuncHandler(oldObj, newObj, object, c)
+			},
+		},
+	)
+	go ffConfigController.Run(wait.NeverStop)
+}
+
+func NewForConfig(c *rest.Config) (*FFCClient, error) {
+	config := *c
```
skyerus, line 95: Need a nil check here. Is there any reason that `c *rest.Config` can't be `c rest.Config`?

pkg/sync/kubernetes/featureflagconfiguration/featureflagconfiguration.go
```diff
@@ -0,0 +1,74 @@
...
+type FeatureFlagClient struct {
+	restClient rest.Interface
+	ns         string
+}
+
+func (c *FeatureFlagClient) List(opts metav1.ListOptions) (*v1alpha1.FeatureFlagConfigurationList, error) {
+	result := v1alpha1.FeatureFlagConfigurationList{}
+	err := c.restClient.
+		Get().
+		Resource("featureflagconfigurations").
```
skyerus, line 29: A constant for `"featureflagconfigurations"` would be nice

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+	k.Logger.Infof("Starting kubernetes sync notifier for resource %s", k.ProviderArgs["featureflagconfiguration"])
+	kubeconfig := os.Getenv("KUBECONFIG")
+
+	// Create the client configuration
+	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		k.Logger.Panic(err.Error())
+		os.Exit(1)
```
skyerus, line 60: If the logger's panic behaves the same way the inbuilt panic does then this line is redundant

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	k.client, err = featureflagconfiguration.NewForConfig(config)
+	if err != nil {
+		panic(err)
```
skyerus, line 64: Why not use the logger's panic here?

cmd/start.go
```diff
@@ -40,6 +43,8 @@ func init() {
 	flags.StringP(evaluatorFlagName, "e", "json", "Set an evaluator e.g. json")
 	flags.StringP(serverCertPathFlagName, "c", "", "Server side tls certificate path")
 	flags.StringP(serverKeyPathFlagName, "k", "", "Server side tls key path")
+	flags.StringToStringVarP(&ProviderArgs, ProviderArgsFlagName,
```
skyerus, line 46: This should be done in the same fashion as the others if possible, otherwise this flag won't be able to fallback to environment variable if not given

README.md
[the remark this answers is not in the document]
AlexsJones: It's relevent to the sync service though, it seems like it wouldn't make sense what it is, but we could do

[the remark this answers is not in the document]
AlexsJones: Sure I don't mind

cmd/start.go
AlexsJones: I have found no other way to assign to maps through cobra, let me know though and ill change it

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
AlexsJones: Added. From what I remember we prefer a pointer to this as parts of the config will be updated dynamically during the session and we don't want copies of it.

README.md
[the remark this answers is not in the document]
beeme1mr: The name just seems a bit too verbose at the moment but it's up to you.

README.md
```diff
@@ -252,6 +252,8 @@ returns
 { "value": true, "reason": "TARGETING_MATCH", "variant": "off" }
 ```
 
+### [Customisation of sync providers](./docs/configuration.md)
```
beeme1mr, line 255:
```suggestion
### [Customization of sync providers](./docs/configuration.md)
```

AlexsJones:
> FeatureFlagConfiguration

Have you done `make install` from open-feature-operator recently?
You'll need the latest CRD's for this to work

AlexsJones:
e

> I was able to get it working but I noticed a few issues:
> 
> * A "configuration modified" log message is repeated ~ 5 seconds when when a configuration hasn't changed.
> * The following error was logged after a flag change was detected:
> 
> ```
> W0830 16:13:45.048499   19741 reflector.go:347] pkg/mod/k8s.io/client-go@v0.25.0/tools/cache/reflector.go:169: watch of *v1alpha1.FeatureFlagConfiguration ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: no kind \"FeatureFlagConfiguration\" is registered for the internal version of group \"core.openfeature.dev\" in scheme \"pkg/runtime/scheme.go:100\"") has prevented the request from succeeding
> ```

The config modified is because the status fields on the K8s object update far more than we're used too with other syncs.. I could remove the notification or lower the log level?

cmd/start.go
skyerus: `flags.StringToStringP(`
and
`_ = viper.BindPFlag(`
and
`viper.GetStringMapString(`
ought to do it unless I've missed something

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
skyerus: This line creates a copy of the value at the pointer so nothing done here changes the passed in config `c`

cmd/start.go
```diff
@@ -15,13 +15,16 @@ const (
 	serviceProviderFlagName = "service-provider"
 	socketPathFlagName      = "socket-path"
 	syncProviderFlagName    = "sync-provider"
+	ProviderArgsFlagName    = "sync-provider-args"
```
skyerus, line 18: Can this be lowercased as it's not exported?

cmd/start.go
```diff
@@ -15,13 +15,16 @@ const (
...
 	evaluatorFlagName       = "evaluator"
 	serverCertPathFlagName  = "server-cert-path"
 	serverKeyPathFlagName   = "server-key-path"
 	uriFlagName             = "uri"
 	bearerTokenFlagName     = "bearer-token"
 )
 
+var ProviderArgs map[string]string
```
skyerus, line 26: Is this still needed?

cmd/start.go
```diff
@@ -51,13 +56,12 @@ func init() {
 	_ = viper.BindPFlag(socketPathFlagName, flags.Lookup(socketPathFlagName))
 	_ = viper.BindPFlag(serviceProviderFlagName, flags.Lookup(serviceProviderFlagName))
 	_ = viper.BindPFlag(syncProviderFlagName, flags.Lookup(syncProviderFlagName))
+	_ = viper.BindPFlag(syncProviderFlagName, flags.Lookup(syncProviderFlagName))
```
skyerus, line 59: This ought to be `providerArgsFlagName`

cmd/start.go
```diff
@@ -80,6 +87,7 @@ var startCmd = &cobra.Command{
 			ServiceKeyPath:    viper.GetString(serverKeyPathFlagName),
 
 			SyncProvider:    viper.GetString(syncProviderFlagName),
+			ProviderArgs:    viper.GetStringMapString(syncProviderFlagName),
```
skyerus, line 90: This ought to be `viper.GetStringMapString(providerArgsFlagName)`

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,118 @@
...
+				// Typically this could be anything from a status field to a spec field
+				// It is important to now assertain if it is an actual flag configuration change
+				updateFuncHandler(oldObj, object, c)
+			},
+		},
+	)
+	go ffConfigController.Run(wait.NeverStop)
+}
+
+func NewForConfig(config *rest.Config) (*FFCClient, error) {
```
skyerus, line 96: Still need a nil check here

cmd/start.go
AlexsJones: yes and using as the type in the config now

cmd/start.go
```diff
@@ -15,13 +15,16 @@ const (
...
 	syncProviderFlagName    = "sync-provider"
+	providerArgsFlagName    = "sync-provider-args"
 	evaluatorFlagName       = "evaluator"
 	serverCertPathFlagName  = "server-cert-path"
 	serverKeyPathFlagName   = "server-key-path"
 	uriFlagName             = "uri"
 	bearerTokenFlagName     = "bearer-token"
 )
 
+var ProviderArgs map[string]string
```
james-milligan, line 26: this variable seems to be unused

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,122 @@
...
+		ns = object.Namespace
+	}
+	_, ffConfigController := cache.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func(lo metav1.ListOptions) (result runtime.Object, err error) {
+				res, err := clientSet.FeatureFlagConfigurations(ns).List(lo)
+				return res, err
+			},
+			WatchFunc: func(lo metav1.ListOptions) (watch.Interface, error) {
+				wiface, err := clientSet.FeatureFlagConfigurations(ns).Watch(lo)
```
skyerus, line 73: can return this directly rather than assigning to variables

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,122 @@
...
+	object client.ObjectKey, c chan<- sync.INotify,
+) {
+	ns := "*"
+	if object.Namespace != "" {
+		ns = object.Namespace
+	}
+	_, ffConfigController := cache.NewInformer(
+		&cache.ListWatch{
+			ListFunc: func(lo metav1.ListOptions) (result runtime.Object, err error) {
+				res, err := clientSet.FeatureFlagConfigurations(ns).List(lo)
```
skyerus, line 69: can return this directly rather than assigning to variables

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -0,0 +1,94 @@
...
+var refreshTime = time.Second * 5
+
+type Sync struct {
+	Logger       *log.Entry
+	ProviderArgs sync.ProviderArgs
+	client       *featureflagconfiguration.FFCClient
+}
+
+func (k *Sync) Fetch(ctx context.Context) (string, error) {
+	if k.ProviderArgs["featureflagconfiguration"] == "" {
```
skyerus, line 27: constants would be nice for these keys

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
AlexsJones: Yeah this was done for debugging, will revert!

docs/configuration.md
```diff
@@ -22,3 +23,22 @@ Environment variable keys are uppercased, prefixed with `FLAGD_` and all `-` are
 `sync-provider` in environment variable form is `FLAGD_SYNC_PROVIDER`.
 
 Config file expects the keys to have the exact naming as the flags.
+
+
+### Customising sync providers
+
+Custom sync providers can be used to provide flag evaluation logic.
+The Kubernetes provider allows flagD to connect to a Kubernetes cluster and evaluate flags against a specified FeatureFlagConfiguration resource as defined within the [open-feature-operator](https://github.com/open-feature/open-feature-operator/blob/main/apis/core/v1alpha1/featureflagconfiguration_types.go) spec.
```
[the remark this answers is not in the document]
AlexsJones: Would just having it as a sub heading be okay? It seems hard to illustrate it nicely in a MD table

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,128 @@
...
+				}
+			},
+			DeleteFunc: func(obj interface{}) {
+				deleteFuncHandler(obj, object, c)
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				// This indicates a change to the custom resource
+				// Typically this could be anything from a status field to a spec field
+				// It is important to now assertain if it is an actual flag configuration change
+				updateFuncHandler(oldObj, object, c)
```
toddbaert, line 96: > It is important to now assertain if it is an actual flag configuration change

I don't see any logic trying to do this (maybe I'm missing something). Is this a place where we'd want to perhaps use something like a hashing mechanism to determine if the actual flag spec changed?

docs/configuration.md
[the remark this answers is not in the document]
beeme1mr: Sure, that works.

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
AlexsJones: I did try to implement it initially but comparing the old/newObject featureflagspec didn't work, there is a time component plus the complexity of the cache refresh that just made it a bit awkward. The current implementation is more naive but works 100%

AlexsJones: I will remove the comments

toddbaert (approved):
Obviously this is a great enhancement! Looks good to me.

cmd/start.go
james-milligan, line 26: `ProviderArgs` is also defined in pkg/sync, this one doesn't seem to be used, within cmd the argument seems to just be referenced as a `map[string]string` with viper

AlexsJones: Removed

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,142 @@
...
+			},
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				// This indicates a change to the custom resource
+				if err := updateFuncHandler(oldObj, newObj, object, c); err != nil {
+					l.Warn(err.Error())
+				}
+			},
+		},
+	)
+	go ffConfigController.Run(wait.NeverStop)
```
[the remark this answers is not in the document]
AlexsJones: I don't think this really matters, I did look at moving the errgroup down into this level but it involves modifying the interface. How else would you see this working? A local errgroup inside this function or something else @james-milligan ?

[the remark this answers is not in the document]
james-milligan: I was thinking we could just pass through the context via `WatchResources` then pass the `ctx.Done` channel, again i'm not sure if its required, but if its offered as standard by the `Run` method it may be worth implementing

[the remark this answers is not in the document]
AlexsJones: Okay updated

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -0,0 +1,141 @@
...
+		c <- &sync.Notifier{
+			Event: sync.Event[sync.DefaultEventType]{
+				EventType: sync.DefaultEventTypeDelete,
+			},
+		}
+	}
+	return nil
+}
+
+func WatchResources(l log.Entry, clientSet FFCInterface, refreshTime time.Duration,
```
james-milligan, line 77: small change, but `ctx` is usually defined as the first argument of a function signature, I can't find an explicit definition for this but its mentioned [in here ](https://go.dev/blog/context-and-structs)

PR #163, opened 2022-09-20
feat!: Updated service to use connect

james-milligan:
Connect is a family of libraries for building browser and gRPC-compatible HTTP APIs. It is backwards compatible with gRPC and allows for all currently listed service providers and their 'variations' (gRPC, http(s), unix / tcp) to be managed by a single service provider with a single set of handlers.

Using connect introduces the `connect protocol`, directly compatible with the `connect-go` andf `connect-web` client libraries. This provides the benefit avoiding the use of heavy libraries such as `grpc-web` in the front end applications. It is also fully  compatible with gRPC meaning that any existing clients that match the interface will still work, thus, providers can still be writen for other languages using pure grpc clients, without requiring a language specifc conenct library.

Breaking changes: 

The http routes are changed, there is currently no support for providing 'custom' routes. A request which previously will have looked like this:
```
$ curl -X POST "localhost:8013/flags/headerColor/resolve/string" -d '{"email": "foo@bar.com"}'
{"value":"#0000FF","reason":"TARGETING_MATCH","variant":"blue"}
```
will not look like this: 
```
$ curl -X POST "localhost:8013/schema.v1.Service/ResolveString" -d ''{"flagKey":"headerColor","context":{"email": "foo@bar.com"}}'' -H "Content-Type: application/json"
{"value":"#0000FF","reason":"TARGETING_MATCH","variant":"blue"}
``` 
This may not be an issue going forwards as these URLs are internal, and will only requre an update to the provider.
route prefixes can be added to allow for /api/ to be appended to the front of each route.

There is no longer a `--service-provider` option, this is because all variations of this flagkey are now available through the single `connect` service. (there may be argument for keeping this flag and setting the default to `'connect'`)


toddbaert:
I'll look at this first thing tomorrow.

I'd like to get this in.

docs/configuration.md.orig
```diff
@@ -0,0 +1,50 @@
...
+Supported flags are as follows (result of running `./flagd start --help`):
+
+```
+  -b, --bearer-token string       Set a bearer token to use for remote sync
+  -e, --evaluator string          Set an evaluator e.g. json (default "json")
+  -h, --help                      help for start
+  -p, --port int32                Port to listen on (default 8013)
+  -c, --server-cert-path string   Server side tls certificate path
+  -k, --server-key-path string    Server side tls key path
+<<<<<<< HEAD
```
AlexsJones, line 15: merge conflict

AlexsJones:
Nice reduction in code, happy times

james-milligan:
> Although the benchmark does seem to be hanging for some reason, and there's errors in it's logs.

I have resolved the error log on the benchmarks, the hanging seems to be on writing to the gh-pages branch so i will look into this

pkg/service/connect_service.go
```diff
@@ -0,0 +1,230 @@
...
+		return res, err
+	}
+	res.Msg.Reason = reason
+	res.Msg.Value = val
+	res.Msg.Variant = variant
+	return res, nil
+}
+
+func newCORS() *cors.Cors {
+	// To let web developers play with the demo ConnectService from browsers, we need a
```
skyerus, line 187: Are there plans for this to be configurable?

skyerus (approved):
Nice work :)

docs/configuration.md
```diff
@@ -12,7 +12,6 @@ Supported flags are as follows (result of running `./flagd start --help`):
   -p, --port int32                Port to listen on (default 8013)
   -c, --server-cert-path string   Server side tls certificate path
   -k, --server-key-path string    Server side tls key path
-  -s, --service-provider string   Set a service provider e.g. http or grpc (default "http")
   -a, --sync-provider-args        Sync provider arguments as key values separated by =
   -d, --socket-path string        Set the flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally
```
beeme1mr, line 16: `With http(s) the grpc-gateway proxy will use this address internally`

Is that still true?

README.md
```diff
@@ -36,13 +36,13 @@ This now provides an accessible http or [https](#https) endpoint for flag evalua
...
 ```sh
-curl -X POST "localhost:8013/flags/myBoolFlag/resolve/boolean"
+curl -X POST "localhost:8013/schema.v1.Service/ResolveBoolean" -d '{"flagKey":"myBoolFlag","context":{}}' -H "Content-Type: application/json"
 ```
 
 Result:
 
 ```sh
-{"value":true,"reason":"STATIC","variant":"on"}
+{"value":true,"reason":"TARGETING_MATCH","variant":"on"}
```
beeme1mr, line 45: I'm surprised this is a targeting match. I would have expected `DEFAULT`.

@toddbaert you had some thoughts on the reason values. What do you think?

james-milligan: apologies this is a typo - ive updated this to STATIC, there is another draft PR open to change these values to DEFAULT, awaiting a conclusion to the discussion in the openfeature-flagd slack channel

docs/configuration.md
james-milligan: removed

AlexsJones (approved):

toddbaert:
I think you need to run a `go mod tidy`.

README.md
beeme1mr: Agreed, thanks.

toddbaert:
@james-milligan You've missed updating a couple old REST examples in the readme.

pkg/service/connect_service.go
toddbaert: CORS will only tell browsers to ignore GET responses or prevent POST responses if the pre-flight fails Neither does it protect flagd from a load perspective, since as an attacker I can simply not use a browser and bypass CORS entirely that way.

CORS mitigates issues where malicious origin (webpage) X uses origin Y's credentials (stored in a cookie usually) to make a request to origin Y's backend, usually performing some kind of mutating action. That's not really applicable to us.

toddbaert: Thinking about it more, you might be right @skyerus . I think there IS some possible DDoS protection here... If I infected a number of web pages (or host malicious web pages) that call a flagd I want to attack, I can use my origin/infected web page to tell  other users to spam that flagd target, adding load to it in a DDoS. Though I could attack flagd directly, my malicious origin is making my users attack that flagd too. CORS actually could help us with this, since a flagd would deny pre-flight requests from my hostile origins.

I think we should actually support CORS configuration, eventually, though it's probably something better configured in an LB or infra of some kind.

toddbaert: I will make an issue to either support CORS, or document that it should be configured in infra.

james-milligan: i initially added this config to allow for the flagd-web client demo to take place, happy to remove this until flagd-web is implemented, as im not sure how the external configuration for this should look as of yet

skyerus: That's interesting, I guess it would be useful in the scenario where a host exposes flagd directly to the public domain.

toddbaert (approved):

pkg/service/connect_service.go
beeme1mr: This seem like something a reverse proxy should handle.

AlexsJones: +1 to CORS being handled at the reverse proxy...

pkg/service/connect_service.go
```diff
@@ -0,0 +1,247 @@
...
+	case model.FlagNotFoundErrorCode:
+		return connect.NewError(connect.CodeNotFound, err)
+	case model.TypeMismatchErrorCode:
+		return connect.NewError(connect.CodeInvalidArgument, err)
+	case model.DisabledReason:
+		return connect.NewError(connect.CodeUnavailable, err)
+	case model.ParseErrorCode:
+		return connect.NewError(connect.CodeDataLoss, err)
+	}
+	return err
```
toddbaert, line 246: Thanks!

AlexsJones:
What does the `feat!` mean @toddbaert is this not ready for merge?

toddbaert:
> What does the `feat!` mean 

`feat!` is just a [semantic convention](https://www.conventionalcommits.org/en/v1.0.0/) for git commits.

`!` means breaking, and can be applied to any commit prefix, but only `feat` and `fix` are user-facing and will appear in the changelog automatically. I don't consider this a "refactor" because it makes user-visible changes (REST endpoints, etc) so it's not a refactor.

> @toddbaert is this not ready for merge?

I'll leave that to @james-milligan . It looks good to me!

docs/configuration.md
```diff
@@ -12,9 +12,8 @@ Supported flags are as follows (result of running `./flagd start --help`):
   -p, --port int32                Port to listen on (default 8013)
   -c, --server-cert-path string   Server side tls certificate path
   -k, --server-key-path string    Server side tls key path
-  -s, --service-provider string   Set a service provider e.g. http or grpc (default "http")
   -a, --sync-provider-args        Sync provider arguments as key values separated by =
-  -d, --socket-path string        Set the flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally
+  -d, --socket-path string        Set the flagd socket path.
   -y, --sync-provider string      Set a sync provider e.g. filepath or remote (default "filepath")
   -f, --uri strings               Set a sync provider uri to read data from this can be a filepath or url. Using multiple providers is supported where collisions between flags with the same key, the later will be used.
```
beeme1mr, line 18: Can you please document the CORS option?

james-milligan: added

AlexsJones (approved):

PR #165, opened 2022-09-20
refactor: Static reason type removal

james-milligan:
Todd: This particular "reason" never made it into the spec, and only the node sdk implements it. I think it's value is minimal. I struggle a bit to find a good differentiation between DEFAULT and STATIC .

I propose removing this and replacing all instances of it with DEFAULT. If you guys think there's significant value in it though, we can make a spec change... I just think the ROI is low at this point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ curl -X POST "localhost:8013/flags/myBoolFlag/resolve/boolean"
 Result:
 
 ```sh
-{"value":true,"reason":"STATIC","variant":"on"}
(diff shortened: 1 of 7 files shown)


toddbaert (approved):

beeme1mr (approved):

beeme1mr:
@james-milligan are you still working on this or can we mark the PR as ready for review?

AlexsJones (approved):

PR #167, opened 2022-09-21
feat: only fire modify event when FeatureFlagConfiguration Generation field has changed

skyerus:
Signed-off-by: Skye Gill <gill.skye95@gmail.com>

### DIFF
--- a/pkg/sync/kubernetes/featureflagconfiguration/clientset.go
+++ b/pkg/sync/kubernetes/featureflagconfiguration/clientset.go
@@ -42,7 +42,7 @@ func createFuncHandler(obj interface{}, object client.ObjectKey, c chan<- sync.I
 	return nil
 }
 
-func updateFuncHandler(oldObj interface{}, newObj interface{}, object client.ObjectKey, c chan<- sync.INotify) error {
+func updateFuncHandler(oldObj interface{}, newObj interface{}, c chan<- sync.INotify) error {
 	if reflect.TypeOf(oldObj) != reflect.TypeOf(&ffv1alpha1.FeatureFlagConfiguration{}) {
 		return errors.New("old object is not a FeatureFlagConfiguration")
 	}
@@ -107,7 +107,7 @@ func WatchResources(ctx context.Context, l log.Entry, clientSet FFCInterface, re
 			},
 			UpdateFunc: func(oldObj, newObj interface{}) {
 				// This indicates a change to the custom resource
-				if err := updateFuncHandler(oldObj, newObj, object, c); err != nil {
+				if err := updateFuncHandler(oldObj, newObj, c); err != nil {
 					l.Warn(err.Error())
 				}
 			},
--- a/pkg/sync/kubernetes/featureflagconfiguration/clientset.go
+++ b/pkg/sync/kubernetes/featureflagconfiguration/clientset.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"reflect"
-	"time"
 
 	"github.com/open-feature/flagd/pkg/sync"
 	ffv1alpha1 "github.com/open-feature/open-feature-operator/apis/core/v1alpha1"
@@ -42,7 +41,7 @@ func createFuncHandler(obj interface{}, object client.ObjectKey, c chan<- sync.I
 	return nil
 }
 
-func updateFuncHandler(oldObj interface{}, newObj interface{}, c chan<- sync.INotify) error {
+func updateFuncHandler(oldObj interface{}, newObj interface{}, object client.ObjectKey, c chan<- sync.INotify) error {
 	if reflect.TypeOf(oldObj) != reflect.TypeOf(&ffv1alpha1.FeatureFlagConfiguration{}) {
 		return errors.New("old object is not a FeatureFlagConfiguration")
 	}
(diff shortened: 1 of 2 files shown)


skyerus:
This field changes when `FeatureFlagConfiguration` has been updated in some way

AlexsJones:
> > Have you tested this? I had this in the original implementation but issues ever getting it to work.
> 
> Confirmed that this works for me. There is a slight delay in the update but there always was, right? Or did the delay get fixed at some point?

Are we talking 5 seconds? Because that's the cache refresh time

skyerus:
> > > Have you tested this? I had this in the original implementation but issues ever getting it to work.
> > 
> > 
> > Confirmed that this works for me. There is a slight delay in the update but there always was, right? Or did the delay get fixed at some point?
> 
> Are we talking 5 seconds? Because that's the cache refresh time

No, closer to a minute sadly.

AlexsJones:
> > > > Have you tested this? I had this in the original implementation but issues ever getting it to work.
> > > 
> > > 
> > > Confirmed that this works for me. There is a slight delay in the update but there always was, right? Or did the delay get fixed at some point?
> > 
> > 
> > Are we talking 5 seconds? Because that's the cache refresh time
> 
> No, closer to a minute sadly.

Interestingly that's the issue I had - I think the notifier has different update cycles based on whether it's the metadata update versus the spec itself. Can I test your changes before we go further? I will try and figure out if there is something we can tweak in the notifier itself.

skyerus:
> > > > > Have you tested this? I had this in the original implementation but issues ever getting it to work.
> > > > 
> > > > 
> > > > Confirmed that this works for me. There is a slight delay in the update but there always was, right? Or did the delay get fixed at some point?
> > > 
> > > 
> > > Are we talking 5 seconds? Because that's the cache refresh time
> > 
> > 
> > No, closer to a minute sadly.
> 
> Interestingly that's the issue I had - I think the notifier has different update cycles based on whether it's the metadata update versus the spec itself. Can I test your changes before we go further? I will try and figure out if there is something we can tweak in the notifier itself.

Sounds good, I'll have a play around too.

AlexsJones:
I will do this, but might need a few days till I get time

skyerus:
> I will do this, but might need a few days till I get time

I was able to get it working @AlexsJones  🎉

Tested alongside OFO, with this change the event is emitted immediately & exclusively when a change occurs.

pkg/sync/kubernetes/featureflagconfiguration/clientset.go
```diff
@@ -91,7 +92,7 @@ func WatchResources(ctx context.Context, l log.Entry, clientSet FFCInterface, re
 			},
 		},
 		&ffv1alpha1.FeatureFlagConfiguration{},
-		refreshTime,
+		0,
```
skyerus, line 95: No need to poll every x seconds

AlexsJones: That's outside the scope of this change really, the polling should be configurable because it puts quite a lot of load on the API server

skyerus: > - resyncPeriod: if non-zero, will re-list this often (you will get OnUpdate
calls, even if nothing changed). Otherwise, re-list will be delayed as
long as possible (until the upstream source closes the watch or times out,
or you stop the controller).

I've gone into a rabbit hole of trying to understand what the re-sync does. Digging around it seems like a re-sync re-fires events from the cached queue (hence why we're warned of OnUpdate calls, even if nothing changed). Based on this assumption I think that having no re-sync period reduces load as it avoids the re-list and re-firing of events.
https://groups.google.com/g/kubernetes-sig-api-machinery/c/PbSCXdLDno0?pli=1
https://github.com/kubernetes-client/java/issues/725#issuecomment-540352746

What do you think?

AlexsJones: Happy to take your word for it, we actually need to do some proper performance testing and until then we should build it to our best ability then we can take stock of the tuning.

AlexsJones: I think for now we make it configurable via the `ProviderArgs` system, what do you think?

skyerus: > Just looking at that thread give me some pauses for concern
> 
> > Recommendations found in the wild as to the right resync period seem to gravitate towards "keep it rather high to avoid burdening the API server too much", and I often see values on the order of multiple minutes or even hours. However, this advice seems to render the resync period unsuitable as a means to compensate for watch errors for somewhat latency-critical controllers.

From my understanding that advice is if you're setting the refresh sync period to a non zero value.
I've just stepped into the code to confirm that having a non-zero re-sync period performs more work than zero:
https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L267

I think leaving it configurable is good too but I'd opt for the default as 0, what do you reckon?

skyerus: > If you've confirmed this via debugger, it seems like the right choice to me. It's too bad that sort of detective work was necessary.

Poor choice of words from me, I didn't use a debugger but it's relatively easy to see the code path in this case thankfully (I wish I'd gone through the code earlier).

AlexsJones: I agree, make it configurable as a parameter but set the default to zero

skyerus: Done

AlexsJones (approved):

james-milligan (approved):

PR #171, opened 2022-09-23
fix: checkout release tag before running container and binary releases

beeme1mr:
Checks out the release tag before running the container and binary release steps. This ensures that Go Releaser is able to detect the version number properly and upload the artifacts.

I've tested this in a branch. You can see the results [GitHub Release](https://github.com/beeme1mr/flagd/releases/tag/v0.2.3). I've also verified that both the container and binaries return the expected version numbers when running `flagd version`.

Signed-off-by: Michael Beemer <beeme1mr@users.noreply.github.com>


AlexsJones (approved):

.github/workflows/release-please.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          submodules: recursive
```
toddbaert, line 35: We don't need the submodules?

beeme1mr: No, it was replaced with [this](https://github.com/open-feature/flagd/blob/main/go.mod#L15).

toddbaert: Thanks 👍

toddbaert (approved):

PR #177, opened 2022-09-28
feat: Event rpc

james-milligan:
*If agreement is made to incorporate these changes the `flagd-dev` schema changes will be merged into the `flagd` schema in the buf registry and the import paths will be updated. At this point ill move the PR from draft to open*

This PR adds a new rpc to the connect service called `EventStream`, it accepts an empty proto message `google.proto.Empty` and returns a stream of message type `EventStreamResponse`. 
`EventStreamResponse` contains a single string field of `type` which references an internal notification type used within flagd, such as `provider_ready` or `configuration_change`. This rpc will allow for the management of the sdk cache, busting it once a configuration change is detected. It will also allow for the implementation of flag change events once/if the OFEP passes.

[flagd-dev schema](https://buf.build/open-feature/flagd-dev/file/main:schema/v1/schema.proto)

### DIFF
--- a/pkg/service/connect_service.go
+++ b/pkg/service/connect_service.go
@@ -13,10 +13,11 @@ import (
 	"github.com/open-feature/flagd/pkg/model"
 	"github.com/rs/cors"
 	log "github.com/sirupsen/logrus"
-	schemaV1 "go.buf.build/open-feature/flagd-connect/open-feature/flagd/schema/v1"
-	schemaConnectV1 "go.buf.build/open-feature/flagd-connect/open-feature/flagd/schema/v1/schemav1connect"
+	schemaV1 "go.buf.build/open-feature/flagd-connect/open-feature/flagd-dev/schema/v1"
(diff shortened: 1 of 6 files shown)


pkg/service/connect_service.go
```diff
@@ -96,14 +98,44 @@ func (s *ConnectService) setupServer() (net.Listener, error) {
...
+		send <- n
+	}
+}
+
+func (s *ConnectService) EventStream(
+	ctx context.Context,
+	req *connect.Request[emptypb.Empty],
+	stream *connect.ServerStream[schemaV1.EventStreamResponse],
+) error {
+	s.subs[req] = make(chan NotificationType, 1)
```
skyerus, line 118: do we need a mutex on this?

james-milligan: added lock to surround the writing to and deleting of the channels

james-milligan: the creation of the channel does not need a mutex as the pointer value will not be reused until after the defer function runs, so will be threadsafe

PR #187, opened 2022-10-04
feat: Eventing

james-milligan:
## This PR

- adds eventing rpc 
- events are emited on successful connection to stream rpc and configuration changes

This PR adds a new rpc to the connect service called EventStream, it accepts an empty proto message google.proto.Empty and returns a stream of message type EventStreamResponse.
EventStreamResponse contains a single string field of type which references an internal notification type used within flagd, such as provider_ready or configuration_change. This rpc will allow for the management of the sdk cache, busting it once a configuration change is detected. It will also allow for the implementation of flag change events once/if the OFEP passes.


toddbaert:
@AlexsJones just for some background, one of the coolest parts of this is it allows providers to cache resolved flag values (based on a sha of the `flag-key` and `context`). The provider can then intelligently bust that cache when the flagd configuration changes because it will get an event. This significantly reduces traffic and latency, even in server deployments. You can imagine that if user X has already evaluated flag "banner-color" seconds ago, there's no reason to evaluate that again.

config/samples/example_flags.json
```diff
@@ -6,8 +6,8 @@
         "on": true,
         "off": false
       },
-      "defaultVariant": "on"
```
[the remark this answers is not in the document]
james-milligan: resolved

pkg/eval/ievaluator.go
```diff
@@ -4,13 +4,27 @@ import (
...
+}
+
 /*
 IEvaluator implementations store the state of the flags,
 do parsing and validation of the flag state and evaluate flags in response to handlers.
 */
 type IEvaluator interface {
 	GetState() (string, error)
-	SetState(source string, state string) error
+	SetState(source string, state string) ([]StateChangeNotification, error)
```
AlexsJones, line 27: This interface is becoming ugly, maybe we should pass an object to encapsulate all this?

```
SetState(state IState) 
```

pkg/eval/ievaluator.go
```diff
@@ -26,3 +40,11 @@ type IEvaluator interface {
 		flagKey string,
 		context *structpb.Struct) (value map[string]any, variant string, reasons string, err error)
 }
+
+func (s *StateChangeNotification) ToMap() map[string]interface{} {
+	return map[string]interface{}{
```
AlexsJones, line 45: Do we need some guards here on s.type?

AlexsJones:
> @AlexsJones just for some background, one of the coolest parts of this is it allows providers to cache resolved flag values (based on a sha of the `flag-key` and `context`). The provider can then intelligently bust that cache when the flagd configuration changes because it will get an event indicating such. This significantly reduces traffic and latency, even in server deployments. You can imagine that if user X has already evaluated flag "banner-color" seconds ago, there's no reason to evaluate that again.

It's a good move, I think the interfaces are getting a bit bruised though - suggesting we encapsulate with a configuration object.

pkg/eval/ievaluator.go
james-milligan: s.Type is a typed string of `StateChangeNotificationType`, it needs to be cast into a plain string type to allow for the structpb library to handle it. This type conversion will never fail.

pkg/eval/ievaluator.go
james-milligan: I think matching the types may lead to a little confusion, we could wrap the input arguments as 
```go
type IEvaluatorStateConfiguration struct {
	Source string
	State  string
}
```
and update the output argument type to be `[]IEvaluatorStateChange`

`SetState(state IEvaluatorStateConfiguration) ([]IEvaluatorStateChange, error)`

what do you think @AlexsJones

AlexsJones: Works for me yes

AlexsJones (approved):

pkg/service/connect_service.go
```diff
@@ -96,14 +105,55 @@ func (s *ConnectService) setupServer() (net.Listener, error) {
...
+func (s *ConnectService) EventStream(
+	ctx context.Context,
+	req *connect.Request[emptypb.Empty],
+	stream *connect.ServerStream[schemaV1.EventStreamResponse],
+) error {
+	s.eventingConfiguration.subs[req] = make(chan Notification, 1)
+	defer func() {
+		s.eventingConfiguration.mu.Lock()
+		delete(s.eventingConfiguration.subs, req)
+		s.eventingConfiguration.mu.Lock()
```
beeme1mr, line 123:
```suggestion
		s.eventingConfiguration.mu.Lock()
		delete(s.eventingConfiguration.subs, req)
		s.eventingConfiguration.mu.Unlock()
```

PR #190, opened 2022-10-11
chore: exclude component in git tag

beeme1mr:
This change prevents Release Please from adding the component name in the release tag. It looks like currently that's not a problem but this will prevent it from happening if the component name detection changes in the future.

Signed-off-by: Michael Beemer <beeme1mr@users.noreply.github.com>

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -3,6 +3,8 @@
   "packages": {
     ".": {
       "release-type": "go",
+      "monorepo-tags": false,
+      "include-component-in-tag": false,
       "prerelease": false,
       "bump-minor-pre-major": true,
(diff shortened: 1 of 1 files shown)


toddbaert:
> Just for posterity can you give an example?

Here's a release that has a component name in Java: https://github.com/open-feature/java-sdk/releases/tag/dev.openfeature.javasdk-v0.2.2

Instead of the tag just being `0.2.2` like would be preferred in a repo with only a single artifact, there's a "component name" in the tag (dev.openfeature.javasdk).

toddbaert (approved):

AlexsJones (approved):

PR #206, opened 2022-11-02
feat: wip support yaml evaluator

### LINKED ISSUE #180: [FEATURE] Yaml Evaluator Support
### Requirements

Extend the evaluation option to support yaml.

`flagd start -e yaml -f example_flags.yaml`

vadasambar:
## This PR
- adds support for YAML evaluator

### Related Issues
Fixes #180 

### Notes

TBD
### Follow-up Tasks

TBD
### How to test

TBD


vadasambar:
This PR is WIP. I will request review once I am done by mentioning the reviewer here.

toddbaert:
Hey @vadasambar ! Since it tends to be trivial to convert yaml <=> json, I would suggest somehow approaching the problem with that in mind. You can leverage the existing JSON evaluator's logic and state storage and just change the parsing, I think. Perhaps composition could help here?

pkg/sync/filepath_sync.go
```diff
@@ -28,7 +34,37 @@ func (fs *FilePathSync) Fetch(_ context.Context) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return string(rawFile), nil
+
+	var r []byte
+	if fs.FileType == "yaml" {
```
AlexsJones, line 39: Should think about supporting. `.yml`

config/samples/example_flags.yaml
```diff
@@ -0,0 +1,88 @@
+flags:
```
vadasambar, line 1: I converted example_flags.json to yaml using https://www.json2yaml.com/ and tweaked the yaml a bit.

pkg/sync/filepath_sync.go
```diff
@@ -90,3 +106,25 @@ func (fs *FilePathSync) Notify(ctx context.Context, w chan<- INotify) {
...
+func yamlToJson(rawFile []byte) (string, error) {
+	var ms map[string]interface{}
+	if err := yaml.Unmarshal(rawFile, &ms); err != nil {
+		return "", fmt.Errorf("unmarshal yaml: %w", err)
+	}
+
+	// Using jsoniter library here because json.Marshal doesn't understand keys of
+	// interface{} type (it can only understand string keys)
+	// More info:https://stackoverflow.com/q/35377477/6874596
+	var jsonit = jsoniter.ConfigCompatibleWithStandardLibrary
```
toddbaert, line 121: I'm not sure about this dependency. We already unmarshal this structure in `pkg/eval/json_evaluator_model.go`, and the structure only has string keys.

Have you tried unmarshaling into a `map[string]interface{}`? I think that should be sufficient.

skyerus: The yaml parser interprets the nested structures as `map[interface{}]interface{}`, a solution for this is to either use a library or [convert the keys to string](https://stackoverflow.com/a/40737676/10039498)

toddbaert: OK I understand the problem a bit better. I'd prefer to do a conversion ourselves instead of using a library though, but that's my opinion.

Interested in what others think is the best way to go.

vadasambar: @skyerus thank you for explaining. @toddbaert basically what @skyerus said (SO link is in the code comment above). My comment might be difficult to understand. I will re-word it.

vadasambar: > OK I understand the problem a bit better. I'd prefer to do a conversion ourselves instead of using a library though, but that's my opinion.

I guess doing the conversion ourselves would give us more control but I am not sure how valuable that would be since all we want to do is convert yaml to json (not to say we'd have to maintain that code + potential bugs in our code vs a popular library)

toddbaert: > > OK I understand the problem a bit better. I'd prefer to do a conversion ourselves instead of using a library though, but that's my opinion.
> 
> I guess doing the conversion ourselves would give us more control but I am not sure how valuable that would be since all we want to do is convert yaml to json (not to say we'd have to maintain that code + potential bugs in our code vs a popular library)

Understood. I think I'd like to hear from @skyerus @AlexsJones @james-milligan and if they think the lib is fine I'm fine with that approach.

skyerus: I'm leaning towards doing the conversion ourselves (if it is as simple as the stackoverflow suggestion implies) to avoid the additional surface area exposed by introducing a new lib.

vadasambar: Do you want me to make this change?

vadasambar: Nevermind, I think the problem is we are using `gopkg.in/yaml.v2` in the above PR.
`gopkg.in/yaml.v3` seems to Unmarshal into map[string]interface{} for sub-maps too which should fix the issue without importing any library or writing the code for conversion.

vadasambar: Example: https://play.golang.com/p/4iCzr8UmEHT

AlexsJones: I would prefer us to use `gopkg.in/yaml.v3`

pkg/sync/filepath_sync.go
```diff
@@ -14,6 +17,9 @@ type FilePathSync struct {
 	URI          string
 	Logger       *logger.Logger
 	ProviderArgs ProviderArgs
+	// FileType indicates the file type e.g., json, yaml/yml etc.,
+	// TODO: limit this to `json` or `yaml only?`
```
beeme1mr, line 21: Please address this TODO.

beeme1mr (commented):
Could you please also update the configuration documentation?

https://github.com/open-feature/flagd/blob/main/docs/configuration.md

vadasambar:
> Could you please also update the configuration documentation?
> 
> https://github.com/open-feature/flagd/blob/main/docs/configuration.md

@beeme1mr slipped my mind. I will do it. Thanks for pointing it out. :pray:

beeme1mr:
Hi @vadasambar, could you please fix the conflicts and resolve the issue with the test?

vadasambar:
> Hi @vadasambar, could you please fix the conflicts and resolve the issue with the test?

Planning to do that. Sorry, a little busy with something else this week.

vadasambar:
Need approval to run the workflow :pray: 
image

pkg/sync/filepath_sync.go
vadasambar: After a second thought, I removed the TODO because we already have a switch case to check the file type.

vadasambar: Please unresolve this conversation if you have any feedback for me.

pkg/sync/filepath_sync.go
vadasambar: Added support for `yml`. Thanks for pointing it out.

vadasambar:
Sorry for the delay in addressing comments. Need another approval to run the workflow :pray:

vadasambar:
I will re-request the reviews once the workflow passes and I fix any issues with the workflow.

Kavindu-Dodan:
> I will re-request the reviews once the workflow passes and I fix any issues with the workflow.

@vadasambar   Please also address the pending reviews

pkg/sync/filepath_sync.go
```diff
@@ -28,6 +33,7 @@ func (fs *FilePathSync) Fetch(_ context.Context) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
 	return string(rawFile), nil
```
vadasambar, line 37: I was checking the file extension here in [an earlier commit](https://github.com/open-feature/flagd/pull/206/commits/3aba0064b4ce431e77e39d8eebe42e3faa3bced7) to ensure only certain file extensions like `json`, `yaml` and `yml` would be allowed but on a second thought I realized there can be files without `.json`/`.yaml`/`.yml`  extension but still have valid `json` or `yaml` content (we use such a file in our tests).

If we check the file extension, we can prevent certain cases where the user passes the wrong file as config (e.g., `config.txt` instead of `config.json` or `config.yaml`) but it also means we might have to throw an error if there is no file extension (since we don't know if the file without extension has `yaml` contents, `json` contents or something else).

vadasambar: I am not sure what we should do here. 
> a second thought I realized there can be files without .json/.yaml/.yml extension but still have valid json or yaml content (we use such a file in our tests).

I removed the file extension check after thinking about this ^ but if the reviewers have any other opinion/feedback I would like to hear it :pray:

vadasambar:
Re-requested review. I have covered the points I could think of (if you find anything missing or if you have any feedback please mention me :pray:).

vadasambar:
> > I will re-request the reviews once the workflow passes and I fix any issues with the workflow.
> 
> @vadasambar Please also address the pending reviews

I thought I addressed all of them. Did I leave out something?

toddbaert:
@vadasambar only 1 minor issue according to the CI: `Error: pkg/sync/filepath_sync.go:102:6: func `yamlToJSON` is unused (unused)`

docs/configuration/configuration.md
```diff
@@ -6,17 +6,18 @@ highest priority, followed by environment variables and finally config file.
...
+  -e, --evaluator string                    Set an evaluator e.g. json, yaml/yml (default "json")
+  -h, --help                                help for start
+  -m, --metrics-port int32                  Port to serve metrics on (default 8014)
+  -p, --port int32                          Port to listen on (default 8013)
+  -c, --server-cert-path string             Server side tls certificate path
+  -k, --server-key-path string              Server side tls key path
+  -d, --socket-path string                  Flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
+  -y, --sync-provider string                DEPRECATED: Set a sync provider e.g. filepath or remote
+  -a, --sync-provider-args stringToString   Sync provider arguments as key values separated by = (default [])
+  -f, --uri strings                         Set a sync provider uri to read data from, this can be a filepath,url or FeatureFlagConfiguration. Using multiple providers is supported however ifflag keys are duplicated across multiple sources it may lead to unexpected behavior
```
beeme1mr, line 20:
```suggestion
  -f, --uri strings                         Set a sync provider uri to read data from, this can be a filepath, url or FeatureFlagConfiguration. Using multiple providers is supported however if flag keys are duplicated across multiple sources it may lead to unexpected behavior.
```

pkg/sync/filepath_sync.go
vadasambar: Please ignore the above comments ^ :pray:

pkg/sync/filepath_sync.go
```diff
@@ -28,7 +33,17 @@ func (fs *FilePathSync) Fetch(_ context.Context) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return string(rawFile), nil
+
+	switch fs.FileType {
```
vadasambar, line 37: `FileType` here is the same as the value for the `--evaluator` flag of `flagd start ... -e json` (if no evaluator is specified, [`json` is used as the default](https://github.com/open-feature/flagd/blob/8c8e155342154b4e1af87e24e68ab71a9446699e/cmd/start.go#L40))

toddbaert:
@vadasambar FYI, you should be able to run `make test` and `make lint` locally to confirm that they are working instead of waiting on the CI runs.

vadasambar:
> @vadasambar only 1 minor issue according to the CI:
> 
> `` Error: pkg/sync/filepath_sync.go:102:6: func `yamlToJSON` is unused (unused) ``

Thank you for noticing that! Helped me fix a mistake in the code. Should be fixed now.

vadasambar:
> @vadasambar FYI, you should be able to run `make test` and `make lint` locally to confirm that they are working instead of waiting on the CI runs.

That's what I am using now. Noticed we had a `make test` after requesting the workflow run twice. Sorry about that!

pkg/sync/filepath_sync_test.go
```diff
@@ -117,8 +117,9 @@ func TestFilePathSync_Fetch(t *testing.T) {
 	}{
 		"success": {
 			fpSync: sync.FilePathSync{
-				URI:    fmt.Sprintf("%s/%s", dirName, fetchFileName),
-				Logger: logger.NewLogger(nil, false),
+				URI:      fmt.Sprintf("%s/%s", dirName, fetchFileName),
+				Logger:   logger.NewLogger(nil, false),
+				FileType: "json", // this is the default
```
vadasambar, line 122: We aren't using `FilePathSync`'s `Fetch` to fetch files via non-command line means (in which case `FileType` won't be set to `json` by default) as of now AFAIK. Might change in the future. If someone has a better approach, I would like know.

docs/configuration/configuration.md
vadasambar: Looks there is a small [typo/bug in the existing code](https://github.com/open-feature/flagd/blob/8c8e155342154b4e1af87e24e68ab71a9446699e/cmd/start.go#L47-L48)

vadasambar: Fixed. Thank you!

cmd/start.go
```diff
@@ -37,15 +37,15 @@ func init() {
...
+	flags.StringP(evaluatorFlagName, "e", "json", "Set an evaluator e.g. json, yaml/yml")
 	flags.StringP(serverCertPathFlagName, "c", "", "Server side tls certificate path")
 	flags.StringP(serverKeyPathFlagName, "k", "", "Server side tls key path")
 	flags.StringToStringP(providerArgsFlagName,
 		"a", nil, "Sync provider arguments as key values separated by =")
 	flags.StringSliceP(
 		uriFlagName, "f", []string{}, "Set a sync provider uri to read data from, this can be a filepath,"+
 			"url or FeatureFlagConfiguration. Using multiple providers is supported however if"+
-			"flag keys are duplicated across multiple sources it may lead to unexpected behavior ",
+			" flag keys are duplicated across multiple sources it may lead to unexpected behavior ",
```
vadasambar, line 48: Added spacing
`...however ifflag keys are duplicated` ->  `...however if flag keys are duplicated`

docs/configuration/configuration.md
skyerus: Looks like these changes are still required

vadasambar: I think I forgot to click `Sign off and commit suggestion` after writing the commit message. Fixed. Thank you for noticing it!

Kavindu-Dodan:
@vadasambar  Other than highlighted "switch case" improvement, I think PR is ready.

vadasambar:
> @vadasambar Other than highlighted "switch case" improvement, I think PR is ready.

Your suggested way looks cleaner. Thanks for the feedback. I will update the PR by tomorrow. :pray:

pkg/runtime/from_config.go
```diff
@@ -49,6 +49,10 @@ func (r *Runtime) setService(logger *logger.Logger) {
 
 func (r *Runtime) setEvaluatorFromConfig(logger *logger.Logger) error {
 	switch r.config.Evaluator {
+	case "yaml":
+		fallthrough
+	case "yml":
+		fallthrough
 	case "json":
```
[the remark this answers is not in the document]
vadasambar: Updated the code with the suggested change

pkg/sync/filepath_sync.go
```diff
@@ -28,7 +34,17 @@ func (fs *FilePathSync) Fetch(_ context.Context) (string, error) {
...
+	case "yaml":
+		fallthrough
+	case "yml":
+		return yamlToJSON(rawFile)
```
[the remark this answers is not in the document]
vadasambar: Updated with the suggested change. Thank you for the link.

vadasambar:
> > @vadasambar Other than highlighted "switch case" improvement, I think PR is ready.
> 
> Your suggested way looks cleaner. Thanks for the feedback. I will update the PR by tomorrow. pray

Updated the code.

docs/configuration/configuration.md
```diff
@@ -6,17 +6,18 @@ highest priority, followed by environment variables and finally config file.
...
+  -b, --bearer-token string                 Set a bearer token to use for remote sync
+  -C, --cors-origin strings                 CORS allowed origins, * will allow all origins
+  -e, --evaluator string                    Set an evaluator e.g. json, yaml/yml (default "json")
+  -h, --help                                help for start
+  -m, --metrics-port int32                  Port to serve metrics on (default 8014)
+  -p, --port int32                          Port to listen on (default 8013)
+  -c, --server-cert-path string             Server side tls certificate path
+  -k, --server-key-path string              Server side tls key path
+  -d, --socket-path string                  Flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
+  -y, --sync-provider string                DEPRECATED: Set a sync provider e.g. filepath or remote
```
skyerus, line 18: It looks like the state of these configuration options have drifted from `main`. For example, `-y` is no longer an option.

skyerus (changes requested):

docs/configuration/configuration.md
toddbaert: @vadasambar I think the only change needed to this file is to mention the yaml/json evaluations work the same, and yaml/json format are auto-detected by the file extension.

vadasambar: @skyerus it is still an option (although `DEPRECATED`)

https://github.com/open-feature/flagd/blob/af6f133688b506f52ac6180890e3d785238d7b40/cmd/start.go#L54

If you are suggesting we remove a deprecated option from the docs, I am not sure if that's such a good idea. If you think I am misunderstanding what you are seeing, please feel free to correct me.

vadasambar: @toddbaert thank you for the feedback. I have taken some liberty in re-wording your suggestion like this:
```
I think the only change needed to this file is to mention the yaml/json evaluations work the same
```
to
```
Please note that yaml/yml and json evaluations work the same (yaml/yml files are converted to json internally)
```
I added extra context. 

As for
```
yaml/json format are auto-detected by the file extension.
```
I have re-worded it to (and put it next it `-f` flag because I think the end-user would look at `-f` flag help for such information)
```
Please note that if you are using filepath, flagd only supports files with .yaml/.yml/.json extension.
```
This is because I thought `auto-detected by file extension` seemed a little vague to the end-user who probably won't be as well-versed in the internals of flagd. The problem with my sentence is, it doesn't explain the "why" but your suggestion does. 

This is not final. If you have any opinions, please let me know :pray: I am open to changing things.

Kavindu-Dodan: Yes, seems @vadasambar  needs a rebase after a fresh upstream sync. As you can see in the current master version [1], we no longer have -y option. And no, @skyerus  is not saying you should remove that, but rather pointing to you that you have diverted from upstream (flagd main).

If you go to your fork [https://github.com/vadasambar/flagd/tree/main] , you can see it says 

> This branch is [13 commits behind](https://github.com/vadasambar/flagd/compare/main...open-feature:flagd:main) open-feature:main.

which is the root cause.

@vadasambar, can you perform the following, 

1. Update your fork's [https://github.com/vadasambar/flagd] main branch from the latest flagd main [https://github.com/open-feature/flagd]. 

This is easy and doable through Git UI. Check the following screenshot for "Sync fork" feature

image


2. Once the above step is done, rebase this branch on your fork's main OR Merge your fork's master to this branch if you are  familiar using merge instead of rebases 

This should resolve the conflict we see here.

[1] https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md?plain=1

vadasambar: @Kavindu-Dodan I haven't updated my `origin` but I don't think it's related (I pull the latest from `upstream` `main` in my branch which is why you don't see weird diffs in this PR).  

I checked `main` of https://github.com/open-feature/flagd before posting my comment. I can still see the `-y` option in the code in the `main` branch. 

Here's a screenshot:
image

Kavindu-Dodan: Yes and that option is in a "deprecation" state. You can check this PR which contains the change - https://github.com/open-feature/flagd/pull/222 . And PR removed the configuration entry. 

Further, the most recent change to this file is done through this commit -  https://github.com/open-feature/flagd/commit/d6147490fd939f227025b2875b2a1888b4e4a3fe  . This is why you need to sync your fork and then update your PR so it reflect recent changes.

vadasambar: I don't think we should [remove a flag from docs](https://github.com/open-feature/flagd/pull/222/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180L18) *just because it is deprecated* (not a good user experience). That is why I have added it back. I haven't seen a tool/software I have used or worked with removing docs for deprecated flags/features until the flag was completely removed (including Kubernetes). 

> While a deprecated software feature remains in the software, its use may raise warning messages recommending alternative practices. Deprecated status may also indicate the feature will be removed in the future. Features are deprecated, rather than immediately removed, to provide backward compatibility and to give programmers time to bring affected code into compliance with the new standard. 

https://en.wikipedia.org/wiki/Deprecation
TL:DR; deprecation is not equal to removal. 

> Further, the most recent change to this file is done through this commit - https://github.com/open-feature/flagd/commit/d6147490fd939f227025b2875b2a1888b4e4a3fe . This is why you need to sync your fork and then update your PR so it reflect recent changes.

You are looking in the wrong place? Check https://github.com/vadasambar/flagd/commits/feat/180/add-yaml-evaluator. Maybe I am not explaining it well. Here's what I did:
```
# I am on my git branch vadasambar:feat/180/add-yaml-evaluator
$ git pull upstream main --rebase
....
```
My upstream points to github.com/open-feature/flagd
```
$ git remote -v
origin  git@github.com:vadasambar/flagd.git (fetch)
origin  git@github.com:vadasambar/flagd.git (push)
upstream        git@github.com:open-feature/flagd.git (fetch)
upstream        git@github.com:open-feature/flagd.git (push)
```
Because of which my `main` branch in my fork i.e., https://github.com/vadasambar/flagd is behind https://github.com/open-feature/flagd repo's `main` branch which is *okay* but my `vadasambar:feat/180/add-yaml-evaluator` is *up-to-date*. 

I have done something analogous to https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork#syncing-a-fork-branch-from-the-command-line but my local branch is `vadasambar:feat/180/add-yaml-evaluator`

vadasambar: If the flagd community as a whole thinks we should not be keeping docs for `deprecated` flags, I will go ahead and remove it (but I strongly disagree with the practice). I will wait for comments from other reviewers before I remove it.

vadasambar:
I will fix the linting error (forgot to run `make lint` locally before pushing my changes)

docs/configuration/configuration.md
Kavindu-Dodan: @vadasambar This shows why a PR must focus on its intended purpose - for example, this PR should focus on "adds support for YAML evaluator" as in the description of the PR. My assumption was that your branch diverted and hence the configuration entry on "-y" option.

IMO, if you think this is a bad practice and the flag should still exist, then it should have been added through a dedicated PR with a dedicated description/justification. This way, we can maintain a clear Git history, have focused reviews & discussions :)

toddbaert: I think I agree with @vadasambar that it actually makes sense to keep this doc and mention it's deprecated, though it's not really the original focus of this PR.

In the future I would generally recommend to open new issues about such things to make things easier for everyone and keep discussions separate. I personally really want this yaml feature added and don't want it to be blocked any longer than necessary for the somewhat unreleated `--sync provider` arg deprecation discussion.

However, since the work has already been done on this branch I'm fine with it.

toddbaert:
@vadasambar ```./flagd start --uri file://home/todd/git/flagd/config/samples/example_flags.yaml  -e=yaml``` works as expected, but

```./flagd start --uri file://home/todd/git/flagd/config/samples/example_flags.yaml``` errors. I think it would improve usability to default to the evaluator matching the file extension. Do you agree?

Kavindu-Dodan:
> vadasambar

Detecting flag configuration format based on file extension is a great usability improvement. However, I think this require substantial code change as well as proper documentation. 

I created a follow-up issue #240, where we could have dedicated focus on this

Kavindu-Dodan (approved):

toddbaert (approved):
Approving with https://github.com/open-feature/flagd/pull/206#issuecomment-1367557964

Thanks again for all your effort, @vadasambar

vadasambar:
> @vadasambar `./flagd start --uri file://home/todd/git/flagd/config/samples/example_flags.yaml -e=yaml` works as expected, but
> 
> `./flagd start --uri file://home/todd/git/flagd/config/samples/example_flags.yaml` errors (I believe because in that case it anticipates JSON). I think it would improve usability to default to the evaluator matching the file extension. Do you agree?
> 
> What do you think @Kavindu-Dodan ?

Makes sense to me. :+1:

vadasambar:
> > vadasambar
> 
> Detecting flag configuration format based on file extension is a great usability improvement. However, I think this require substantial code change as well as proper documentation.
> 
> I created a follow-up issue #240, where we could have dedicated focus on this

Thank you for creating the issue :pray:

vadasambar:
> Approving with [#206 (comment)](https://github.com/open-feature/flagd/pull/206#issuecomment-1367557964)
> 
> Thanks again for all your effort, @vadasambar

Not at all. Thank you for the feedback and review :pray:

vadasambar:
I think we need review from @AlexsJones to merge this PR (not sure if we want to keep it open until https://github.com/open-feature/flagd/issues/240 is done)

vadasambar:
I think we need review from @AlexsJones 
image

toddbaert:
> I think we need review from @AlexsJones to merge this PR (not sure if we want to keep it open until #240 is done)

I don't think we need to include that in this PR, it seems like something we can roll in later, but feel free to start working on it if you are interested.

Issue #184, opened 2022-10-04
Review Logging


beeme1mr:
There are a few minor issues and inconsistencies in the way logs are handled in FlagD. Review the logging strategy to see how they can be made more consistent.

Areas to focus on:

- Many messages contain a carriage return in the message.
- The structure of the logs vary based on the message.
- Make sure the log message aligns with the message.
- Add additional logging when in debug mode. Examples include:
   - Request source (e.g. gRPC, REST) if this information is easily available
   - The flag key and context used in the flag evaluation
   - The result of the flag evaluation

PR #207, opened 2022-11-07
refactor: logging standardisation

james-milligan:
## This PR

- introduce logging service to wrap zap logs
- standardise logging pattern across services

### Related Issues

https://github.com/open-feature/flagd/issues/184


james-milligan:
This fails on benchmarking however I believe this is to be expected with the addition of the new logs

toddbaert:
> This fails on benchmarking however I believe this is to be expected with the addition of the new logs

@james-milligan @skyerus @AlexsJones @beeme1mr  this comment made me wonder:

Could we implement async logging? We want flagd to be as low-latency as possible - it hink it would be nice to make all the logging async. It looks like logrus doesn't support this out of the box, but there's a more-or-less standard solution:

https://github.com/sirupsen/logrus/issues/725

toddbaert:
@james-milligan this is a good start!

AlexsJones:
> > This fails on benchmarking however I believe this is to be expected with the addition of the new logs
> 
> @james-milligan @skyerus @AlexsJones @beeme1mr this comment made me wonder:
> 
> Could we implement async logging? We want flagd to be as low-latency as possible - I think it would be nice to make all the logging async. It looks like logrus doesn't support this out of the box, but there's a more-or-less standard solution:
> 
> [sirupsen/logrus#725](https://github.com/sirupsen/logrus/issues/725)

I am not really sure if it's going to be worth it as long as we can disable logging if required. I don't mind checking it out though.

AlexsJones:
Is this ready to review @james-milligan , should we just ignore the benchmark?

james-milligan:
> Is this ready to review @james-milligan , should we just ignore the benchmark?

yes its ready for revie, I think we should ignore the benchmark, it may even be worthwhile updating the benchmark to be more inline with the e2e tests so we can view performance impacts across the full flag evaluation process, not just across the span of the connect service.

If we do find that performance becomes an issue we can introduce an option to noop the logging service for specific loggers, e.g. silence request logs and only show configuration changes

pkg/eval/json_evaluator.go
```diff
@@ -29,6 +30,17 @@ const (
...
 )
 
+func NewJSONEvaluator(logger *logger.Logger) *JSONEvaluator {
+	ev := JSONEvaluator{
+		Logger: logger.WithFields(
+			zap.String("component", "evaluator"),
+			zap.String("evaluator", "json"),
+		),
+	}
+	jsonlogic.AddOperator("fractionalEvaluation", ev.fractionalEvaluation)
```
skyerus, line 40: Adding the operator multiple times is redundant I believe, I think that's why I put it in an `init` to begin with

james-milligan: I moved it here to provide it access to the same loggers as the rest of the packages - it seemed to have a specific function signature so i felt this was the cleanest way to handle it.
Unless im missing something, happy to refactor

pkg/logger/logger.go
```diff
@@ -0,0 +1,139 @@
...
+	l.Logger.Debug(msg, l.getFieldsForLog(reqID)...)
+}
+
+func (l *Logger) Debug(msg string, fields ...zap.Field) {
+	fields = append(fields, l.fields...)
+	l.Logger.Debug(msg, fields...)
+}
+
+func (l *Logger) InfoWithID(reqID string, msg string, fields ...zap.Field) {
+	l.Logger.Info(msg, l.getFieldsForLog(reqID)...)
```
skyerus, line 52: The `fields` passed into the func aren't used (same for other similar funcs)

pkg/logger/logger.go
```diff
@@ -0,0 +1,139 @@
...
+request fields pool.
+
+Example:
+
+WrappedLogger := NewLogger(myLogger)
+WrappedLogger.WriteFields("my-id", zap.String("foo", "bar"))
+WrappedLogger2 := WrappedLogger.WithFields(zap.String("ping", "pong"))
+
+WrappedLogger.DebugWithID("my-id", "my log line")
+	=> {"level":"debug","foo":"bar","msg":"my log line""}
```
skyerus, line 25: Should `"id":"my-id"` be present in these examples?

pkg/logger/logger.go
```diff
@@ -0,0 +1,139 @@
...
+		if ok {
+			res = r
+		}
+	}
+	return res
+}
+
+func (l *Logger) getFieldsForLog(reqID string) []zap.Field {
+	fields := l.getFields(reqID)
+	fields = append(fields, zap.String("requestID", reqID))
```
skyerus, line 108: constant for "requestID" perhaps

pkg/logger/logger.go
james-milligan: resolved

pkg/logger/logger.go
james-milligan: resolved

pkg/logger/logger.go
james-milligan: added constant

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -82,29 +82,33 @@ func (k *Sync) buildConfiguration() (*rest.Config, error) {
...
+		k.Logger.Error("No target feature flag configuration set")
 		return
 	}
 	if k.ProviderArgs[featureFlagNamespaceName] == "" {
-		k.Logger.Info("No target feature flag configuration namespace set")
+		k.Logger.Error("No target feature flag configuration namespace set")
 		return
 	}
-	k.Logger.Infof("Starting kubernetes sync notifier for resource %s", k.ProviderArgs["featureflagconfiguration"])
+	k.Logger.Info(
```
toddbaert, line 92: I think we are missing an equivalent log message for updates in the K8s sync.

In filepath, we have:

> `logger.Info("Filepath notifier event: %s %s", event.Name, event.Op.String())`

In HTTP we have:

> `logger.Info("http notifier event: %s has been modified", fs.URI)`

But we have no equivalent for the k8s sync.

pkg/logger/logger.go
```diff
@@ -0,0 +1,146 @@
...
+	=> {"level":"debug","requestID":"myID","foo":"bar","msg":"my log line""}
+
+WrappedLogger2.DebugWithID("myID", "my log line")
+	=> {"level":"debug","requestID":"myID","foo":"bar","ping":"pong","msg":"my log line""}
+
+WrappedLogger2.WriteFields("myID", zap.String("food", "bars"))
+
+WrappedLogger.DebugWithID("myID", "my log line")
+	=> {"level":"debug","requestID":"myID","foo":"bar","food":"bars","msg":"my log line""}
+*/
```
toddbaert, line 34: I like this tiny API you've created and I appreciate the doc.

toddbaert (approved):

cmd/start.go
```diff
@@ -71,15 +73,43 @@ var startCmd = &cobra.Command{
...
+		var level zapcore.Level
+		var err error
 		if Debug {
-			log.SetLevel(log.DebugLevel)
+			level = zapcore.DebugLevel
 		} else {
-			log.SetLevel(log.InfoLevel)
+			level = zapcore.InfoLevel
 		}
+		cfg := zap.Config{
```
Kavindu-Dodan, line 83: Minor thing - per my understanding, we mostly use default zap logger configurations, except for log level. Hence, what if `NewLogger(logger)` contract change to `NewLogger(zapcore.Level)` and internally perform the default loading ? We could of course expose `NewLogger(zap.Config)` if we really needs to :) IMO this makes `start.go` little cleaner.

pkg/logger/logger.go
```diff
@@ -0,0 +1,146 @@
...
+func (l *Logger) Fatal(msg string, fields ...zap.Field) {
+	fields = append(fields, l.fields...)
+	l.Logger.Debug(msg, fields...)
+}
+
+// WriteFields adds field key and value pairs to the highest level Logger, they will be applied to all
+// subsequent log calls using the matching requestID
+func (l *Logger) WriteFields(reqID string, fields ...zap.Field) {
+	res := append(l.getFields(reqID), fields...)
+	l.requestFields.Store(reqID, res)
```
skyerus, line 98: Are there any concerns with the performance of `sync.Map` with a high number of concurrent read/writes?

skyerus: If it is, I wonder whether we might benefit from https://pkg.go.dev/go.uber.org/zap#Logger.Check to avoid retrieving fields when unnecessary

james-milligan: While switching to a mutex may be more efficient performance wise I believe this to be a suitable use case for the `sync.Map` as per their documentation [here](https://pkg.go.dev/sync#Map). I believe we fit use case (2). 

That being said we can alter the structure of the logger to be more inline with the following:
```
type Logger struct {
	requestFields map[string]RequestFields
	Logger        *zap.Logger
	fields        []zap.Field
	mu            *sync.Mutex
}

type RequestFields struct {
	mu     *sync.Mutex
	fields []zap.Field
}
```
Top level mu for locking the map on initial write / delete then a mutex for the reading and writing of the request specific fields? Is this what you had in mind

cmd/start.go
james-milligan: Great suggestion, i'll introduce this change

pkg/logger/logger.go
skyerus: I agree that sync.Map is suited to the current pattern, I was just wondering whether the performance impact is large enough to consider an alternative pattern (perhaps passing a logger instance down the request stack).
Probably a conversation for another time if performance issues arise but I thought it worth raising at least.
I do think we'd benefit from using https://pkg.go.dev/go.uber.org/zap#Logger.Check to avoid retrieving fields unless strictly necessary.

cmd/start.go
james-milligan: Ive moved the generation of the logger into a separate function, this way we can still pass nil loggers into the NewLogger function, keeping both functions very simple

pkg/sync/kubernetes/kubernetes_sync.go
james-milligan: I have now introduced these logs

pkg/logger/logger.go
james-milligan: I have implemented `Logger.Check` to wrap the field fetching - I opted to avoid passing loggers through the request stack to keep interfaces clean across the different services, however I'm happy to refactor if we feel like its required for performance

toddbaert:
> > Is this ready to review @james-milligan , should we just ignore the benchmark?
> 
> yes its ready for revie, I think we should ignore the benchmark, it may even be worthwhile updating the benchmark to be more inline with the e2e tests so we can view performance impacts across the full flag evaluation process, not just across the span of the connect service.
> 
> If we do find that performance becomes an issue we can introduce an option to noop the logging service for specific loggers, e.g. silence request logs and only show configuration changes, which would also bypass the setting of values in the logger

I'm in favor of merging this and creating a new issue for a noop logger for evaluations that can be configured to still log update events.

toddbaert (approved):

beeme1mr (approved):

PR #220, opened 2022-11-23
fix: changed eventing configuration mutex to rwmutex and added missing lock

### LINKED ISSUE #219: [BUG] Missing lock in assignment of channel to an event stream request
### Observed behavior

Haven't observed a race condition but it could occur [here](https://github.com/open-feature/flagd/blob/main/pkg/service/connect_service.go#L147).

### Expected Behavior

No race conditions.

### Steps to reproduce

_No response_

skyerus:
## This PR

Changed eventing configuration mutex to rwmutex and added missing lock.

### Related Issues

Fixes #219

### DIFF
--- a/pkg/service/connect_service.go
+++ b/pkg/service/connect_service.go
@@ -45,7 +45,7 @@ type ConnectServiceConfiguration struct {
 }
 
 type eventingConfiguration struct {
-	mu   *sync.Mutex
+	mu   *sync.RWMutex
 	subs map[interface{}]chan Notification
 }
 
@@ -53,7 +53,7 @@ func (s *ConnectService) Serve(ctx context.Context, eval eval.IEvaluator) error 
 	s.Eval = eval
 	s.eventingConfiguration = &eventingConfiguration{
 		subs: make(map[interface{}]chan Notification),
-		mu:   &sync.Mutex{},
+		mu:   &sync.RWMutex{},
 	}
 	lis, err := s.setupServer()
 	if err != nil {
@@ -144,13 +144,16 @@ func (s *ConnectService) EventStream(
 	req *connect.Request[emptypb.Empty],
 	stream *connect.ServerStream[schemaV1.EventStreamResponse],
 ) error {
-	s.eventingConfiguration.subs[req] = make(chan Notification, 1)
+	requestNotificationChan := make(chan Notification, 1)
+	s.eventingConfiguration.mu.Lock()
+	s.eventingConfiguration.subs[req] = requestNotificationChan
+	s.eventingConfiguration.mu.Unlock()
 	defer func() {
 		s.eventingConfiguration.mu.Lock()
 		delete(s.eventingConfiguration.subs, req)
 		s.eventingConfiguration.mu.Unlock()
 	}()
-	s.eventingConfiguration.subs[req] <- Notification{
+	requestNotificationChan <- Notification{
 		Type: ProviderReady,
 	}
 	for {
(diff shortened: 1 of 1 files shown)


pkg/service/connect_service.go
```diff
@@ -181,11 +184,11 @@ func (s *ConnectService) EventStream(
 }
 
 func (s *ConnectService) Notify(n Notification) {
-	s.eventingConfiguration.mu.Lock()
+	s.eventingConfiguration.mu.RLock()
```
skyerus, line 187: Converted to RLock/RUnlock for efficiency.

pkg/service/connect_service.go
```diff
@@ -144,25 +144,29 @@ func (s *ConnectService) EventStream(
...
 	}()
-	s.eventingConfiguration.subs[req] <- Notification{
+	requestNotificationChan <- Notification{
 		Type: ProviderReady,
 	}
+	ticker := time.NewTicker(20 * time.Second)
 	for {
 		select {
-		case <-time.After(20 * time.Second):
+		case <-ticker.C:
```
skyerus, line 162: I assumed that this is meant to occur every 20 seconds but will remove if I'm wrong. @james-milligan can hopefully shine some light

james-milligan: I believe these have a similar outcome due to the wrapping for loop. Original intention was to keep this as a timeout, as we don need to send a keep alive if either the request is closed, or we send a notification (this is to prevent read/write timeouts)

james-milligan (approved):

beeme1mr (approved):

pkg/service/connect_service.go
```diff
@@ -181,11 +184,11 @@ func (s *ConnectService) EventStream(
 }
 
 func (s *ConnectService) Notify(n Notification) {
-	s.eventingConfiguration.mu.Lock()
+	s.eventingConfiguration.mu.RLock()
 	for _, send := range s.eventingConfiguration.subs {
 		send <- n
 	}
-	s.eventingConfiguration.mu.Unlock()
+	s.eventingConfiguration.mu.RUnlock()
```
toddbaert, line 191: Might be better to pull this up and defer it?

toddbaert (approved):

pkg/service/connect_service.go
skyerus: done

beeme1mr:
@AlexsJones, could you please review this when you have a moment?

pkg/service/connect_service.go
```diff
@@ -144,13 +144,16 @@ func (s *ConnectService) EventStream(
 	req *connect.Request[emptypb.Empty],
 	stream *connect.ServerStream[schemaV1.EventStreamResponse],
 ) error {
-	s.eventingConfiguration.subs[req] = make(chan Notification, 1)
+	requestNotificationChan := make(chan Notification, 1)
+	s.eventingConfiguration.mu.Lock()
```
AlexsJones, line 148: Can we defer this for nicer semantics?

AlexsJones:
I might be wrong here but
```
	requestNotificationChan := make(chan Notification, 1)
	s.eventingConfiguration.mu.Lock()
	s.eventingConfiguration.subs[req] = requestNotificationChan
	s.eventingConfiguration.mu.Unlock()
```

If you have two goroutines calling the `eventStream` function the second will have it's write queued. 
I think what it's missing is a check prior to the lock to see if `requestNotificationChan := make(chan Notification, 1)` has been set at all.

skyerus:
I'm not following the concern here, each invocation of `EventStream()` will lock while setting the channel to the service's state of ongoing connections.
> I think what it's missing is a check prior to the lock to see if requestNotificationChan := make(chan Notification, 1) has been set at all.

Each `EventStream()` has its own channel, in what scenario would `requestNotificationChan` have failed to have been set?

skyerus:
Is the concern that we're overwriting an already existing channel in the map?
The key is a pointer to the request structure, this is unique (at least until garbage collected).

AlexsJones:
I can't see any tangible benefits to this so I defer to others.

skyerus:
The benefit is that we're avoiding a potential race condition from writing to the same map concurrently.

Issue #217, opened 2022-11-22
Update FlagD to support multiple Kubernetes syncs


beeme1mr:

james-milligan:
resolved with [this PR](https://github.com/open-feature/flagd/pull/222), also adds behaviour for using multiple sync providers simultaneously

PR #222, opened 2022-11-23
feat!: start command flag refactor

james-milligan:

Corresponding OFO changes [here](https://github.com/open-feature/open-feature-operator/pull/256)

## This PR

- refactors the start command flags to remove `--sync-provider`
- multiple `--uri` flags can be passed indicating the use of different existing `sync-provider` types, all of which will work
- uses a prefix on the uri to define the `sync-provider`; `http(s)://` will be passed to the http sync, `file://` will be passed to the file path sync and the Kubernetes sync uses the following pattern `core.openfeature.dev/{namespace}/{name}`, this will also allow for the Kubernetes sync to watch multiple `FeatureFlagConfigurations` from different namespaces.
- adds deprecation warning when the `--sync-provider` flag is passed as an argument

`./flagd start --uri file://etc/flagd/end-to-end.json --uri core.openfeature.dev/test/end-to-end-2`

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/251


toddbaert:
This makes sense to me. I think watching multiple k8s-resources is sort of a must have, and this seems like a clean way to do it. I'll test this out a bit, but I like the approach.

pkg/eval/json_evaluator_model.go
```diff
@@ -40,6 +43,16 @@ func (f Flags) Merge(source string, ff Flags) (Flags, []StateChangeNotification)
 				FlagKey: k,
 			})
 		} else if !reflect.DeepEqual(val, v) {
+			if val.Source != source {
+				logger.Warn(
```
toddbaert, line 47: You may want to update the viper doc, for `--uri`, it kinda conflicts with this now.

docs/configuration.md
```diff
@@ -15,16 +15,26 @@ Supported flags are as follows (result of running `./flagd start --help`):
...
 
+### URI patterns
+
+Any URI passed to flagd via the `--uri` flag must follow one of the 3 following patterns to ensure that it is passed to the correct implementation: 
+
+| Sync      | Pattern | Example |
+| ----------- | ----------- | ----------- |
+| Kubernetes      | `core.openfeature.dev/namespace.name`       | `core.openfeature.dev/default.my-crd`       |
+| Filepath   | `file://path/to/my/flag`        | `file://etc/flagd/my-flags.json`       |
+| Remote   | `http(s)://flag-source-url`        | `https://my-flags.com/flags`       |
```
toddbaert, line 35:
```suggestion
| HTTP   | `http(s)://flag-source-url`        | `https://my-flags.com/flags`       |
```

toddbaert:
Tested this locally as well and everything works as expected, including updates, merging and warnings about duplicate flags across sources.. The experience is definitely improved.

pkg/eval/json_evaluator_model.go
james-milligan: updated viper doc

james-milligan:
https://github.com/open-feature/flagd/issues/217

docs/configuration.md
```diff
@@ -15,16 +15,26 @@ Supported flags are as follows (result of running `./flagd start --help`):
...
 Config file expects the keys to have the exact naming as the flags.
 
+### URI patterns
+
+Any URI passed to flagd via the `--uri` flag must follow one of the 3 following patterns to ensure that it is passed to the correct implementation: 
+
+| Sync      | Pattern | Example |
+| ----------- | ----------- | ----------- |
+| Kubernetes      | `core.openfeature.dev/namespace/name`       | `core.openfeature.dev/default/my-crd`       |
+| Filepath   | `file://path/to/my/flag`        | `file://etc/flagd/my-flags.json`       |
```
Kavindu-Dodan, line 34: Let's adopt [rfc8089](https://datatracker.ietf.org/doc/html/rfc8089) standard here (There are lot of examples in the spec [1]).

 And this will make `file:` as the standard prefix, optinally allowing 'file://' which is more or less the same you proposed. 


[1] https://datatracker.ietf.org/doc/html/rfc8089#appendix-A

docs/configuration.md
```diff
@@ -15,16 +15,31 @@ Supported flags are as follows (result of running `./flagd start --help`):
...
 Config file expects the keys to have the exact naming as the flags.
 
+### URI patterns
+
+Any URI passed to flagd via the `--uri` flag must follow one of the 3 following patterns to ensure that it is passed to the correct implementation: 
+
+| Sync      | Pattern | Example |
+| ----------- | ----------- | ----------- |
+| Kubernetes      | `core.openfeature.dev/namespace/name`       | `core.openfeature.dev/default/my-crd`       |
+<<<<<<< HEAD
```
skyerus, line 34: looks like some unresolved conflict has made it in here

pkg/runtime/from_config.go
```diff
@@ -59,34 +66,37 @@ func (r *Runtime) setEvaluatorFromConfig(logger *logger.Logger) error {
 }
 
 func (r *Runtime) setSyncImplFromConfig(logger *logger.Logger) error {
+	regCrd := regexp.MustCompile(crdRegex)
```
skyerus, line 69: General practice for these regexp compiles is to do them in an `init()` func but I'll leave that decision to you

james-milligan: I've moved these to an `init()` function, thanks for the suggestion

pkg/runtime/from_config.go
```diff
@@ -15,6 +16,24 @@ import (
...
+)
+
+var (
+	regCrd  *regexp.Regexp
+	regURL  *regexp.Regexp
+	regFile *regexp.Regexp
+)
+
+func init() {
+	regCrd = regexp.MustCompile(crdRegex)
```
skyerus, line 32: Nitpicking but it might suffice to just use the regex strings here directly rather than constants, I'll leave the decision to you

skyerus (approved):
LGTM, nice

toddbaert (approved):

Issue #212, opened 2022-11-18
[FEATURE] Flagd benchmarking


tcarrio:
### Requirements

I’ve heard memory usage cited for flagd previously in regards to the sidecar container pattern in the [OpenFeature Operator](https://github.com/open-feature/open-feature-operator). I would be interested in seeing a standard benchmark suite and docs around this.

### Adopters

For usage of the operator, one of the major cons listed in our research of OpenFeature was the memory usage per pod on the cluster and its potential impact on the cluster and our AWS bill. Regular benchmarking of the service would help understand what the expected resource usage of this would be means adopters can easily determine the impact.

### Regression Analysis

As change requests are merged, increases in memory or CPU requirements could change. Standard benchmarks, preferably automated, would help to identify regressions associated with specific changes and releases.

#### Regression Example

`HypotheticalPullRequest#892` increased memory usage by 20% and slowed processing of flags by 12%. This was flagged in CI and triaged. Upon further inspection, it was found that a struct was being copied to the heap on every execution of the new implementation.

## Sidecar Impact Examples

Memory usage increase for the pods directly translate into resource pricing increases. This can be fairly negligible depending on the architecture of your applications. The table below demonstrates your pods provisioned resources in memory, along with a few columns indicating what the typical memory usage is for `flagd`.

|Pod Memory|Flagd 1MB|Flagd 4MB|Flagd 16MB|
|----------|---------|---------|----------|
|16MB      |6.25%    |25%      |100%      |
|32MB      |3.125%   |12.5%    |50%       |
|64MB      |1.5625%  |6.25%    |25%       |
|128MB     |0.78125% |3.125%   |12.5%     |
|256MB     |0.390625%|1.5625%  |6.25%     |
|512MB     |0.195312%|0.78125% |3.125%    |
|1024MB    |0.097656%|0.390625%|1.5625%   |
|2048MB    |0.048828%|0.195312%|0.78125%  |
|4096MB    |0.024414%|0.097656%|0.390625% |

### Disclaimer

:warning: These are __examples__, not based on __any existing benchmark__, so **do not use this for actually estimating resource utilization**. I am sampling pseudorandom resource usage as a way to statistically showcase the impact.

beeme1mr:
There's is a basic benchmark running already on each PR as well as nightly.

- https://github.com/open-feature/flagd/blob/main/.github/workflows/build.yaml#L14
- https://open-feature.github.io/flagd/

[...]

PR #225, opened 2022-11-25
test: load testing tooling and initial observations 

Kavindu-Dodan:
## This PR
PR introduce tools to support performance testing for flagd. Content include,

- A random feature flag generator
- Sample [K6](https://k6.io/) script 
- [go pprof](https://pkg.go.dev/net/http/pprof) enabled flagd build with a dedicated Dockerfile (`profile.Dockerfile`)
- Initial performance observations 

### Related Issues
PR is related to #212 

### How to test
Consider the readme contents at `/test/loadtest/README.MD`


loadtest/ff_gen.go
```diff
@@ -0,0 +1,108 @@
...
+		fmt.Printf("Invalid type %s. Falling back to default %s", flagType, BOOL)
+		flagType = BOOL
+	}
+
+	root := Flags{}
+	root.Flags = make(map[string]Flag)
+
+	switch flagType {
+	case BOOL:
+		root.setStringFlags(flagCount)
```
tcarrio, line 50: Without diving too far into this myself, these two cases seem reversed. Bool sets string, strings sets bool. Is that correct?

beeme1mr (commented):
How was the `loadTestResults.png` created?

loadtest/ff_gen.go
Kavindu-Dodan: @tcarrio  thanks. yes this is a mix in the final refactoring (extracting constants to consts). Fixed with the latest commit.

Kavindu-Dodan:
> How was the `loadTestResults.png` created?

This is an extraction from the load test results [1]. If you are interested in how numbers were derived, it was by running a containerised flagd with random FFs (generated with `ff_gen.go`) and using the K6 script (included in the PR). Final result is the average of 3 test runs. And there was a warm-up time provided to flagd. 

[1] https://docs.google.com/spreadsheets/d/1YLtkA-WryJXaZ4GWrbKhT_Izx3dh4spS/edit?usp=sharing&ouid=109303231969053490300&rtpof=true&sd=true

loadtest/ff_gen.go
Kavindu-Dodan: Further, this does not have any impact on the results as results were obtained from the script version preset at 07f833b56eacd0870099e53b5f07b201478d8625

beeme1mr (approved):

toddbaert:
I think you should note somewhere that HTTP is used as the interface. Based on what I'm seeing, I think encoding and (de)serialization is probably where we are spending a lot of time. I think it would be ideal to run a suite in gRPC as well, using https://github.com/grpc/grpc/blob/master/doc/command_line_tool.md or https://github.com/fullstorydev/grpcurl, potentially. Do you think we should create an issue for this? Should we do it as part of this issue? 

cc @beeme1mr

Kavindu-Dodan:
> I think you should note somewhere that HTTP is used as the interface. Based on what I'm seeing, I think encoding and (de)serialization is probably where we are spending a lot of time. I think it would be ideal to run a suite in gRPC as well, using https://github.com/grpc/grpc/blob/master/doc/command_line_tool.md or https://github.com/fullstorydev/grpcurl, potentially. If we do this we'll want to make sure the same connection is used for all the RPCs. Do you think we should create an issue for this? Should we do it as part of this issue?
> 
> cc @beeme1mr

Great input. I highlighted the API type used in the observations section.

Yes, testing and comparing the gRPC would be great. And yes, pprof proved the same point: most of the cpu time and memory is spent on reading and writing request

loadtest/README.MD
```diff
@@ -0,0 +1,32 @@
+## Load Testing
+
```
AlexsJones, line 2: Can we add load testing under a new `tests` directory? Would help us future proof more root level test folders

loadtest/README.MD
```diff
@@ -0,0 +1,32 @@
...
+To enable pprof profiling, build docker image with build argument `GOBUILD` set to `profile`
+
+ex:- `docker build --build-arg GOBUILD=profile . -t flagdprofile`
+
+This image now exposes port `6060` for pprof data. 
+
+### Performance observations
+
+flagd performs well under heavy loads. Consider the following results observed against the HTTP API of the flagd,
```
skyerus, line 21:
```suggestion
flagd performs well under heavy loads. Consider the following results observed against the HTTP API of flagd,
```

Dockerfile
```diff
@@ -7,6 +7,7 @@ ARG TARGETARCH
 ARG VERSION
 ARG COMMIT
 ARG DATE
+ARG GOBUILD
```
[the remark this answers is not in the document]
Kavindu-Dodan: Do you mean to create a dedicated Dockerfile where we enable the profiler by default 🤔

loadtest/README.MD
```diff
@@ -0,0 +1,32 @@
...
+- sample_k6.js : sample K6 load test script
+
+### Profiling
+
+It's possible to utilize `profiler.go` included with flagd source to profile flagd during
+load test. Profiling is enabled through [go pprof package](https://pkg.go.dev/net/http/pprof).
+
+To enable pprof profiling, build docker image with build argument `GOBUILD` set to `profile`
+
+ex:- `docker build --build-arg GOBUILD=profile . -t flagdprofile`
```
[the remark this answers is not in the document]
Kavindu-Dodan: I added a section with an example test run with current resources. This should help with future contributions.

loadtest/README.MD
Kavindu-Dodan: Restructured in the latest commit :)

Dockerfile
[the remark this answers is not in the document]
Kavindu-Dodan: From what I read [1] the impact of ARG is minimal especially if it's not changing. And introducing a dedicated dockerfile would add more maintenance burden. 

Do you see any other way to improve this or is this fine ? :) 

[1] - https://docs.docker.com/engine/reference/builder/#impact-on-build-caching

beeme1mr:
@Kavindu-Dodan please resolve any conversations that have been addressed.

Kavindu-Dodan:
> @Kavindu-Dodan please resolve any conversations that have been addressed.

ACK. There's one more conversation to resolve and I am waiting for the response from @AlexsJones .

Dockerfile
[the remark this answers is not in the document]
AlexsJones: Yes I am talking about a profile without a switch in it

AlexsJones:
> > @Kavindu-Dodan please resolve any conversations that have been addressed.
> 
> ACK. There's one more conversation to resolve and I am waiting for the response from @AlexsJones .

Sorry just remarked on it

Dockerfile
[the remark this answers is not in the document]
Kavindu-Dodan: @AlexsJones Thanks, I added the suggested change and hope you can have another look at the PR.

profiler.go
```diff
@@ -0,0 +1,19 @@
+//go:build profile
```
AlexsJones, line 1: Can you please move this out of root, somewhere like `pkg/utils` ?

AlexsJones (changes requested):

profiler.go
Kavindu-Dodan: This will break the tag-based profiler enabling (ex:- local build with `go build -tags profile -o flagd` to produce flagd with profile). The reason is the usage of package "main" here.

Kavindu-Dodan: @AlexsJones  if this is not a strong preference, I will go ahead and resolve the conversation :)

Kavindu-Dodan: @james-milligan  yes that's correct. I thought about these options but they have many disadvantages compared to the current one. IMO, this usage is clean and confirms with use cases discussed by others [1]

[1] https://github.com/golang/go/issues/33389

skyerus (approved):

PR #226, opened 2022-11-28
feat: enable request logging via the --debug flag

james-milligan:
## This PR

- The `--debug` flag now also controls the logging for requests, when set all logging levels are enabled for the `XXXWithID` `logger` methods and allows the setting of request fields, allowing for improved performance when logs are not required.
- `NewLogger` now takes an additional boolean argument to set the internal `reqIDLogging` field, this field is also set to false when the `*zap.Logger` argument is nil

### Follow-up Tasks

This flag should be set by default in the operator
https://github.com/open-feature/open-feature-operator/pull/260


pkg/logger/logger.go
```diff
@@ -57,6 +61,9 @@ func (l *Logger) Debug(msg string, fields ...zap.Field) {
 }
 
 func (l *Logger) InfoWithID(reqID string, msg string, fields ...zap.Field) {
+	if l.noopReqIDLogging {
+		return
+	}
```
toddbaert, line 66: I think you possibly avoid these duplicated early returns by defining a new internal interface which has these `...WithID` functions and encapsulating that within the logger. One implantation would run the `Check` function and the other would no-op, but I'm not sure if it's worth it, honestly, so I'll leave it to you. The value might be too low.

toddbaert (approved):

AlexsJones:
To disable logging I have to pass `--disable-request-logging` that's horrible DX :(

cmd/start.go
```diff
@@ -52,6 +53,7 @@ func init() {
 	flags.StringP(
 		bearerTokenFlagName, "b", "", "Set a bearer token to use for remote sync")
 	flags.StringSliceP(corsFlagName, "C", []string{}, "CORS allowed origins, * will allow all origins")
+	flags.Bool(NoopRequestLoggingFlagName, false, "Disable all logging levels pertaining to flag evaluation requests")
```
AlexsJones, line 56: Can we use something a bit more common such as `debug` and when that's enabled logs are on?
We then use a `v` verbosity flag like programs e.g. kubectl?

```
flagd start -f ... --debug --v=5
```

pkg/logger/logger.go
AlexsJones: I would agree that we could interface the logger out and have a shim for no-ops.
I do personally think this is a bit over the top though.

james-milligan:
> To disable logging I have to pass `--disable-request-logging` that's horrible DX :( 

We could have the request logging disabled by default and only enable it with the --debug flag?

AlexsJones:
> > To disable logging I have to pass `--disable-request-logging` that's horrible DX :(
> 
> We could have the request logging disabled by default and only enable it with the --debug flag?

Sounds much better to me

cmd/start.go
```diff
@@ -13,17 +13,18 @@ import (
...
+	socketPathFlagName         = "socket-path"
+	syncProviderFlagName       = "sync-provider"
+	providerArgsFlagName       = "sync-provider-args"
+	evaluatorFlagName          = "evaluator"
+	serverCertPathFlagName     = "server-cert-path"
+	serverKeyPathFlagName      = "server-key-path"
+	uriFlagName                = "uri"
+	bearerTokenFlagName        = "bearer-token"
+	corsFlagName               = "cors-origin"
+	NoopRequestLoggingFlagName = "disable-request-logging"
```
skyerus, line 27: can we lowercase this for consistency?

james-milligan: done

james-milligan:
> Sounds much better to me

I have introduced this change

toddbaert:
> > > To disable logging I have to pass `--disable-request-logging` that's horrible DX :(
> > 
> > 
> > We could have the request logging disabled by default and only enable it with the --debug flag?
> 
> Sounds much better to me

This makes sense to me as well.

pkg/logger/logger.go
```diff
@@ -170,14 +192,18 @@ func NewZapLogger(level zapcore.Level) (*zap.Logger, error) {
 }
 
 // NewLogger returns the logging wrapper for a given *zap.logger,
```
skyerus, line 194:
```suggestion
// NewLogger returns the logging wrapper for a given *zap.logger.
```

pkg/logger/logger.go
```diff
@@ -170,14 +192,18 @@ func NewZapLogger(level zapcore.Level) (*zap.Logger, error) {
...
-// will return a wrapped zap noop logger if none is provided
-func NewLogger(logger *zap.Logger) *Logger {
+// Noop logger will bypass the setting of fields, improving performance
```
skyerus, line 195:
```suggestion
// Noop logger bypasses the setting of fields, improving performance.
```

pkg/logger/logger.go
```diff
@@ -170,14 +192,18 @@ func NewZapLogger(level zapcore.Level) (*zap.Logger, error) {
...
+// if no *zap.Logger is nil, a noop logger will be set,
```
skyerus, line 196:
```suggestion
// If *zap.Logger is nil a noop logger is set
```

pkg/logger/logger.go
```diff
@@ -170,14 +192,18 @@ func NewZapLogger(level zapcore.Level) (*zap.Logger, error) {
...
+// and the noopReqIDLogging argument will be overwriten to true
```
skyerus, line 197:
```suggestion
// and the noopReqIDLogging argument is overwritten to true
```

AlexsJones:
Is this ready for review now?

james-milligan:
> Is this ready for review now?

yes :)

beeme1mr:
@james-milligan please resolve conversations that have been addressed.

AlexsJones (approved):

AlexsJones (approved):

Issue #234, opened 2022-12-16
[FEATURE] flagd bulk evaluation


toddbaert:
### Requirements

In order to support synchronous evaluation (something particularly important for [client use cases](https://github.com/open-feature/spec/issues/167)), we need a bulk evaluation endpoint in flagd. The endpoint would accept a `context` and evaluate all configured flags with that `context`. The returned result should be a map of `flagKey: resolutionDetails`.

- [x] update schema (https://github.com/open-feature/schemas/issues/63)
- [x] implement function
- [ ] add tests
- [x] add documentation

Example curl:

```shell
curl -X POST "localhost:8013/schema.v1.Service/ResolveAll" -d '{"context":{"email": "foo@bar.com"}}' -H "Content-Type: application/json"

{
   "flag1":{
      "value":"#0000FF",
      "reason":"TARGETING_MATCH",
      "variant":"blue"
   },
   "flag2":{
      "value":false,
      "reason":"STATIC",
      "variant":"on"
   }
}

```

Things to consider:

- possible enhancements in the future to allow an array of flag keys to evaluate, instead of all
- should we include additional metadata such as flag type to aid in parsing?

beeme1mr:
Depends on https://github.com/open-feature/open-feature-operator/issues/291

PR #239, opened 2022-12-21
feat: ResolveAll endpoint for bulk evaluation

james-milligan:
## This PR

- introduces the resolve all endpoint for bulk flag evaluation

### Related Issues

https://github.com/open-feature/flagd/issues/234


james-milligan:
LINTING FIXES:
Added the `//nolint:funlen` annotation to the Notify function of `Kubernetes_sync.go`. To drop the functions lines by 2 we will need to abstract out functionality that is used once, or drop comments. Both options will make the code more confusing for developers.

pkg/eval/json_evaluator.go
```diff
@@ -103,6 +103,68 @@ func resolve[T constraints](reqID string, key string, context *structpb.Struct,
...
+				context,
+				je.evaluateVariant,
+				je.state.Flags[flagKey].Variants,
+			)
+			if err != nil {
+				je.Logger.WarnWithID(reqID, fmt.Sprintf("Bulk evaluation: key %s returned error %s", flagKey, err.Error()))
+				continue
+			}
+			values = append(values, NewAnyValue(value, variant, reason, flagKey))
+		case string:
```
skyerus, line 124: Can we simplify these cases somehow to avoid duplicating the logic?

toddbaert:
I tested this and it works as expected. I agree with @skyerus that it might be nice to reduce some duplication, but that may not be possible.

skyerus (approved):

pkg/eval/json_evaluator.go
```diff
@@ -103,6 +103,57 @@ func resolve[T constraints](reqID string, key string, context *structpb.Struct,
...
+			value, variant, reason, err = resolve[map[string]any](
+				reqID,
+				flagKey,
+				context,
+				je.evaluateVariant,
+				je.state.Flags[flagKey].Variants,
+			)
+		}
+		if err != nil {
+			je.Logger.WarnWithID(reqID, fmt.Sprintf("Bulk evaluation: key %s returned error %s", flagKey, err.Error()))
```
skyerus, line 149: Why is this only a warn?

pkg/service/connect_service.go
```diff
@@ -138,6 +138,61 @@ func (s *ConnectService) setupServer() (net.Listener, error) {
...
+				Reason:  value.Reason,
+				Variant: value.Variant,
+				Value: &schemaV1.AnyFlag_DoubleValue{
+					DoubleValue: v,
+				},
+			}
+		case map[string]any:
+			val, err := structpb.NewStruct(v)
+			if err != nil {
+				s.Logger.WarnWithID(reqID, fmt.Sprintf("struct response construction: %v", err))
```
skyerus, line 180: Why not error level?

pkg/eval/json_evaluator.go
james-milligan: resolved

toddbaert (approved):

Issue #244, opened 2023-01-04
[BUG] flagd (`transposeEvaluators` fn) doesn't understand json without indentations


vadasambar:
### Observed behavior
[`transposeEvaluators`](https://github.com/open-feature/flagd/blob/cef8463a7aff8f48dbc049dcaab86177c3aadd08/pkg/eval/json_evaluator.go#L207-L235) doesn't behave correctly if the json passed to it doesn't have the right indentation.

### Expected Behavior
[`transposeEvaluators`](https://github.com/open-feature/flagd/blob/cef8463a7aff8f48dbc049dcaab86177c3aadd08/pkg/eval/json_evaluator.go#L207-L235) should handle cases where the json passed to it doesn't have the right indentation.

### Steps to reproduce

1.  Check out [this](https://github.com/open-feature/flagd/pull/206/files) PR's branch
2. Put `fmt.Println` statements just after [this](https://github.com/open-feature/flagd/blob/cef8463a7aff8f48dbc049dcaab86177c3aadd08/pkg/eval/json_evaluator.go#L225) line and after [this](https://github.com/open-feature/flagd/blob/cef8463a7aff8f48dbc049dcaab86177c3aadd08/pkg/eval/json_evaluator.go#L229) line.
3. Use `json.Marshal` instead of `json.MarshalIndent` [here](https://github.com/open-feature/flagd/pull/206/files#diff-50331e813809741b1189e199f877dfb890e721ff79bfe0acaa535ec7966ea3b1R118)
4. From `flagd` root directory run (example_flags.yaml can be found [here](https://github.com/open-feature/flagd/pull/206/files#diff-cf62f044b3f87a6f9cfef61cc16571c557de628a867b1d2cc964a70ce2e50279)):
```
$ go run main.go  start -f file:./config/samples/example_flags.yaml -e=yaml 
{"level":"info","ts":1672807996.7619019,"caller":"service/connect_service.go:108","msg":"metrics listening at 8014","component":"service"}
evalValue before {"in":["@faas.com",{"var":["email"]}]}
evalValue after "in":["@faas.com",{"var":["email"]}
[...]

PR #250, opened 2023-01-10
fix: replace character slice with regex replace

james-milligan:

## This PR

- fixes transpose evaluator state parsing when state contains trailing characters 

### Related Issues

https://github.com/open-feature/flagd/issues/244

### Notes

switches from a hard coded `1:n-2` slice to use regex to match and replace the first `{` and last `}` in the string.


james-milligan:
These changes also mean that we no longer need to `json.MarshalIndent` anymore (I believe), happy to replace with `json.Marshal` provided im not missing another reason for its use?

cc @vadasambar

toddbaert (approved):

beeme1mr (approved):

Issue #240, opened 2022-12-29
[FEATURE] Detect flag configuration format from file extension


Kavindu-Dodan:
### Requirements

With #206, flagd now has the ability to support **yaml** formatted flag configurations.

However, to use yaml evaluation, users should use `-e, --evaluator string` set to **.yaml** or **.yml** even if the flag configuration file contains the correct extension.

To enhance the user experience, flag format can be detected from the file extension and omit the requirement of evaluator string parameter.

This simplifies, `flagd start -e yaml -f example_flags.yaml` command to `flagd start -f example_flags.yaml`.

PR #257, opened 2023-01-10
feat: file extension detection

james-milligan:

## This PR

- file type is detected via the file extension, it is now possible to mix and match `json` and `yaml` files as flag configuration sources 

```
start -f file:./config/samples/example_flags.yaml -f file:./config/samples/example_flags.json
```

### Related Issues

https://github.com/open-feature/flagd/issues/240

### Follow-up Tasks

It is my opinion that the `-e` or `--evaluator` flag is now a little misleading / confusing, as only the `json_evaluator` is used internally, and changing the flag value has no effect internally. Open to opinions on how this should be handled, if at all.


toddbaert:
> It is my opinion that the -e or --evaluator flag is not a little misleading / confusing, as only the json_evaluator is used internally, and changing the flag value has no effect internally. Open to opinions on how this should be handled, if at all.

I agree with this, but I'm not sure what the best approach is.

toddbaert (approved):

Kavindu-Dodan (approved):

Kavindu-Dodan:
> > It is my opinion that the -e or --evaluator flag is not a little misleading / confusing, as only the json_evaluator is used internally, and changing the flag value has no effect internally. Open to opinions on how this should be handled, if at all.
> 
> I agree with this, but I'm not sure what the best approach is.

We only have one evaluator type and that's JSON (shared with yaml handling). So if we do not have any intention to introduce another evaluator, we can remove this option.

beeme1mr:
> We only have one evaluator type and that's JSON (shared with yaml handling). So if we do not have any intention to introduce another evaluator, we can remove this option.

Can we think of any future scenario where we would want this? If not, I think it makes sense to deprecate this option.

james-milligan:
> > We only have one evaluator type and that's JSON (shared with yaml handling). So if we do not have any intention to introduce another evaluator, we can remove this option.
> 
> Can we think of any future scenario where we would want this? If not, I think it makes sense to deprecate this option.

Ive just pushed the changes to deprecate the `-e`/`--evaluator` flag, a warning is logged when it is used

skyerus (approved):

beeme1mr (approved):

PR #291, opened 2023-01-17
refactor: refactor sync providers

Kavindu-Dodan:
## This PR

Revisits sync providers and refactor them.

- unify event handling
- packaging improvements
- removal of unused codes (cleanup)
- clean contracts between Runtime and Sync implementations 

With this change, we get a clear isolation between runtime and sync providers (ex:- file, k8s, ....). For example, the runtime is not aware of any sync implementation details such as events. It simply coordinates sync impls, store (evaluator) and notifications. This should bring more extensibility when adding future extensions. 

Below is the overview of the internals and interactions,

image


james-milligan:
This looks like a great improvement to the current sync / eval implementation. 

Ill test this properly once it is ready for review

pkg/runtime/runtime.go
```diff
@@ -36,48 +40,63 @@ type Config struct {
 	CORS []string
 }
 
-func (r *Runtime) startSyncer(ctx context.Context, syncr sync.ISync) error {
-	if err := r.updateState(ctx, syncr); err != nil {
-		return err
+func (r *Runtime) Start() error {
```
skyerus, line 43: This is clean, nice!

pkg/sync/file/filepath_sync.go
```diff
@@ -0,0 +1,140 @@
...
+				}
+			}
+			fs.Logger.Info(fmt.Sprintf("Filepath event: %s %s", event.Name, event.Op.String()))
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return errors.New("watcher error")
+			}
+			fs.Logger.Error(err.Error())
+		case <-ctx.Done():
+			fs.Logger.Debug("exiting file watcher")
```
skyerus, line 96: We should aim for consistency on whether capitalizing or not logs

pkg/sync/file/filepath_sync.go
```diff
@@ -0,0 +1,140 @@
...
+	default:
+		return "", fmt.Errorf("filepath extension for URI '%s' is not supported", fs.URI)
+	}
+}
+
+// yamlToJSON is a generic helper function to convert
+// yaml to json
+func yamlToJSON(rawFile []byte) (string, error) {
+	var ms map[string]interface{}
+	// yaml.Unmarshal unmarshals to map[interface]interface{}
```
skyerus, line 129: `var ms map[string]interface{}` implies otherwise

pkg/sync/http/http_sync.go
```diff
@@ -0,0 +1,140 @@
...
+	AddFunc(spec string, cmd func()) error
+	Start()
+	Stop()
+}
+
+func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
+	// Initial fetch
+	fetch, err := hs.Fetch(ctx)
+	if err != nil {
+		hs.Logger.Error(err.Error())
```
skyerus, line 43: Is returning an error after handling it (with a log) the right thing to do?

pkg/sync/http/http_sync.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+
+	dataSync <- sync.DataSync{FlagData: fetch, Source: hs.URI}
+
+	_ = hs.Cron.AddFunc("*/5 * * * *", func() {
```
skyerus, line 49: This ought to be configurable

pkg/sync/http/http_sync.go
```diff
@@ -0,0 +1,140 @@
...
+	if hs.BearerToken != "" {
+		bearer := "Bearer " + hs.BearerToken
+		req.Header.Set("Authorization", bearer)
+	}
+
+	resp, err := hs.Client.Do(req)
+	if err != nil {
+		return []byte(""), err
+	}
+	defer func() { _ = resp.Body.Close() }()
```
skyerus, line 110: We shouldn't ignore an error

pkg/sync/file/filepath_sync.go
Kavindu-Dodan: Yes, I will revisit logs and try to unify them

pkg/sync/http/http_sync.go
```diff
@@ -0,0 +1,140 @@
...
+	<-ctx.Done()
+	hs.Cron.Stop()
+
+	return nil
+}
+
+func (hs *Sync) fetchBodyFromURL(ctx context.Context, url string) ([]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, "GET", url, bytes.NewBuffer(nil))
+	if err != nil {
+		return []byte(""), err
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is logic as it is from the old implementation. I will see if we rely on non nil value here.

pkg/sync/http/http_sync.go
Kavindu-Dodan: Right now, `Runtime` setup an error group and adds sync impl to routines. The initial fetch on the other hand makes sure our source is correct. And if fetch fails, then we return this error essentially to the error group so that it could exit [1]. I added log here for clarity, so that we do not need explicit error handing at `Ruintime` and rely on other logs from error group internals. 

I am open for improvements here

[1] - https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Go

pkg/sync/http/http_sync.go
Kavindu-Dodan: I will create a follow up issue :)

pkg/sync/file/filepath_sync.go
Kavindu-Dodan: yes, this is from existing logic and I did not touch this

pkg/sync/http/http_sync.go
Kavindu-Dodan: Created issue #293 . IMO, this is a "good first issue" candidate.

pkg/sync/kubernetes/ffclientset.go
```diff
@@ -1,113 +0,0 @@
-package kubernetes
```
Kavindu-Dodan, line 1: This was not used hence I opted for a cleanup.

@james-milligan @skyerus fyi

pkg/sync/http/http_sync.go
Kavindu-Dodan: Sure, I updated to not to log with the latest commit

pkg/runtime/runtime.go
```diff
@@ -36,48 +40,63 @@ type Config struct {
...
+// updateWithNotify helps to update state and notify listeners
+func (r *Runtime) updateWithNotify(data sync.DataSync) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
-	notifications, err := r.Evaluator.SetState(syncr.Source(), msg)
+
+	notifications, err := r.Evaluator.SetState(data.Source, data.FlagData)
 	if err != nil {
-		return fmt.Errorf("set state: %w", err)
+		r.Logger.Error(err.Error())
```
skyerus, line 98: it is perhaps useful to log including the `fmt.Errorf("set state: %w", err)` that was there before for additional context

pkg/sync/http/http_sync_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/open-feature/flagd/pkg/logger"
+	syncmock "github.com/open-feature/flagd/pkg/sync/http/mock"
+)
+
+func TestSimpleSync(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	t.Run("SimpleSync", func(t *testing.T) {
```
skyerus, line 21: Is there a reason why this is executed in a subtest? Apologies if you haven't introduced this

pkg/runtime/runtime.go
james-milligan: This error log should also contain a stack trace, so its no longer necessary, seeing as the log can more easily be traced back to this line

pkg/sync/http/http_sync_test.go
Kavindu-Dodan: Thank you for catching this. I will convert this to a simple test case.  Same is true for file sync test

pkg/runtime/runtime.go
Kavindu-Dodan: If no hard opinion here, I will keep this as it is :)

skyerus: If there is now a stack trace then happy for it to remain as is :)

skyerus (approved):

pkg/sync/file/filepath_sync.go
```diff
@@ -0,0 +1,138 @@
...
+	URI          string
+	Logger       *logger.Logger
+	ProviderArgs sync.ProviderArgs
+	// FileType indicates the file type e.g., json, yaml/yml etc.,
+	fileType string
+}
+
+//nolint:funlen
+func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
+	fs.Logger.Info("Starting filepath sync notifier")
```
toddbaert, line 29: Let's use lower case in all log statments.

toddbaert:
`go mod tidy` leaves some deltas for me.

toddbaert (approved):
This looks really good to me. We've added to flagd a number of times, and whenever that happens it usually creates need for a refactor to clean things up. I find the new interfaces much clearer! Thanks!

pkg/sync/file/filepath_sync.go
Kavindu-Dodan: I think this fits into a ticket rather than this PR. Flagd logs usually start with uppercase characters. This is why I used uppercase here.

Kavindu-Dodan: I created a follow up to focus on full soruce #298. Another good first task for a contributor :)

james-milligan (approved):

PR #297, opened 2023-01-19
feat: introduce grpc sync for flagd

### LINKED ISSUE #249: Real-time Flag Updates Beyond Kubernetes
## Problem

Flagd supports various sync options include watching a local file, HTTP polling, and subscribing to the Kubernetes API. In most configurations, the Kubernetes API is the only option that currently supports near real-time flag configuration updates. This introduces some challenges in Kubernetes clusters that do not allow direct access to the API and environments that are not using Kubernetes at all. A more flexible sync option that supports real-time flag updates would help overcome these challenges.

## Requirements

- [x] Flag changes are reflected in near real-time in flagd
- [x] The new real-time flagd sync does not rely on the Kubernetes API
- [x] More robust testing of TLS support in CI

## Next steps

- [Real-time Flag Update via gRPC in Kubernetes](https://github.com/orgs/open-feature/projects/8/views/1)

Kavindu-Dodan:
## This PR

Attempts to resolve #249 by introducing grpc sync provider to flagd.

OFEP [approved] - https://github.com/open-feature/ofep/blob/main/OFEP-flagd-grpc-sync.md

### How to test/run ?

Flagd acts as the grpc client, hence you need at least a minimal mock server. For this, you can utilize this [1] server implementation. 

Startup arguments of flagd now support grpc target uri. This can be provided with `grpc://` , for example, 

`./flagd start --uri grpc://localhost:8090`

### Technical highlights

- GRPC protobuf definitions are available in buf [2] and are backed by the schema repository (https://github.com/open-feature/schemas) 
- Initial connection must be successful (i.e- grpc server/target must be accepting connections)
- Subsequent server connection losses will not result in a runtime failure and connection re-establishment attempts will be performed

### What is not included (follow up improvements)

-  Connection security: This version does not enforce connection security. This will be addressed with follow-up improvements (ex:- TLS enabled connections). Hence, strongly recommends not using this version in production scenarios (fixed by #398)
- Server implementations: This sync provider was designed to be open and connects to any server implementation. Hence there is no default server implementation. You may create your own server implementation based on grpc schemas.

[1] - https://github.com/Kavindu-Dodan/flagd-grpc-sync-server
[2] - https://buf.build/open-feature/flagd


xvzf:
That's amazing!

toddbaert:
I'm going to give this a try and a thorough review on Monday!

toddbaert:
Works for me! I couple things I noted in testing:

- There's no way to change flag values in server, is there? I know this is just a demo, so that makes sense, but I guess without that there's no way to test that the grpc sync gets updates appropriately? Or am I missing somthing?
- I see `1.6757003525209875e+09  warn    grpc/grpc_sync.go:111   receivied unknown state: 41     {"component": "sync", "sync": "grpc"}` in logs - not sure, but this could be a gRPC keepalive - we might not want to log every message received as a warning.
- If the server does, the connection fails and the client flagd crashes. I think we need to add elegant re-connection logic here and make this pretty resilient.

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		g.Logger.Error(fmt.Sprintf("Error establishing connection: %s", err.Error()))
+		return err
+	}
+
+	return g.streamListener(ctx, dial, dataSync)
+}
+
+// streamListener performs the grpc listening on provided client connection and push updates through dataSync channel
+func (g *Sync) streamListener(ctx context.Context, dial *grpc.ClientConn, dataSync chan<- sync.DataSync) error {
```
toddbaert, line 41: I think we will need stream-reconnect logic.

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,126 @@
...
+				FlagData: data.Flags,
+				Source:   g.Target,
+				Type:     sync.DELETE,
+			}
+
+			g.Logger.Debug("received a delete payload")
+		case v1.SyncState_SYNC_STATE_PING:
+			g.Logger.Debug("received server ping")
+		default:
+			g.Logger.Warn(fmt.Sprintf("receivied unknown state: %s", data.State.String()))
```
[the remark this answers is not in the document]
Kavindu-Dodan: Yes, this was a test case to check the state handling. I should have actually used `42` 😉 

I am also now 100% sure about the log level here. I opted for warning as this status is unknow and could happen because of an incorrect server impl / or any other unknown. IMO it was better to log a warn than a debug (compared for example to a server ping)

I am open for suggestions here :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Yes totally agree. This is why we have a simple logic at isync contract method [1]. I kept it simple, so we can build the retry logic there. 

However, I am not confident to include try logic and related complexity with this initial version. Because that require unit tests, and may be some more configurations (making this review extensive). Should I create a follow up issue ? Or do you think we need this withthis PR :) ? 


[1] - https://github.com/open-feature/flagd/pull/297/files#diff-5bd4f361280cd88498f36d821f08f603ba13af72ad0c1ce0313d7b3fa1ed5af5R30-R37

Kavindu-Dodan:
> Works for me! I couple things I noted in testing:
> 
> * There's no way to change flag values in server, is there? I know this is just a toy demo server, so that makes sense, but I guess without that there's no way to test that the grpc sync gets updates appropriately? Or am I missing something in regards to how to test that?
> * I see `1.6757003525209875e+09  warn    grpc/grpc_sync.go:111   receivied unknown state: 41     {"component": "sync", "sync": "grpc"}` in logs - not sure, but this could be a gRPC keepalive - we might not want to log every message received as a warning.
> * If the server does, the connection fails and the client flagd crashes. I think we need to add elegant re-connection logic here and make this pretty resilient.

Yeah, unfortunately, that's the downside of grpc. We need the server impl and currently, it's my toy server 😿  And the reason is simple, unlike file, crds, the server impl can be anything. It could be triggered from DB change, github workflow or anything. But maybe I can expand my server with few more use-cases 🤔 

I addressed other concerns in respective comments :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: I will attempt to introduce the first version of re-try logic :)

Kavindu-Dodan:
@toddbaert I have added connection retry logic with the latest change set :) if you can do another round of review, and make sure this fits flagd standard, thne I will convert this draft to a normal PR 🙌

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+
+	// initial dial and connection. Failure here must result in a startup failure
+	dial, err := grpc.DialContext(ctx, g.Target, options...)
+	if err != nil {
+		g.Logger.Error(fmt.Sprintf("Error establishing grpc connection: %s", err.Error()))
+		return err
+	}
+
+	serviceClient := syncv1grpc.NewFlagSyncServiceClient(dial)
+	syncClient, err := serviceClient.SyncFlags(context.Background(), &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
```
james-milligan, line 54:
```suggestion
	syncClient, err := serviceClient.SyncFlags(ctx, &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
```
Adding suggestions which will allow for graceful shutdown, also requires the below suggestions

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+	if err != nil {
+		g.Logger.Warn(fmt.Sprintf("Error with stream listener: %s", err.Error()))
+	}
+
+	// retry connection establishment
+	for {
+		g.Logger.Warn(fmt.Sprintf("Connection re-establishment attempt in-progress for grpc target: %s", g.Target))
+
+		syncClient = g.connectWithRetry(ctx, options...)
+		err = g.streamListener(ctx, syncClient, dataSync)
```
james-milligan, line 71:
```suggestion
		if syncClient == nil {
			return nil
		}
		err = g.streamListener(ctx, syncClient, dataSync)
```
a `nil` sync client will be returned once the context is closed, so no error is required here

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+	for {
+		var sleep time.Duration
+		if iteration >= backoffLimit {
+			sleep = constantBackoffDelay
+		} else {
+			iteration++
+			sleep = time.Duration(math.Pow(backOffBase, float64(iteration)))
+		}
+
+		time.Sleep(sleep * time.Second)
```
james-milligan, line 95:
```suggestion
		select {
		case <-ctx.Done():
			return nil
		case <-time.After(sleep * time.Second):
```
this allows us to sleep indefinitely, up until we see the context close. At which point we return the nil client, which allows the parent functions for loop to be broken leading to a graceful shutdown

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+
+		serviceClient := syncv1grpc.NewFlagSyncServiceClient(dial)
+		syncClient, err := serviceClient.SyncFlags(context.Background(), &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
+		if err != nil {
+			g.Logger.Debug(fmt.Sprintf("Error openning service client: %s", err.Error()))
+			continue
+		}
+
+		g.Logger.Info(fmt.Sprintf("Connection re-established with grpc target: %s", g.Target))
+		return syncClient
```
james-milligan, line 111:
```suggestion
		return syncClient
	}
```
closing bracket for the select statement

james-milligan (changes requested):

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+		time.Sleep(sleep * time.Second)
+
+		dial, err := grpc.DialContext(ctx, g.Target, options...)
+		if err != nil {
+			g.Logger.Debug(fmt.Sprintf("Error dialing target: %s", err.Error()))
+			continue
+		}
+
+		serviceClient := syncv1grpc.NewFlagSyncServiceClient(dial)
+		syncClient, err := serviceClient.SyncFlags(context.Background(), &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
```
james-milligan, line 104:
```suggestion
		syncClient, err := serviceClient.SyncFlags(ctx, &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
```

james-milligan: missed this in initial review, this prevents the 're-established' client from also blocking the graceful shutdown

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+			continue
+		}
+
+		g.Logger.Info(fmt.Sprintf("Connection re-established with grpc target: %s", g.Target))
+		return syncClient
+	}
+}
+
+// streamListener wraps the grpc listening on provided stream and push updates through dataSync channel
+func (g *Sync) streamListener(
```
james-milligan, line 116: It looks like this function can actually be simplified to:
```
func (g *Sync) streamListener(
	stream syncv1grpc.FlagSyncService_SyncFlagsClient, dataSync chan<- sync.DataSync,
) error {
		err := g.handleFlagSync(stream, dataSync)
	if err == io.EOF {
		g.Logger.Info("Stream closed by the server")
	}
	return err
}
```
The `handleFlagSync` doesn't need to be run in a goroutine, it will block until the parent context closes and it either spits out an EOF, or another error.
Another option is to directly invoke `handleFlagSync` in the `Sync` method, and check the error / log the error there instead, allowing the `handleFlagSync` to block until it is shutdown by the parent context closing.

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,193 @@
...
+
+const (
+	// Prefix for GRPC URL inputs. GRPC does not define a prefix through standard. This prefix helps to differentiate
+	// remote URLs for REST APIs (i.e - HTTP) from GRPC endpoints.
+	Prefix = "grpc://"
+
+	// Connection retry constants
+	// Backoff period is calculated with backOffBase ^ #retry-iteration. However, when backoffLimit is reached, fallback
+	// to constantBackoffDelay
+	backoffLimit         = 3
```
james-milligan, line 30: both `backoff` and `backOff` are used in this PR - my vote is for `backOff`

pkg/sync/grpc/grpc_sync_test.go
```diff
@@ -0,0 +1,330 @@
...
+			go func() {
+				err := grpcSyncImpl.handleFlagSync(test.stream, syncChan)
+				if err != nil {
+					t.Errorf("Error handling flag sync: %s", err.Error())
+				}
+			}()
+			data := <-syncChan
+
+			if data.Type != test.want {
+				t.Errorf("Reuturned data sync state = %v, wanted %v", data.Type, test.want)
```
james-milligan, line 119:
```suggestion
				t.Errorf("Returned data sync state = %v, wanted %v", data.Type, test.want)
```

pkg/sync/grpc/grpc_sync_test.go
```diff
@@ -0,0 +1,330 @@
...
+					t.Errorf("Error from stream listener:  %s", err.Error())
+				}
+			}
+		}()
+
+		for _, expected := range test.output {
+			out := <-syncChan
+
+			if expected.Type != out.Type {
+				t.Errorf("Reuturned sync type = %v, wanted %v", out.Type, expected.Type)
```
james-milligan, line 269:
```suggestion
				t.Errorf("Returned sync type = %v, wanted %v", out.Type, expected.Type)
```

james-milligan (commented):
Few typos

james-milligan:
Ive left a few comments / changes, but overall I really like these changes + am excited to see them merged

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Good one, this slipped through somehow  😞

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Will go ahead with `backOff` :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Yes, we need to handle context here. I will work based on your suggestion :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Yes, correct. We can skip the `streamListener`. 

My initial implression was `stream.Recv()` to be a blocking call, without handling parent context. But I was wrong. Underlying stream is context aware per documentation [1] and it respect the context cancellations. 


[1] - https://pkg.go.dev/google.golang.org/grpc#ClientConn.NewStream

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Updated with the latest commit

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Updated with the latest commit. Rather than nil, I went ahead with an `ok`, which per my view is ideal than a nil :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Updated with the latest commit

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Updated with the latest commit

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Updated with the latest commit

Kavindu-Dodan:
> Ive left a few comments / changes, but overall I really like these changes + am excited to see them merged

Perfect suggestions. I refactored the context handling and things look ok per my view. Please feel free to do another run :)

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+		if !ok {
+			// We shall exit
+			return nil
+		}
+
+		err = g.handleFlagSync(syncClient, dataSync)
+		if err != nil {
+			g.Logger.Warn(fmt.Sprintf("Error with stream listener: %s", err.Error()))
+			continue
+		}
```
toddbaert, line 76: I think one interesting thing to consider here is the behavior of flag evaluations when we are in a disconnected state...

Currently, if the gRPC server is down, the evaluations still work, and I think that's probably optimal. Even dynamic evaluations will still be "correct", they are just possibly out of sync since the connection is down. The obvious question is whether or not we think this constitutes and evaluation error... we could remove all the flags for that sync from the state when this connection is lost. I think that would be the wrong choice though (I just want to point it out). The other thing we could do is consider this a "Reason.STALE" evaluation at that point.

Just food for thought @Kavindu-Dodan @james-milligan @beeme1mr @skyerus

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+)
+
+const (
+	// Prefix for GRPC URL inputs. GRPC does not define a prefix through standard. This prefix helps to differentiate
+	// remote URLs for REST APIs (i.e - HTTP) from GRPC endpoints.
+	Prefix = "grpc://"
+
+	// Connection retry constants
+	// Back off period is calculated with backOffBase ^ #retry-iteration. However, when #retry-iteration count reach
+	// backOffLimit, retry delay fallback to constantBackOffDelay
```
toddbaert, line 27: I think this is the right strategy for a sync... back off, but never stop retrying, and with a max delay. :+1:

toddbaert:
`grpc://` is not an [official, registered URI scheme
](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml). I think the only practical drawback for us here is that it might make it a bit tricky to interpret non IP URLs (such as unix/file sockets, unix://)... Though we could maybe use the "xxx+yyy" convention for this (`grpc+file://`)... anyway, this is just for flagd so it's probably not much of an issue.

Also, though it's not official, ant [least we're not alone in using it](http://nginx.org/en/docs/http/ngx_http_grpc_module.html) (ctrl+f "grpc://")

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Added connection retry handler

pkg/runtime/from_config.go
```diff
@@ -21,12 +21,14 @@ import (
...
 	regCrd  *regexp.Regexp
 	regURL  *regexp.Regexp
+	regGRPC *regexp.Regexp
 	regFile *regexp.Regexp
 )
 
 func init() {
 	regCrd = regexp.MustCompile("^core.openfeature.dev/")
 	regURL = regexp.MustCompile("^https?://")
+	regGRPC = regexp.MustCompile("^" + grpc.Prefix)
```
toddbaert, line 31: Do we need to handle `grpc` as well as `grpcs`? How would we indicate that the remote server expects the connection to be encrypted?

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Thank you :) 

The idea was to exponentially back off till a certain point and then retry with constant delays. 

Current values work out to 1, 4, 16, and then to 60 seconds from there on.

pkg/runtime/from_config.go
toddbaert, line 31: Probably a question for another day / PR... but how do we handle authentication, especially if we have multiple gRPC syncs with different credentials?

I'll also leave this [here](https://grpc.io/docs/guides/auth/) to plug my mutual TLS suggestion :sweat_smile:

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Interesting, I did not think about the flag evaluation perspective.

I checked the existing implementations (ex:- file, k8s and http) and they seem to react differently to errors. For example, k8s CRD removal doesn't do anything other than logging [1] and http sync performs a return [2] (this is a potential bug ? 🤔 ) . 

I think we need to define this behavior and adopt all sync providers. I am leaning toward the `Reason.STALE` suggestion but this also means we need some advanced `flag store` state updates (detecting non-stale -> stale -> non-stale changes per sync source). But the removal of flags when the sync provider has errors in my view is wrong.  

[1] - https://github.com/open-feature/flagd/blob/main/pkg/sync/kubernetes/kubernetes_sync.go#L73
[2] - https://github.com/open-feature/flagd/blob/main/pkg/sync/http/http_sync.go#L53

pkg/runtime/from_config.go
Kavindu-Dodan: I am actively working on a follow-up PR to introduce secure connectivity based on TLS. Haven't considered tokens yet though.

> but how do we handle authentication, especially if we have multiple gRPC syncs with different credentials?

This is the challenge. At least for the initial version, I am considering having an `enable security` configuration along with a `certificate path` to support self-signed certificates. This will be applicable for "all"  grpc connections though. 

May be once that's in place, we can think about `key:value` based solution where key helps to identify specific grpc target and use specific value and so on. IMO a conversation for the next pr 😉

Kavindu-Dodan: Good point. 

As I [highlighted](https://github.com/open-feature/flagd/pull/297#discussion_r1101958011), my current approach considers a configuration option to disable tls connectivity. If this is not set, then default is to dial with system certs or if overriding cert is provided, use that for tls connectivity. 

You can see the draft PR highlights here - https://github.com/open-feature/flagd/pull/398/files#diff-5bd4f361280cd88498f36d821f08f603ba13af72ad0c1ce0313d7b3fa1ed5af5R185-R218

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+		var sleep time.Duration
+		if iteration >= backOffLimit {
+			sleep = constantBackOffDelay
+		} else {
+			iteration++
+			sleep = time.Duration(math.Pow(backOffBase, float64(iteration)))
+		}
+
+		// Block the next connection attempt and check the context
+		select {
```
james-milligan, line 99: nice 👍

james-milligan (approved):

pkg/sync/grpc/grpc_sync.go
toddbaert: I think we can make this a separate concern, and we can talk there about the other sync providers as well.

docs/configuration/flagd_start.md
```diff
@@ -16,12 +16,13 @@ flagd start [flags]
   -z, --log-format string                   Set the logging format, e.g. console or json  (default "console")
   -m, --metrics-port int32                  Port to serve metrics on (default 8014)
   -p, --port int32                          Port to listen on (default 8013)
+  -i, --provider-id string                  Set the identifier of this flagd runtime
```
beeme1mr, line 19: How is the provider id used? Is it required when using grpc?

beeme1mr (approved):

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: I created #400 to follow up on this.

docs/configuration/flagd_start.md
Kavindu-Dodan: It's not mandatory. If not set, the default empty string value gets passed. And yes, this is added to the grpc request [1]

[1] - https://github.com/open-feature/flagd/blob/71f5db9eee014ee12ad255701e64c508ad0e1d34/pkg/sync/grpc/grpc_sync.go#L52

beeme1mr:
Please make sure to include grpc in the docs here.
https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md#uri-patterns

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+	constantBackOffDelay = 60
+)
+
+type Sync struct {
+	Target     string
+	ProviderID string
+	Logger     *logger.Logger
+}
+
+func (g *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
```
ekkinox, line 39: do you think, maybe in another PR, we could allow users to pass their own `DialOption` and fallback on insecure transport if none provided ?

cause since flagd may be used in a distributed environment (k8s pods sidecars), I would for example need to add [OTEL tracing instrumentation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/example/client/main.go#L46-L49) and users could tweak depending on their needs

ekkinox (approved):

docs/configuration/flagd_start.md
beeme1mr: But what's its purpose?

Kavindu-Dodan: Idea is to identify the specific flagd deployment at the grpc server side, thus allowing the server to push flag configurations unique to a flagd deployment. 

For example, in a K8s cluster,  both apps A & B could have their own flagd side-cars. And this cluster can have a single grpc server implementation. When flagd connects to the server, they can send their identifier (ex:- appA, appB), which the server can use to push unique flag configurations. 

Note that this is not an authentication mechanism but rather an identifier for filtering.

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: I am actively working on extending grpc connectivity to support TLS connections [1]. This should allow altering connectivity between TLS and Insecure based on schema (`grpc://` vs `grpcs://`). Further, we are having high-level discussions/investigations on improving configuration options of flagd [2].

However, at least for now, there is no plan to support OTEL tracing and related configurations. If you see this as a valid use case, if you have some spare time, you can always contribute with a PR. Our community is happy to help you here. 

[1] - https://github.com/open-feature/flagd/pull/398
[2] - https://github.com/open-feature/flagd/issues/405

pkg/runtime/from_config.go
Kavindu-Dodan: I will resolve this conversation in this context. We can handle this with TLS related PR coming up soon - https://github.com/open-feature/flagd/pull/398

docs/configuration/flagd_start.md
james-milligan: It may be worthwhile waiting for[ this discussion](https://github.com/open-feature/flagd/issues/405) to conclude before we merge this, it looks like the ProviderID is only currently being used in the grpc sync provider, may only be relevant to specific implementations of individual grpc providers, and there is currently no way to configure this ID on a per sync basis - i.e. if we have 2 grpc syncs running simultaneously -  would like to avoid needing to deprecate this flag if possible

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+	options := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+
+	// initial dial and connection. Failure here must result in a startup failure
+	dial, err := grpc.DialContext(ctx, g.Target, options...)
+	if err != nil {
+		g.Logger.Error(fmt.Sprintf("Error establishing grpc connection: %s", err.Error()))
+		return err
```
skyerus, line 48: If the errors are being handled (logged) then is it necessary to return them too? Could this lead the errors to be logged multiple times?

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+
+func (g *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
+	options := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
+
+	// initial dial and connection. Failure here must result in a startup failure
+	dial, err := grpc.DialContext(ctx, g.Target, options...)
+	if err != nil {
+		g.Logger.Error(fmt.Sprintf("Error establishing grpc connection: %s", err.Error()))
```
skyerus, line 47: Did we decide on lower case logging?

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+	serviceClient := syncv1grpc.NewFlagSyncServiceClient(dial)
+	syncClient, err := serviceClient.SyncFlags(ctx, &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
+	if err != nil {
+		g.Logger.Error(fmt.Sprintf("Error calling streaming operation: %s", err.Error()))
+		return err
+	}
+
+	// initial stream listening
+	err = g.handleFlagSync(syncClient, dataSync)
+	if err != nil {
```
skyerus, line 60: err can never be nil, could either leave a comment indicating this or remove the nil check

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,186 @@
...
+
+		err = g.handleFlagSync(syncClient, dataSync)
+		if err != nil {
+			g.Logger.Warn(fmt.Sprintf("Error with stream listener: %s", err.Error()))
+			continue
+		}
+	}
+}
+
+// connectWithRetry is a helper to perform exponential back off till provided configurations and then retry connection
```
skyerus, line 80: Could you check this first sentence

skyerus (commented):
Minor concerns, nice work :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: yes, will correct :)

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: Sync providers are backed by an error group. And the caller (the `Runtime`) has the option to log or ignore logging if there's an error. In my view, logging and returning error gives better details when something goes wrong. Consider the following log for a startup failure, 

```
2023-02-14T08:48:18.905-0800    info    service/connect_service.go:110  metrics listening at 8014       {"component": "service"}
2023-02-14T08:48:18.908-0800    error   grpc/grpc_sync.go:54    error calling streaming operation: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:9090: connect: connection refused"     {"component": "sync", "sync": "grpc"}
github.com/open-feature/flagd/pkg/sync/grpc.(*Sync).Sync
        /Users/kavindu.dodanduwa/Development/Git/flagd/pkg/sync/grpc/grpc_sync.go:54
github.com/open-feature/flagd/pkg/runtime.(*Runtime).Start.func2
        /Users/kavindu.dodanduwa/Development/Git/flagd/pkg/runtime/runtime.go:77
golang.org/x/sync/errgroup.(*Group).Go.func1
        /Users/kavindu.dodanduwa/go/pkg/mod/golang.org/x/sync@v0.1.0/errgroup/errgroup.go:75
2023-02-14T08:48:18.909-0800    fatal   cmd/start.go:122        rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:9090: connect: connection refused"        {"component": "start"}
github.com/open-feature/flagd/cmd.glob..func1
        /Users/kavindu.dodanduwa/Development/Git/flagd/cmd/start.go:122
github.com/spf13/cobra.(*Command).execute
        /Users/kavindu.dodanduwa/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920
github.com/spf13/cobra.(*Command).ExecuteC
        /Users/kavindu.dodanduwa/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044
github.com/spf13/cobra.(*Command).Execute
        /Users/kavindu.dodanduwa/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/open-feature/flagd/cmd.Execute
        /Users/kavindu.dodanduwa/Development/Git/flagd/cmd/root.go:37
main.main
        /Users/kavindu.dodanduwa/Development/Git/flagd/main.go:30
runtime.main
        /usr/local/go/src/runtime/proc.go:250
```

docs/configuration/flagd_start.md
james-milligan: My vote is to remove this option for now, and reintroduce it as a feature once we have a better foundation for cleanly introducing it

Kavindu-Dodan: I removed this option from configurations. Let's add this later with proper focus when #405 is done.

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: I improved the code documentation

pkg/sync/grpc/grpc_sync.go
toddbaert: I think in _legitimate_ error scenarios (ones that are unexpected or occur when you're first fitting things together) some duplicated errors is probably better than missing context.

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,183 @@
...
+
+		err = g.handleFlagSync(syncClient, dataSync)
+		if err != nil {
+			g.Logger.Warn(fmt.Sprintf("error with stream listener: %s", err.Error()))
+			continue
+		}
+	}
+}
+
+// connectWithRetry is a helper to perform exponential back off and then retry connection periodically till a successful
```
skyerus, line 78:
```suggestion
// connectWithRetry is a helper that performs exponential back off after retrying connection attempts periodically until a successful
```

pkg/sync/grpc/grpc_sync.go
```diff
@@ -0,0 +1,183 @@
...
+		dial, err := grpc.DialContext(ctx, g.Target, options...)
+		if err != nil {
+			g.Logger.Debug(fmt.Sprintf("error dialing target: %s", err.Error()))
+			continue
+		}
+
+		serviceClient := syncv1grpc.NewFlagSyncServiceClient(dial)
+		syncClient, err := serviceClient.SyncFlags(ctx, &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
+		if err != nil {
+			g.Logger.Debug(fmt.Sprintf("error openning service client: %s", err.Error()))
```
skyerus, line 115:
```suggestion
			g.Logger.Debug(fmt.Sprintf("error opening service client: %s", err.Error()))
```

skyerus (approved):

Issue #237, opened 2022-12-20
[FEATURE] Pretty-print logging


james-milligan:
### Requirements

Flagd currently prints structured logs in json format, when running flagd in the console this can cause the logs to be difficult to read.

the zap logger `zap.Config` struct contains the field `Encoding` which when set to `console` instead of `json` updates the logging structure as follows:

`json`
```
{"level":"info","ts":1671548378.306176,"caller":"runtime/runtime.go:84","msg":"configuration change (write) for flagKey headerColor (config/samples/example_flags.json)","component":"runtime"}
{"level":"info","ts":1671548378.3061912,"caller":"runtime/runtime.go:84","msg":"configuration change (write) for flagKey myBoolFlag (config/samples/example_flags.json)","component":"runtime"}
{"level":"info","ts":1671548378.306196,"caller":"runtime/runtime.go:84","msg":"configuration change (write) for flagKey myStringFlag (config/samples/example_flags.json)","component":"runtime"}
{"level":"info","ts":1671548378.306199,"caller":"runtime/runtime.go:84","msg":"configuration change (write) for flagKey myFloatFlag (config/samples/example_flags.json)","component":"runtime"}
{"level":"info","ts":1671548378.3062022,"caller":"runtime/runtime.go:84","msg":"configuration change (write) for flagKey myIntFlag (config/samples/example_flags.json)","component":"runtime"}
```
`console`
```
1.671549411055416e+09   info    runtime/runtime.go:84   configuration change (write) for flagKey myFloatFlag (config/samples/example_flags.json)        {"component": "runtime"}
1.6715494110554419e+09  info    runtime/runtime.go:84   configuration change (write) for flagKey myIntFlag (config/samples/example_flags.json)  {"component": "runtime"}
1.671549411055447e+09   info    runtime/runtime.go:84   configuration change (write) for flagKey myObjectFlag (config/samples/example_flags.json)       {"component": "runtime"}
1.671549411055451e+09   info    runtime/runtime.go:84   configuration change (write) for flagKey isColorYellow (config/samples/example_flags.json)      {"component": "runtime"}
```

This behaviour should be controlled via a viper flag, e.g. `--log-encoding=console` with a `json` default

mihirm21:
Hi, can I be assigned this issue in case no one else is working on it, or should I directly submit my PR?

mihirm21:
plz check whether the required work is done

mihirm21:
Is the issue resolved now ?

[...]

PR #302, opened 2023-01-22
feat: updated logging to console

mihirm21:
Signed-off-by: Mihir Mittal <105881639+mihirm21@users.noreply.github.com>

## This PR

-updated logging to console from JSON format as assigned

### Related Issues

[FEATURE] Pretty-print logging #237


james-milligan:
Ideally this value would be configurable via a startup flag, with `json` as the default, e.g. passing `--log-encoding=console` would set the encoding to `console`

james-milligan (changes requested):
This value needs to be configurable via a startup flag, with json as the default, e.g. passing --`log-encoding=console` would set the encoding to console

PR #307, opened 2023-01-23
refactor: introduce add, update and delete operations to sync providers

Kavindu-Dodan:
## This PR
Introduce and wire new strategies to sync provider state updates.

Following new mechanisms were added to be used by sync providers (ex:- file based flag configuration provider),

- ADD: Add new flag(s) to internal state
- UPDATE: Update existing flag(s) to internal state
- DELETE: Remove flag(s) from internal state

These operations provide extra flexibility compared to the existing "merge" operation, which is now mapped to "ALL" (i.e - Add and replace all). 

 Test coverage of 100% for newly added methods.

### Related Issues

Related to #249


pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
 	return je.state.Flags[flagKey].DefaultVariant, reason, nil
 }
 
+// configToFlags convert string configurations to flags and store them to pointer newFLags
```
beeme1mr, line 271:
```suggestion
// configToFlags convert string configurations to flags and store them to pointer newFlags
```

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
...
+	}
+
+	transposedConfig, err := je.transposeEvaluators(config)
+	if err != nil {
+		return fmt.Errorf("transpose evaluators: %w", err)
+	}
+
+	err = json.Unmarshal([]byte(transposedConfig), &newFlags)
+	if err != nil {
+		return fmt.Errorf("unmarshal new state: %w", err)
```
toddbaert, line 291: I think these messages could be improved. Besides the fact that it probably outputs to stderr, it's not clear that it's an error. I think I'd prefer something like "unable to transpose evaluators..."

Kavindu-Dodan: No, this is not writing to stderr. `fmt.Errorf` generates and error, with the formatted error message [1] :) .

Sure, I will improve the error message here. I moved this code and did not change the logic, but given I am touching this, it makes sense to improve it 

[1] - https://pkg.go.dev/fmt#Errorf

toddbaert (approved):
Though I'm not sure we will anytime soon implement a sync that would take advantage of this... we might need one at some point to avoid large payloads for only small flag changes.

I think it's worth merging this to be able to easily add that support, but I would like @skyerus, @james-milligan or @AlexsJones 's opinion.

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
 	return je.state.Flags[flagKey].DefaultVariant, reason, nil
 }
 
+// configToFlags convert string configurations to flags and store them to pointer newFlags
+func (je *JSONEvaluator) configToFlags(config string, newFlags *Flags) error {
+	schemaLoader := gojsonschema.NewStringLoader(schema.FlagdDefinitions)
+	flagStringLoader := gojsonschema.NewStringLoader(config)
+	result, err := gojsonschema.Validate(schemaLoader, flagStringLoader)
+
```
skyerus, line 276: err check ought to come directly after the assignment (no return)

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
...
+	if err != nil {
+		return err
+	} else if !result.Valid() {
+		err := errors.New("invalid JSON file")
+		return err
```
skyerus, line 281:
```suggestion
		return errors.New("invalid JSON file")
```

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
...
+	}
+
+	transposedConfig, err := je.transposeEvaluators(config)
+	if err != nil {
+		return fmt.Errorf("error transposing evaluators: %w", err)
```
skyerus, line 286:
```suggestion
		return fmt.Errorf("transposing evaluators: %w", err)
```
error is inferred

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
...
+	}
+
+	err = json.Unmarshal([]byte(transposedConfig), &newFlags)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling provided configurations: %w", err)
```
skyerus, line 291:
```suggestion
		return fmt.Errorf("unmarshalling provided configurations: %w", err)
```

pkg/eval/json_evaluator_model.go
```diff
@@ -8,45 +8,139 @@ import (
...
+		}
+
+		notifications[k] = map[string]interface{}{
+			"type":   string(NotificationCreate),
+			"source": source,
+		}
+
+		// Store the new version of the flag
+		newFlag.Source = source
+		f.Flags[k] = newFlag
```
skyerus, line 48: Is there a risk of race conditions if these methods are called by multiple goroutines? Is that a possibility?

pkg/eval/json_evaluator.go
```diff
@@ -274,8 +268,37 @@ func (je *JSONEvaluator) evaluateVariant(
...
+// configToFlags convert string configurations to flags and store them to pointer newFlags
+func (je *JSONEvaluator) configToFlags(config string, newFlags *Flags) error {
+	schemaLoader := gojsonschema.NewStringLoader(schema.FlagdDefinitions)
+	flagStringLoader := gojsonschema.NewStringLoader(config)
+	result, err := gojsonschema.Validate(schemaLoader, flagStringLoader)
+
+	if err != nil {
+		return err
+	} else if !result.Valid() {
+		err := errors.New("invalid JSON file")
```
james-milligan, line 280: `result.Errors()` provides access to an array of validation errors, this information may be useful for debugging configurations so it may be worth wrapping returned errors, as an example these errors are from the negative tests in the schemas repo:
```
flags.not-a-flag: Must validate one and only one schema (oneOf)
flags.not-a-flag: state is required
```

pkg/eval/json_evaluator_model.go
```diff
@@ -8,45 +8,139 @@ import (
...
+}
+
+// Add new flags from source
+func (f Flags) Add(logger *logger.Logger, source string, ff Flags) map[string]interface{} {
+	notifications := map[string]interface{}{}
+
+	for k, newFlag := range ff.Flags {
+		if storedFlag, ok := f.Flags[k]; ok && storedFlag.Source != source {
+			logger.Warn(fmt.Sprintf(
+				"flag with key %s from source %s already exist, overriding this with flag from source %s",
```
james-milligan, line 34: do we want to document this behaviour? Or, expand the new signatures to add a priority, e.g. pass the uri argument index to maintain a priority order, preventing lower priority indexes from overwriting existing values? (this may sit outside of the scope of this PR)

```
flagd start --uri file:example_flags.json --uri file:example_flags_lowest_priority.json
```
in this example updates to the `example_flags_lowest_priority.json` source would not be able to overwrite values from `example_flags.json`

pkg/eval/json_evaluator_model.go
Kavindu-Dodan: We avoid race conditions with a mutex from Runtime [1]. So this operation is safe. 

[1] - https://github.com/open-feature/flagd/blob/main/pkg/runtime/runtime.go#L93

pkg/eval/json_evaluator.go
Kavindu-Dodan: Good suggestion. I did not check the internals as this code came from a refactoring.

I will add detailed logs when this occur 

ex:

`"msg":"JSON schema validation failed:  1:flags.myBoolFlag: Must validate one and only one schema (oneOf)  2:flags.myBoolFlag: state is required  3:flags.myBoolFlag: Must validate all the schemas (allOf) `

pkg/eval/json_evaluator_model.go
Kavindu-Dodan: IMO setting a priority makes the implementation complex (i.e - logic to validate priority and store the priority alongside the configuration source). 

We already highlight something similar for startup configurations [1]

> Using multiple providers is supported however if flag keys are duplicated across multiple sources it may lead to unexpected behavior

Do you think we should improve this description ?

[1] - https://github.com/open-feature/flagd/blob/main/cmd/start.go#L46

pkg/eval/json_evaluator.go
Kavindu-Dodan: please check the latest changes :)

pkg/eval/json_evaluator_model.go
skyerus: It may be worth adding a comment to the `SetState` function that it's not safe to call it concurrently

pkg/eval/json_evaluator_model.go
james-milligan: You're right, no I think what we have here is OK

pkg/eval/json_evaluator_model.go
Kavindu-Dodan: Improved comments to highlight non-thread safe fact

james-milligan (approved):

pkg/eval/json_evaluator_model.go
toddbaert: Yep I agree. I think duped flags are enough of a weird edge case that we don't need to define it.

toddbaert (approved):

beeme1mr (approved):

PR #310, opened 2023-01-23
chore: add architecture overview for recent refactoring

Kavindu-Dodan:
## This PR

Adds a component diagram for flag configuration providers and their interactions with the runtime and other systems.

Related PR #291


docs/other_resources/high_level_architecture.md
```diff
@@ -1,3 +1,8 @@
 # High level architecture
 
+### Component overview
 <img src="../images/of-flagd-0.png" width="560">
+
+### Flag configuration provider
```
beeme1mr, line 6: Should these be H2s? Also, could you add a brief description under each section?

Kavindu-Dodan: Yeah they can be H2. I will update the sections with next commit :)

beeme1mr (approved):

toddbaert:
I find the label "ISync Impl Listner" (sp) a bit misleading, probable because in code/comments, we refer to it something like "DataSync channel watcher"... That is what this represents, right? I think that might be a better name. Or just "DataSync Channel"? I think we should choose a consistent name for this abstraction.

Kavindu-Dodan:
> I find the label "ISync Impl Listner" (sp) a bit misleading, probable because in code/comments, we refer to it something like "DataSync channel watcher"... That is what this represents, right? I think that might be a better name. Or just "DataSync Channel"? I think we should choose a consistent name for this abstraction.

I agree with this. I will rephrase this to "DataSync channel listener"

james-milligan (approved):

beeme1mr (approved):

toddbaert (approved):

toddbaert:
@Kavindu-Dodan I meant to ask this before merge, but forgot... can you also commit the excalidraw source?

PR #339, opened 2023-01-28
fix: send datasync on remove fs events

toddbaert:
Looks like during a refactor, we accidentally removed the "meat" of the remove handler.

In K8S, we never get write events, just remove. It's necessary to send the data for these as well.

Also extracted some duplicated code into a func.


beeme1mr (approved):

beeme1mr:
Bypassing branch protection in order to release this bug fix because updates don't work in Kubernetes environments that are in file watching mode.

PR #344, opened 2023-01-30
docs: updated filepaths for schema store regex

james-milligan:

## This PR

- updates file paths to conform to schema store regex


skyerus (approved):

Kavindu-Dodan (approved):

beeme1mr:
Homebrew must be updated prior to making this change. Please also review if other repos are directly references those flagd configs.

https://github.com/Homebrew/homebrew-core/blob/master/Formula/flagd.rb#L41

beeme1mr:
It may be safer to copy the existing examples instead of moving them. We'll just want to watch out for references to the old style so they can be updated.

beeme1mr (approved):
Please keep the original samples and add the new style next to them. We can then update all the references to the old style before completely removing them.

beeme1mr (changes requested):
Please retain the original examples as well.

beeme1mr (approved):

PR #361, opened 2023-02-01
fix: use event.Has func for file change notification handling (increased stability across OS)

### LINKED ISSUE #313: [BUG] File change notifications behaviour is inconsistent (OS specific)
### Observed behavior

While implementing integration tests of flagd, I initially ran the tests against the flagd binary which seemingly arbitrarily passed/failed. Conversely, the tests always pass while using the built flagd image directly.

The "flakey" tests are the ones that make a change to the flag config file, sleeps 50 milliseconds & then asserts that the flag has been cleared from cache.

Increasing the sleep (all the way up to 2 seconds) did nothing to fix the flakiness. This leads me to believe the behaviour of my mac's file change notifications system is inconsistent with the behaviour of flagd's distroless image's system.

### Expected Behavior

The tests to pass as consistently when using the flagd binary as when using flagd's image.

### Steps to reproduce

Not easily reproducible right now (the flagd integration tests and its dependencies are yet to be merged). Will update when they have been

skyerus:
## This PR

- Uses event.Has func for file change notification handling (instead of direct comparison). This is the correct approach as [expressed directly in fsnotify's code](https://github.com/fsnotify/fsnotify/blob/main/fsnotify.go#L39).

### Related Issues

Fixes #313

### DIFF
--- a/pkg/sync/file/filepath_sync.go
+++ b/pkg/sync/file/filepath_sync.go
@@ -57,10 +57,9 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 
 			fs.Logger.Info(fmt.Sprintf("filepath event: %s %s", event.Name, event.Op.String()))
 
-			switch event.Op {
-			case fsnotify.Create, fsnotify.Write:
+			if event.Has(fsnotify.Create) || event.Has(fsnotify.Write) {
 				fs.sendDataSync(ctx, event, dataSync)
-			case fsnotify.Remove:
+			} else if event.Has(fsnotify.Remove) {
 				// Counterintuively, remove events are the only meanful ones seen in K8s.
 				// At the point the remove event is fired, we have our new data, so we can send it down the channel.
 				fs.sendDataSync(ctx, event, dataSync)
--- a/pkg/sync/file/filepath_sync.go
+++ b/pkg/sync/file/filepath_sync.go
@@ -57,6 +57,8 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 
 			fs.Logger.Info(fmt.Sprintf("filepath event: %s %s", event.Name, event.Op.String()))
 
+                        // event.Op is bitmask and some systems may send multiple operations at once
+                        // event.Has(...) checks that the bitmask contains the particular event (among others)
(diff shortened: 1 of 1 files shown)


james-milligan (approved):

beeme1mr:
I still don't understand why this change increases stability across OSes.

pkg/sync/file/filepath_sync.go
```diff
@@ -57,10 +57,9 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 
 			fs.Logger.Info(fmt.Sprintf("filepath event: %s %s", event.Name, event.Op.String()))
 
-			switch event.Op {
-			case fsnotify.Create, fsnotify.Write:
+			if event.Has(fsnotify.Create) || event.Has(fsnotify.Write) {
```
toddbaert, line 60:
```suggestion
                        // event.Op is bitmask and some systems may send multiple operations at once
                        // event.Has(...) checks that the bitmask contains the particular event (among others)
			if event.Has(fsnotify.Create) || event.Has(fsnotify.Write) {
```

pkg/sync/file/filepath_sync.go
```diff
@@ -57,10 +57,9 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
...
 				fs.sendDataSync(ctx, event, dataSync)
-			case fsnotify.Remove:
+			} else if event.Has(fsnotify.Remove) {
 				// Counterintuively, remove events are the only meanful ones seen in K8s.
 				// At the point the remove event is fired, we have our new data, so we can send it down the channel.
```
toddbaert, line 64:
```suggestion
				// K8s handles mounted ConfigMap updates by modifying symbolic links, which is an atomic operation.
				// At the point the remove event is fired, we have our new data, so we can send it down the channel.
```

toddbaert: This one isn't exactly related to your change, but I was going to open a separate PR for this, so I'm piggy-backing here :sweat_smile:

toddbaert:
> I still don't understand why this change increases stability across OSes.

@beeme1mr I added a comment to @skyerus 's change that should explain it.

pkg/sync/file/filepath_sync.go
skyerus: Thank you!

PR #401, opened 2023-02-10
fix: fix race in http sync test

toddbaert:
In the `TestSimpleSync` test we are asserting that `hs.Cron.AddFunc` and `hs.Cron.Start` are called by the time the `dataSync` is sent. This is a race, because these methods are actually called after the first send. I get failures here ~25% of the time (seems like @skyerus does too).

This changes things to add the handler and start the cron _before_ the first send, which makes things deterministic.

### DIFF
--- a/pkg/sync/http/http_sync.go
+++ b/pkg/sync/http/http_sync.go
@@ -44,8 +44,6 @@ func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 		return err
 	}
 
-	dataSync <- sync.DataSync{FlagData: fetch, Source: hs.URI, Type: sync.ALL}
-
 	_ = hs.Cron.AddFunc("*/5 * * * *", func() {
 		body, err := hs.fetchBodyFromURL(ctx, hs.URI)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/sync/http/http_sync_test.go
```diff
@@ -55,6 +56,37 @@ func TestSimpleSync(t *testing.T) {
...
+		Logger:      logger.NewLogger(nil, false),
+	}
+
+	ctx := context.Background()
+	dataSyncChan := make(chan sync.DataSync)
+
+	// this should return immediately.
+	err := httpSync.Sync(ctx, dataSyncChan)
+	if err == nil {
+		t.Errorf("expected error")
```
Kavindu-Dodan, line 86: I assume the error is returned due to invalid cron expression ? If so, can we improve the test log 🤔

toddbaert: Resolved by completely removing this check...we know the expression is valid since we have it hard coded.

pkg/sync/http/http_sync.go
```diff
@@ -82,6 +80,9 @@ func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 	})
 
 	hs.Cron.Start()
+
+	dataSync <- sync.DataSync{FlagData: fetch, Source: hs.URI, Type: sync.ALL}
+
```
toddbaert, line 85: By moving this here, we know the add/start ran, which is what we assert in our tests, resolving the race.

Kavindu-Dodan (approved):

james-milligan (approved):

skyerus (approved):

PR #414, opened 2023-02-15
feat: introduce flag merge behaviour

james-milligan:
## This PR

- introduces a priority merge order for sync providers
- triggers a resync of the state store on delete events, ensuring the store doesn't fall out of sync with flag sources

### DIFF
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -62,7 +62,16 @@ func (r *Runtime) Start() error {
 		for {
 			select {
 			case data := <-dataSync:
-				r.updateWithNotify(data)
+				if resyncRequired := r.updateWithNotify(data); resyncRequired {
+					for _, s := range r.SyncImpl {
+						p := s
+						go func() {
+							g.Go(func() error {
+								return p.ReSync(gCtx, dataSync)
+							})
+						}()
+					}
+				}
 			case <-gCtx.Done():
 				return nil
 			}
@@ -89,14 +98,14 @@ func (r *Runtime) Start() error {
 }
 
 // updateWithNotify helps to update state and notify listeners
-func (r *Runtime) updateWithNotify(payload sync.DataSync) {
+func (r *Runtime) updateWithNotify(payload sync.DataSync) bool {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
-	notifications, err := r.Evaluator.SetState(payload)
+	notifications, resyncRequired, err := r.Evaluator.SetState(payload)
 	if err != nil {
 		r.Logger.Error(err.Error())
-		return
+		return false
 	}
 
 	r.Service.Notify(service.Notification{
@@ -105,4 +114,6 @@ func (r *Runtime) updateWithNotify(payload sync.DataSync) {
 			"flags": notifications,
 		},
 	})
+
+	return resyncRequired
 }
--- a/pkg/store/flags_test.go
+++ b/pkg/store/flags_test.go
@@ -361,6 +361,10 @@ func TestFlags_Delete(t *testing.T) {
 					"B": {Source: mockSource},
 					"C": {Source: mockSource2},
 				},
+				FlagSources: []string{
+					mockSource,
+					mockSource2,
+				},
 			},
 			deleteRequest: map[string]model.Flag{
 				"A": {Source: mockSource},
@@ -369,6 +373,10 @@ func TestFlags_Delete(t *testing.T) {
 				Flags: map[string]model.Flag{
 					"B": {Source: mockSource},
 					"C": {Source: mockSource2},
+				},
+				FlagSources: []string{
+					mockSource,
+					mockSource2,
 				},
 			},
 			expectedNotificationKeys: []string{"A"},
@@ -381,6 +389,10 @@ func TestFlags_Delete(t *testing.T) {
 					"B": {Source: mockSource},
 					"C": {Source: mockSource2},
 				},
+				FlagSources: []string{
+					mockSource,
+					mockSource2,
+				},
 			},
 			deleteRequest: map[string]model.Flag{
 				"C": {Source: mockSource},
@@ -390,6 +402,10 @@ func TestFlags_Delete(t *testing.T) {
 					"A": {Source: mockSource},
 					"B": {Source: mockSource},
 					"C": {Source: mockSource2},
+				},
+				FlagSources: []string{
+					mockSource,
+					mockSource2,
 				},
 			},
 			expectedNotificationKeys: []string{},
--- a/pkg/store/flags_test.go
+++ b/pkg/store/flags_test.go
@@ -122,7 +122,7 @@ func TestMergeFlags(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			gotNotifs := tt.current.Merge(logger.NewLogger(nil, false), tt.newSource, tt.new)
+			gotNotifs, _ := tt.current.Merge(logger.NewLogger(nil, false), tt.newSource, tt.new)
 			require.Equal(t, tt.want, tt.want)
 			require.Equal(t, tt.wantNotifs, gotNotifs)
 		})
--- a/pkg/sync/isync.go
+++ b/pkg/sync/isync.go
@@ -16,6 +16,7 @@ const (
 	UPDATE
 	// DELETE - Delete for flag(s) previously provided
 	DELETE
+	RESYNC
 )
 
 func (t Type) String() string {
@@ -28,6 +29,8 @@ func (t Type) String() string {
 		return "UPDATE"
 	case DELETE:
 		return "DELETE"
+	case RESYNC:
+		return "RESYNC"
 	default:
 		return "UNKNOWN"
 	}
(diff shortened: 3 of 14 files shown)


toddbaert:
I'm looking at this now. Sorry about the wait @james-milligan

pkg/store/flags_test.go
```diff
@@ -9,6 +9,68 @@ import (
...
+			currentState: &Flags{
+				FlagSources: []string{
+					"A",
+					"B",
+				},
+			},
+			storedSource: "B",
+			newSource:    "A",
+			hasPriority:  false,
+		},
```
toddbaert, line 50:
```suggestion
		{
			name: "priority",
			currentState: &Flags{
				FlagSources: []string{
					"A",
					"B",
				},
			},
			storedSource: "A",
			newSource:    "B",
			hasPriority:  true,
		},
```

I think this test is a lie. I believe it should be the above, which then covers an additional and important (I think?) code path.

The coverage seems to back this up, it goes from `87.9% of statements` to `89.0% of statements` for this package with this change.

pkg/sync/isync.go
```diff
@@ -16,6 +16,7 @@ const (
 	UPDATE
 	// DELETE - Delete for flag(s) previously provided
 	DELETE
+	// DELETE - All flags of sync provider. Prevents the triggering of a second resync event
```
toddbaert, line 19: This comment seems out of place.

pkg/sync/isync.go
```diff
@@ -41,6 +42,10 @@ type ISync interface {
 	// Sync is the contract between Runtime and sync implementation.
 	// Note that, it is expected to return the first data sync as soon as possible to fill the store.
 	Sync(ctx context.Context, dataSync chan<- DataSync) error
+
+	// ReSync is used to validate the flag store following a delete event
```
toddbaert, line 46:
```suggestion
	// ReSync is used to validate the flag store following a delete event
```

I'm not sure how I might change it, but I think this function is more "generally useful" than this comment implies. It might also be useful if we moved to using deltas in gRPC and we needed to get the whole state for a brand new subscriber, for instance.

toddbaert, line 46: I'm not sure how I might change it, but I think this function is more "generally useful" than this comment implies. It might also be useful if we moved to using deltas in gRPC and we needed to get the whole state for a brand new subscriber, for instance.

toddbaert (approved):

pkg/sync/isync.go
james-milligan: Updated this comment to remove the specific example

Kavindu-Dodan:
I had a look into the PR and I think we need some agreements. 

Current implementation focus on resyncs after a deletion. This means the changes we introduce only concern the store delete scenarios. In my view, we need to be careful about core logic (ex:- store - runtime - isync ) changes and breaking changes must be justifiable.

With that, can't we simply rely on the following and ignore resyncs altogether, 

- Sync provider priority is defined by the order of sync args 
- If low-priority providers override an existing flag simply ignore 
- If high priority provider overrides an existing flag, accept that 
- Ignore any other assumptions - we rely on ISync impls to refresh any changes 

wouldn't such an approach work or am I not seeing the big picture? 🤔

toddbaert:
> * Sync provider priority is defined by the order of sync args
> * If low-priority providers override an existing flag simply ignore
> * If high priority provider overrides an existing flag, accept that
> * Ignore any other assumptions - we rely on ISync impls to refresh any changes
> 
> wouldn't such an approach work or am I not seeing the big picture? thinking

That seems to be (to me) more or less what this PR does. See the tests here: https://github.com/open-feature/flagd/pull/414/files/5415b75a4c48bf3ff8014bd92428666711a0a1f8#diff-575f3e52e026f848e85c7a6cd9561cda5def99f4a58a9786801d4d4097f0ae03R12-R73

Kavindu-Dodan:
> > * Sync provider priority is defined by the order of sync args
> > * If low-priority providers override an existing flag simply ignore
> > * If high priority provider overrides an existing flag, accept that
> > * Ignore any other assumptions - we rely on ISync impls to refresh any changes
> > 
> > wouldn't such an approach work or am I not seeing the big picture? thinking
> 
> That seems to be (to me) more or less what this PR does. See the tests here: https://github.com/open-feature/flagd/pull/414/files/5415b75a4c48bf3ff8014bd92428666711a0a1f8#diff-575f3e52e026f848e85c7a6cd9561cda5def99f4a58a9786801d4d4097f0ae03R12-R73

Yes, I also think this is what it does. But I am yet to see the need for a resync 🤔   If a low-priority sync provider can't override an existing flag & if sync providers publish updates for changes, then is there any reason for a force refill?

james-milligan:
> > > * Sync provider priority is defined by the order of sync args
> > > * If low-priority providers override an existing flag simply ignore
> > > * If high priority provider overrides an existing flag, accept that
> > > * Ignore any other assumptions - we rely on ISync impls to refresh any changes
> > > 
> > > wouldn't such an approach work or am I not seeing the big picture? thinking
> > 
> > 
> > That seems to be (to me) more or less what this PR does. See the tests here: https://github.com/open-feature/flagd/pull/414/files/5415b75a4c48bf3ff8014bd92428666711a0a1f8#diff-575f3e52e026f848e85c7a6cd9561cda5def99f4a58a9786801d4d4097f0ae03R12-R73
> 
> Yes, I also think this is what it does. But I am yet to see the need for a resync 🤔 If a low-priority sync provider can't override an existing flag & if sync providers publish updates for changes, then is there any reason for a force refill?

If flag-a exists in 2 sync providers, (provider-A and provider-B) with provider-A taking priority, then the starup and initial datasync will result in the value of flag-a from provider-B being discarded. In the event that the value of flag-a is then deleted in provider-A then we will not recieve the previously discarded flag-a value from provider-B until there is another update event from it, and the sdk will always return the default value.
We could habdle this with with the existing Sync method, however is many cases we will be starting a new watcher just to close it after the first event is written to the channel IMO this is innapropriate.
Another option is to store all 'discarded' states incase they are needed (or become the highest priority via a deletion) but again this is overly complex for an event that will not happen too often

Kavindu-Dodan:
> > > > * Sync provider priority is defined by the order of sync args
> > > > * If low-priority providers override an existing flag simply ignore
> > > > * If high priority provider overrides an existing flag, accept that
> > > > * Ignore any other assumptions - we rely on ISync impls to refresh any changes
> > > > 
> > > > wouldn't such an approach work or am I not seeing the big picture? thinking
> > > 
> > > 
> > > That seems to be (to me) more or less what this PR does. See the tests here: https://github.com/open-feature/flagd/pull/414/files/5415b75a4c48bf3ff8014bd92428666711a0a1f8#diff-575f3e52e026f848e85c7a6cd9561cda5def99f4a58a9786801d4d4097f0ae03R12-R73
> > 
> > 
> > Yes, I also think this is what it does. But I am yet to see the need for a resync 🤔 If a low-priority sync provider can't override an existing flag & if sync providers publish updates for changes, then is there any reason for a force refill?
> 
> If flag-a exists in 2 sync providers, (provider-A and provider-B) with provider-A taking priority, then the starup and initial datasync will result in the value of flag-a from provider-B being discarded. In the event that the value of flag-a is then deleted in provider-A then we will not recieve the previously discarded flag-a value from provider-B until there is another update event from it, and the sdk will always return the default value. We could habdle this with with the existing Sync method, however is many cases we will be starting a new watcher just to close it after the first event is written to the channel IMO this is innapropriate. Another option is to store all 'discarded' states incase they are needed (or become the highest priority via a deletion) but again this is overly complex for an event that will not happen too often. The infrequency of these update events also suggests that in the example described above flagd would not have a flag value for flag-a for a long time

Thank you. I now understand the scenario you want to handle. This shows the impact of caching and the difficulty to maintain them.

I am still in favor of not having a refill strategy for removals to minimize the complexity of store and runtime components. But may be others have different opinions and this could be a desirable feature.

pkg/runtime/runtime.go
```diff
@@ -51,46 +51,48 @@ func (r *Runtime) Start() error {
...
 	g.Go(func() error {
 		for {
 			select {
 			case data := <-dataSync:
-				r.updateWithNotify(data)
+				if resyncRequired := r.updateWithNotify(data); resyncRequired {
+					for _, s := range r.SyncImpl {
+						p := s
+						go func() {
+							g.Go(func() error {
```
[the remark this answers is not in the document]
james-milligan: RESYNC events may cycle until there is no longer any DELETE events, e.g. there is the possibility of `number of syncs -1` RESYNC cycles, once no delete events are found during an ALL merge the cycle will break. This is enforced by delete events requiring a matching source.

https://github.com/james-milligan/flagd/blob/8c2b391e9d632ad41ca70125aafa06678af49ed0/pkg/store/flags.go#L189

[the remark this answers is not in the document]
skyerus: Is it worth including this as a comment?

skyerus (approved):

beeme1mr:
Could you please add a section in the docs that describes the merge behavior. It can basically be a summary of the conversations we've had in this thread.

beeme1mr:
Please add some debug logging when a duplicate flag key is detected from another source. While the behavior defined in this PR makes sense, it would be very difficult to troubleshoot without any indication that merging occurred.

Issue #408, opened 2023-02-13
[FEATURE] Support Health and Readiness Probes in flagd


beeme1mr:
## Overview

In production environments, feature flags may be considered a mission critical service that must be available. It's important for flagd to expose a means for Kubernetes to know when the container is ready to perform flag evaluation. Also, needs to be aware if flagd is in an unhealth state so that it can take corrective measures. These settings should be configurable so that a Kubernetes cluster administrator can decide various error conditions should be handled.

## Requirements

- [x] Conduct research on the pros and cons of adding health and readiness probes.
- [x] Add health and readiness probes to flagd to monitor the state of the services.
- [x] Update the OpenFeature Operator to automatically configure the health and readiness checks.
- [x] Document the new behavior and configuration options in the OFO documentation.

## Resources

- https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

thisthat:
I had a look into this topic. We can re-use the HTTP server created for serving Prom metrics for the probes. This simplifies the implementation. Mind that K8s has a basic behavior regarding health probes: when it fails, it restarts the POD. Hence, we should not add any fancy configuration for this.
For the ready probe, I propose splitting the current `ISync` interface into two functions: `Init(Context)` and `Sync(Context, chan)`. This would also make a cleaner implementation for providers such as [gRPC](https://github.com/open-feature/flagd/blob/main/pkg/sync/grpc/grpc_sync.go).

PR #418, opened 2023-02-18
Add basic impl structure

thisthat:
Signed-off-by: Giovanni Liva <giovanni.liva@dynatrace.com>

## This PR

Add health and readiness probes to flagd to monitor the state of the services.

### Related Issues

#408

### Follow-up Tasks

Modify the OFO to use the newly exposed health and readiness probes.


pkg/runtime/runtime.go
```diff
@@ -89,6 +98,15 @@ func (r *Runtime) Start() error {
 	return nil
 }
 
+func (r *Runtime) isReady() bool {
+	// if at least a provider can watch for flags changes, we are ready.
+	pReady := false
+	for _, p := range r.SyncImpl {
+		pReady = pReady || p.IsReady()
```
skyerus, line 105: if pReady we can return straightaway

pkg/service/connect_service.go
```diff
@@ -107,15 +107,24 @@ func (s *ConnectService) setupServer() (net.Listener, error) {
...
+			switch r.URL.Path {
+			case "/healthz":
+				w.WriteHeader(http.StatusOK)
+			case "/readyz":
+				if svcConf.ReadinessProbe() {
+					w.WriteHeader(http.StatusOK)
+				} else {
+					w.WriteHeader(http.StatusPreconditionFailed)
+				}
+			case "/metric":
```
skyerus, line 125:
```suggestion
			case "/metrics":
```

pkg/sync/file/filepath_sync.go
```diff
@@ -22,31 +22,49 @@ type Sync struct {
...
+	err = fs.watcher.Add(fs.URI)
 	if err != nil {
 		return err
 	}
+	return nil
+}
 
-	fs.sendDataSync(ctx, sync.ALL, dataSync)
+func (fs *Sync) IsReady() bool {
+	fs.Mux.RLock()
```
skyerus, line 48: Is the mutex strictly necessary if only one goroutine can change the state?

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -198,8 +209,8 @@ func (k *Sync) notify(ctx context.Context, c chan<- INotify) {
 			EventType: DefaultEventTypeReady,
 		},
 	}
-
 	informer.Run(ctx.Done())
+	k.ready = true
```
skyerus, line 213: A better place for this may be on reception of the Ready event (L86)

pkg/sync/file/filepath_sync.go
thisthat: I added this because there were complains on race-conditions while running the tests

pkg/runtime/runtime.go
```diff
@@ -89,6 +98,16 @@ func (r *Runtime) Start() error {
 	return nil
 }
 
+func (r *Runtime) isReady() bool {
+	// if at least a provider can watch for flags changes, we are ready.
```
Kavindu-Dodan, line 102: In my view, we should check for all sync providers 🤔  

The reason is that, as per current ISync implementations, if the sync provider fails at the startup (first fetch), then this results in a flagd startup failure. So if we only check for one sync provider, it might mislead the readiness (this has the potential to impact OFO if we plan to utilize probes there).

thisthat: I had the assumption that, if I configure multiple providers, then they all return the same set of flags. Hence my implementation. Could it be the case that multiple providers see only a partition of the whole FF set?

toddbaert: > Could it be the case that multiple providers see only a partition of the whole FF set?

Yep! For instance, a different provider is set on different mounted-files/CRs. This is a common use case, since frequently there's a set of common flags shared by multiple services, but also flags unique to a single service. These may be respresented by different files/CRs, and also even come from different source types.

config/samples/example_flags.flagd.json
```diff
@@ -126,9 +126,9 @@
   },
   "$evaluators": {
     "emailWithFaas": {
-          "in": ["@faas.com", {
-            "var": ["email"]
-          }]
+      "in": ["@faas.com", {
+        "var": ["email"]
+      }]
```
toddbaert, line 131: :sunglasses:

pkg/runtime/from_config.go
```diff
@@ -67,52 +68,82 @@ func (r *Runtime) setSyncImplFromConfig(logger *logger.Logger) error {
...
-					zap.String("component", "sync"),
-					zap.String("sync", "remote"),
-				),
-				ProviderArgs: r.config.ProviderArgs,
-				Cron:         cron.New(),
-			})
+			r.SyncImpl = append(
+				r.SyncImpl,
+				r.newHTTP(uri, logger),
+			)
```
toddbaert, line 86: This looks much nicer, thanks.

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -30,6 +30,17 @@ type Sync struct {
...
+	ready        bool
+}
+
+func (k *Sync) Init(ctx context.Context) error {
+	// noop
+	return nil
+}
+
+func (k *Sync) IsReady() bool {
+	// we cannot reliably check external HTTP(s) sources
```
toddbaert, line 42: I think this comment is an error (it seems like it might have been copied from the http impl.

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -74,6 +85,7 @@ func (k *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 				k.Logger.Debug("configuration deleted")
 			case DefaultEventTypeReady:
 				k.Logger.Debug("notifier ready")
+				k.ready = true
```
toddbaert, line 88: @Kavindu-Dodan @thisthat do you guys think it's a bit weird that we will see the provider as ready after the first data is sent to the data-sync with the [initial fetch](https://github.com/open-feature/flagd/pull/418/files#diff-6b4b20bd0ff0a642f35bc03ca46986f7e27feb576ac3c459918d86f282a33642R48)?

pkg/sync/isync.go
```diff
@@ -38,9 +38,15 @@ ISync implementations watch for changes in the flag sources (HTTP backend, local
...
 type ISync interface {
+	// Init is used by the sync provider to initialize its data structures and external dependencies.
+	Init(ctx context.Context) error
+
 	// Sync is the contract between Runtime and sync implementation.
 	// Note that, it is expected to return the first data sync as soon as possible to fill the store.
 	Sync(ctx context.Context, dataSync chan<- DataSync) error
+
+	// IsReady shall return true if the provider is ready to communicate with the Runtime
+	IsReady() bool
```
toddbaert, line 49: This API makes sense to me.

pkg/sync/http/http_sync.go
```diff
@@ -37,6 +37,16 @@ type Cron interface {
...
 
+func (hs *Sync) Init(ctx context.Context) error {
+	// noop
+	return nil
+}
+
+func (hs *Sync) IsReady() bool {
+	// we cannot reliably check external HTTP(s) sources
+	return true
+}
```
toddbaert, line 48: Could we call the http endpoint as we are about to in the the sync anyway?

toddbaert:
@thisthat I like your API changes and this makes sense to me. I have a few thoughts in some of the sync implementations which I would like your feedback on, but none of this seems like a blocker to me.

pkg/runtime/runtime.go
thisthat: Thanks for the explanation! I will adapt the code accordingly :)

pkg/sync/kubernetes/kubernetes_sync.go
thisthat: Initially, I put it after the informer was configured. But I follow the suggestion of this [comment](https://github.com/open-feature/flagd/pull/418#discussion_r1111760603) since the event `DefaultEventTypeReady` cleary state the provider is ready 😅

pkg/sync/http/http_sync.go
thisthat: I am not sure. I have some concerns w.r.t. the performance. Given this will eventually be used for health/ready probes, we'd hit the HTTP endpoint too frequently. Usually, probes are tested every 5s. We would do an order of magnitude more HTTP calls than what is currently configured in the cron expression

toddbaert: What if we just checked it once, then cached the result? That would be similar to what is happening in the k8s sync, I think.

pkg/sync/kubernetes/kubernetes_sync.go
beeme1mr: FYI @skyerus. What do you think?

toddbaert: Ya I am conflicted on this. :thinking:

pkg/sync/http/http_sync.go
toddbaert: I guess the other thing to consider here is that the remote server _might_ be down only temporarily during the initial fetch, which could cause a cascading startup issue.

pkg/runtime/runtime.go
```diff
@@ -101,11 +101,11 @@ func (r *Runtime) Start() error {
 func (r *Runtime) isReady() bool {
 	// if at least a provider can watch for flags changes, we are ready.
 	for _, p := range r.SyncImpl {
-		if p.IsReady() {
-			return true
+		if !p.IsReady() {
```
toddbaert, line 104: Sometimes I wish go had a array reducer. I'm sorry if that's too JS of me :sweat_smile:

pkg/sync/http/http_sync.go
thisthat: I adapted the impl to match the K8s one :) please, have another look @toddbaert :)

toddbaert: Makes sense to me!

toddbaert (approved):
Approved, pending @thisthat confirms he's OK with my test changes: https://github.com/open-feature/flagd/pull/418#issuecomment-1442269373

thisthat:
> @thisthat I [pushed](https://github.com/open-feature/flagd/pull/418/commits/ffe8d64783da90b31185c93c81b7e158c8410d36) a [few](https://github.com/open-feature/flagd/pull/418/commits/ef550eeb41d593b1072afdc51a3c742a05d10f4c) small [changes](https://github.com/open-feature/flagd/pull/418/commits/acc9031e5d05cfaf502d84ff15b655840afa9821) to the sync tests which improves those tests (they were previously testing private funcs in some cases) and hopefully fixes the codecov faliure. I hope you dont mind.
> 
> EDIT: codecov is _nearly_ passing. I will add a tiny bit more.
> 
> EDIT 2: there we go! Passing now

Sure thing I am happy with your changes @toddbaert - that's why "allow edits by maintainers" is checked 😉 thanks for bumping the coverage 🙌

Kavindu-Dodan (approved):
Looks good and approving with minor chage suggestion

pkg/runtime/runtime.go
toddbaert, line 102:
```suggestion
	// if all providers can watch for flag changes, we are ready.
```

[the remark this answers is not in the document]
toddbaert: I made this change.

PR #443, opened 2023-02-27
feat: refactor and improve K8s sync provider

### LINKED ISSUE #434: Investigate K8s-go-client caching/load implications
Flagd is designed to watch K8s CRD changes and update flag configurations on CRD changes (ex:- updates, removals). This is performed by K8s ISync implementation [1]

The Focus of this task is to know the impact of on-demand CRD fetching on the K8s API. This knowledge is a pre-requisite to remove any caching in-between sync providers and flag evaluations/grpc streams. 

Research: 

- Does K8s API cache its responses 
- The impact of K8s API overloading (ex:- throttling, performance impact and stability)

Outcome:

Whether we require a cache OR whether we can rely on K8s API for high load scenarios 

[1] - https://github.com/open-feature/flagd/blob/main/pkg/sync/kubernetes/kubernetes_sync.go

Kavindu-Dodan:
## This PR
fixes #434

This is a refactoring and internal improvement for K8s ISync provider.

Improvements include,

- Reduce K8s API load by utilizing Informer cache
- Yet, provide a fallback if cache miss occurs (Note - we rely on K8s Informer for cache refill and consistency)
- Informer now only watches a specific namespace (compared to **\***) - this is a potential performance improvement and a security improvement
- Reduced informer handlers with extracted common logics 
- Unit tests where possible


thisthat (approved):
I left only minor suggestions to improve logging a bit. I see that parts of the code are not tested, in particular the one with K8s details. For that, I recommend having a look at the fake client [(docs)](https://pkg.go.dev/k8s.io/client-go/dynamic/fake)

pkg/sync/kubernetes/kubernetes_sync_test.go
```diff
@@ -0,0 +1,362 @@
+package kubernetes
```
toddbaert, line 1: Thanks for adding the tests. I kept meaning to add an issue to implement these and forgot to do it.

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -23,27 +22,67 @@ import (
...
+	if err != nil {
+		return err
+	}
+
+	k.readClient, err = client.New(clusterConfig, client.Options{Scheme: scheme.Scheme})
+	if err != nil {
+		return err
+	}
+
+	clusterClient, err := dynamic.NewForConfig(clusterConfig)
```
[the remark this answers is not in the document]
Kavindu-Dodan: changed in the latest commit

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -210,48 +195,52 @@ func (k *Sync) notify(ctx context.Context, c chan<- INotify) {
...
+// commonHandler emits the desired event if and only if handler receive an object matching apiVersion and resource name
+func commonHandler(obj interface{}, object client.ObjectKey, emitEvent DefaultEventType, c chan<- INotify) error {
+	ffObj, err := toFFCfg(obj)
 	if err != nil {
 		return err
 	}
-	if ffObj.APIVersion != fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version) {
+
+	if ffObj.APIVersion != apiVersion {
 		return errors.New("invalid api version")
```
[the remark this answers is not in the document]
Kavindu-Dodan: improved in the latest commit

toddbaert:
I tested this in OFO locally (latest version) and it works as expected.

pkg/sync/kubernetes/kubernetes_sync_test.go
Kavindu-Dodan: Thanks. Yes, I added tests for easily testable parts. However, we still need coverage on K8s-backed logic. These need either mocking or true e2e tests.

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -91,101 +130,47 @@ func (k *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
...
+// fetch attempts to retrieve the latest feature flag configurations
 func (k *Sync) fetch(ctx context.Context) (string, error) {
-	if k.URI == "" {
-		k.Logger.Error("no target feature flag configuration set")
-		return "{}", nil
-	}
-
-	ns, name, err := parseURI(k.URI)
+	// first check the store - avoid overloading API
+	object, exist, err := k.informer.GetStore().GetByKey(k.URI)
```
toddbaert, line 136:
```suggestion
	rawCfg, exist, err := k.informer.GetStore().GetByKey(k.URI)
```

It doesn't have to be this name, but I think implying this is unparsed JSON or a raw payload would be clearer.

Kavindu-Dodan:
> I left only minor suggestions to improve logging a bit. I see that parts of the code are not tested, in particular the one with K8s details. For that, I recommend having a look at the fake client [(docs)](https://pkg.go.dev/k8s.io/client-go/dynamic/fake)

Thank you. I addressed the log and renamed related comments. Regarding testing, K8s backed logic. Even with a fake client, we still not to mock the shared informer interface as we have a fallback. I am not sure how worth these mocks are against real e2e tests which easily test most of the logic.

pkg/sync/kubernetes/kubernetes_sync_test.go
toddbaert: We already have the "true" e2e tests in OFO. I think that's enough.

I think though we should improve coverage in this repo as well..

toddbaert:
I agree with @thisthat that we probably need a bit more coverage here (I know that there was no tests here before, so that's not your fault). I think now is the time. If there's any way to mock the k8s data I think we should do some basics here. It's worth another day or so of effort I think.

Maybe @thisthat has some specific recommendations.

Besides that everything here looks good to me and I think this is an improvement.

thisthat:
K8s provides great fake clients that can be used to unit-test your informer logic. 
I think a good starting point to see an example is this test file in Keptn: https://github.com/keptn/lifecycle-toolkit/blob/91e57cadba32fce6d873bc480408f90bcb8964d9/metrics-operator/cmd/metrics/adapter/provider/provider_test.go

toddbaert:
@Kavindu-Dodan we can create a new issue for additional testing. I think at a minimum we should get the coverage diff here to 0% with some meaningful tests unless it's really impossible. This isn't a place where we can afford to lose coverage.

pkg/sync/kubernetes/kubernetes_sync.go
Kavindu-Dodan: resolved with the latest commit

Kavindu-Dodan:
> @Kavindu-Dodan we can create a new issue for additional testing. I think at a minimum we should get the coverage diff here to 0% with some meaningful tests unless it's really impossible. This isn't a place where we can afford to lose coverage.

It is not impossible. It's time to deliver. Soon there will be merged conflicts with upcoming PRs and given that we constantly change API contracts, the efforts can go wasted.

Besides, I introduced new tests and simply touched code with the refactoring, reducing complexity. Even with new tests, for some reason, we see a reduction in coverage. To be honest I do not know how this happened.

Kavindu-Dodan:
> > @Kavindu-Dodan we can create a new issue for additional testing. I think at a minimum we should get the coverage diff here to 0% with some meaningful tests unless it's really impossible. This isn't a place where we can afford to lose coverage.
> 
> It is not impossible. It's time to deliver. Soon there will be merged conflicts with upcoming PRs and given that we constantly change API contracts, the efforts can go wasted.
> 
> Besides, I introduced new tests and simply touched code with the refactoring, reducing complexity. Even with new tests, for some reason, we see a reduction in coverage. To be honest I do not know how this happened.

To prove my point, here is a PR with the change we need - https://github.com/open-feature/flagd/pull/451

This will work and since I am not refactoring, code coverage won't complain.

toddbaert (approved):

beeme1mr (approved):

Kavindu-Dodan:
@thisthat @toddbaert  I added more tests however the coverage is below the previous value due to the reason I explained [1]

Please have another round of review :) 

[1] https://github.com/open-feature/flagd/pull/443#issuecomment-1450710014

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -273,22 +281,48 @@ func updateFuncHandler(oldObj interface{}, newObj interface{}, object client.Obj
 	return nil
 }
 
-func deleteFuncHandler(obj interface{}, object client.ObjectKey, c chan<- INotify) error {
+// toFFCfg attempts to covert unstructured payload to configurations
+func toFFCfg(object interface{}) (*v1alpha1.FeatureFlagConfiguration, error) {
```
james-milligan, line 285: if I add the following test to the `Test_toFFCfg` function:
```
{
			name: "Wrong resource type",
			input: toUnstructured(t, corev1.Pod{
				TypeMeta: v1.TypeMeta{
					Kind: "Pod",
				},
			}),
			wantErr: true,
		},
```
the test fails as there is no error returned, this is expected behaviour as the marshalling would not fail, however, IMO it should. We should add an if statement to ensure that the apiversion and kind match at this point, of course this would require the k8s api to return us the wrong object, however I think its worthwhile covering this case as it currently passes with any object

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -220,48 +224,52 @@ func (k *Sync) notify(ctx context.Context, c chan<- INotify) {
...
-	err := runtime.DefaultUnstructuredConverter.FromUnstructured(u.Object, &ffOldObj)
+	ffOldObj, err := toFFCfg(oldObj)
 	if err != nil {
 		return err
 	}
-	if ffOldObj.APIVersion != fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version) {
-		return errors.New("invalid api version")
+
+	if ffOldObj.APIVersion != apiVersion {
+		return fmt.Errorf("invalid api version %s, expected %s", ffOldObj.APIVersion, apiVersion)
```
james-milligan, line 261: can duplicate `"Simple error - API version mismatch"` test and change the api version in `validFFCfgOld` to cover this line

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -2,16 +2,15 @@ package kubernetes
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"strings"
+	parallel "sync"
```
james-milligan, line 8: nitpick but when the flagd sync package and `"sync"` are being imported in the same package it  has previously been imported as "msync", might cause confusion to change the name in this package, maybe?

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -101,101 +159,47 @@ func (k *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
...
-		return "{}", nil
-	}
-
-	ns, name, err := parseURI(k.URI)
+	// first check the store - avoid overloading API
+	item, exist, err := k.informer.GetStore().GetByKey(k.URI)
 	if err != nil {
-		k.Logger.Error(err.Error())
-		return "{}", nil
+		return "", err
```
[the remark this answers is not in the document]
Kavindu-Dodan: Return value consists of (flagConfig, error). And if there is an error, return the value of flagConfig must be ignored (error handling). This is why the simple empty string return here.

pkg/sync/kubernetes/kubernetes_sync.go
Kavindu-Dodan: Yes, noticed this while testing the function. However, I will keep this as it is. The reason is that if we try to validate, we have to enforce validation criteria. API version is only one such validation. So it is best to allow the caller to perform such validation based on its requirement and have this common logic of parsing here.

skyerus (approved):

toddbaert:
@Kavindu-Dodan @thisthat I created https://github.com/open-feature/flagd/issues/460 for testing.

Issue #435, opened 2023-02-24
Research versioning (maybe separate) and releasing for kube-flagd-proxy and flagd


beeme1mr:
The Release Please configuration would have to be updated. It may require a [workspace plugin](https://github.com/googleapis/release-please/blob/main/src/plugins/workspace.ts) to be configured in order to properly bump versions when a local dependency changes.

skyerus:
### Versioning solutions

Ideally, `flagd` & `kube-flagd-proxy`'s versioning should be entirely separate (e.g. to ensure a breaking change in one doesn't enforce a semantic breaking change in the other).

#### Single Go module

Using release-please to manage the release process of `flagd`/`kube-flagd-proxy` as a single Go module monorepo has the limitation of not knowing how to handle changes of shared local dependencies (`pkg/`). There are release-please plugins available to manage this in other languages (e.g. node) but not Go. It also doesn't seem possible to implement a custom plugin, or if it is there is no documentation on how to do this and scouring the web for any custom plugins comes up empty.

Example implementation: https://github.com/open-feature/flagd/pull/445

#### Multiple Go modules

Separating` flagd`, `kube-flagd-proxy` & `internal` (module of shared dependencies of `flagd`/`kube-flagd-proxy`) into their own Go modules managed by release-please as a monorepo (in similar to fashion as what is in place in go-sdk-contrib) provides the means to upgrade versions independently. If a shared dependency (under the `internal` module) has a breaking change, `flagd` &` kube-flagd-proxy` can integrate it and make its own decision about whether it is a breaking change for that module.
A downside of this method is that local development will require the use of `replace ()` in the go.mod file to use the in development shared dependencies. Also, changes to the shared dependencies will have to be released and integrated into `flagd`/`kube-flagd-proxy` in a separate step (otherwise CI will fail to build), this creates more friction in the development process.

Example implementation: https://github.com/open-feature/flagd/pull/447

### Opinion

If there were a Go plugin for release-please (or means to create one easily) I would opt for the single Go module architecture. Without possibility for this plugin I opt for the multiple Go modules architecture.

skyerus:
The approach of **Multiple Go modules** also provides the additional benefit of being able to expose packages to be used by components outside of `flagd`.

What are your thoughts?
@beeme1mr @toddbaert @james-milligan @Kavindu-Dodan

beeme1mr:
Multiple Go modules seems like a good approach to me. It would be nice to sperate each core component so that they can be used independently.

[...]

PR #445, opened 2023-02-28
feat: restructure flagd to flagd & kube-flagd-proxy

skyerus:
## This PR

- Restructures flagd to flagd & kube-flagd-proxy

### Related Issues

https://github.com/open-feature/flagd/issues/435


release-please-config.json
```diff
@@ -1,16 +1,24 @@
 {
+  "monorepo-tags": true,
+  "tag-separator": "/",
   "bootstrap-sha": "c4e030ed651ab800aa56a555471da5913eb95259",
   "packages": {
-    ".": {
+    "cmd/flagd": {
```
beeme1mr, line 6: How will shared resources work? Do you know if dependencies can be defined?

skyerus: This only works if we are able to leverage a plugin that parses the dependency tree. There exists one in node but not Go, I've written more about this [here](https://github.com/open-feature/flagd/issues/435#issuecomment-1448576876).

beeme1mr: Yeah, sorry about that. I just noticed you include a nice write up already in the issue.

Issue #405, opened 2023-02-13
Research the best way to support advanced configuration per sync provider


toddbaert:
flagd supports various sync providers (filepath, kubernetes, http, grpc), and also supports the usage of multiple providers simultaneously (just as an example, flagd can be started to listen to an HTTP endpoint, watch a file, and connect to 2 gRPC servers).

Each of these providers may require individual configuration. For instance, each gRPC stream might need different certs, or http syncs may each require different auth headers. We currently do not have a solution to handle this sort of configuration from the command line.

We need to decide on a way to support this. It should also work well with CRs we define in the OpenFeature operator.

Ideas: 

- the use of multiple `syncProvider` args, supporting key/value encoding of some kind
- defining a configuration file structure and passing that to flagd
- ?

Edit: See Jame's research/suggestions below. I'm somewhat partial to [this](https://github.com/open-feature/flagd/issues/405#issuecomment-1430109194) one.

Implementation issue: https://github.com/open-feature/open-feature-operator/issues/353

james-milligan:
I like the idea of a config file, its a pretty common pattern that would allow for developers to very easily interpret configurations (compared to a multi line list of startup flags), we can also create a schema definition for it to allow for linting + validation

james-milligan:
Ive done some more investigation into a clean method for handling per-sync configurations:

Currently we are using viper to handle our flag values, out of the box viper allows for the parsing of a config file (.json/.yaml), we already have this configuration set up within `flagd` under the `--config` flag.
This means we can provide a config file such as:

```
port: 8080
sync-provider-args:
    food: bars
    cat: dog
```

resulting in each sync provider receiving the following `map[string]string`: `map[cat:dog food:bars]`

This flag structure feels inappropriate for our use case, instead we can update the `sync.ProviderArgs` type from `map[string]string` to `map[string]interface{}` and decode the `--sync-provider-args` flag value as `map[string]sync.ProviderArgs`
```
conf := &map[string]sync.ProviderArgs{}
err = viper.UnmarshalKey(providerArgsFlagName, conf)
```
this means we can now pass the following configuration file:
```
sync-provider-args:
    sync-uri-1:
        food: 5
        cat: dog
        fizz: 
            buzz: 7
```
this allows us to define configuration on a per-sync basis, using the sync URI as the map key, which we can filter for on startup, with each configuration defined as `map[string]interface{}`, to be handled internally by each sync as required.

[...]

PR #448, opened 2023-02-28
feat: introduce per-sync configurations

james-milligan:
## This PR

- introduces the `SyncProviderConfig` object and associated start flags to pass configuration to specific sync providers

### Related Issues

https://github.com/open-feature/flagd/issues/405

### DIFF
--- a/config/samples/example_flags.flagd.json
+++ b/config/samples/example_flags.flagd.json
@@ -122,6 +122,29 @@
           }, null
         ]
       }
+    },
+    "targetedFlag": {
+      "variants": {
+        "first": "AAA",
+        "second": "BBB",
+        "third": "CCC"
+      },
+      "defaultVariant": "first",
+      "state": "ENABLED",
+      "targeting": {
+        "if": [{
+            "in": ["@openfeature.dev", {
+              "var": "email"
+            }]
+          }, "second",
+          {
+            "in": ["Chrome", {
+              "var": "userAgent"
+            }]
+          }, "third",
+          null
+        ]
+      }
     }
   },
   "$evaluators": {
--- a/config/samples/example_flags.flagd.yaml
+++ b/config/samples/example_flags.flagd.yaml
@@ -80,6 +80,24 @@ flags:
         - - yellow
           - 25
       - null
+  targetedFlag:
+    variants:
+      first: "AAA"
+      second: "BBB"
+      third: "CCC"
+    defaultVariant: first
+    state: ENABLED
+    targeting:
+if:
+- in:
+  - "@openfeature.dev"
+  - var: email
+- second
+- in:
+  - Chrome
+  - var: userAgent
+- third
+-
 "$evaluators":
   emailWithFaas:
     in:


pkg/runtime/from_config.go
```diff
@@ -109,41 +118,88 @@ func (r *Runtime) newGRPC(uri string, logger *logger.Logger) *grpc.Sync {
 	}
 }
 
-func (r *Runtime) newHTTP(uri string, logger *logger.Logger) *httpSync.Sync {
+func (r *Runtime) newHTTP(config sync.ProviderConfig, logger *logger.Logger) *httpSync.Sync {
```
james-milligan, line 121: Ive opted to not introduce unit tests for these contractors, although happy to if code coverage is required

cmd/start.go
```diff
@@ -96,26 +103,38 @@ var startCmd = &cobra.Command{
...
 		if viper.GetString(evaluatorFlagName) != "json" {
 			rtLogger.Warn("DEPRECATED: The --evaluator flag has been deprecated. " +
 				"Docs: https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md")
 		}
+
+		syncProviders, err := runtime.SyncProvidersFromURIs(viper.GetStringSlice(uriFlagName))
+		if err != nil {
+			log.Fatal(err)
+		}
+		syncProviders2 := []sync.ProviderConfig{}
```
skyerus, line 115: Is there a more appropriate name for this var?

pkg/runtime/from_config.go
```diff
@@ -64,43 +73,43 @@ func (r *Runtime) setService(logger *logger.Logger) {
...
+		switch syncProvider.Provider {
+		case syncProviderFile:
 			r.SyncImpl = append(
 				r.SyncImpl,
-				r.newFile(uri, logger),
+				r.newFile(syncProvider, logger),
 			)
-			rtLogger.Debug(fmt.Sprintf("using filepath sync-provider for: %q", uri))
-		case regCrd.Match(uriB):
+			rtLogger.Debug(fmt.Sprintf("using filepath sync-provider for: %q", syncProvider.URI))
```
skyerus, line 84:
```suggestion
			rtLogger.Debug(fmt.Sprintf("using filepath sync-provider for: %s", syncProvider.URI))
```

pkg/runtime/from_config.go
```diff
@@ -64,43 +73,43 @@ func (r *Runtime) setService(logger *logger.Logger) {
...
+			rtLogger.Debug(fmt.Sprintf("using kubernetes sync-provider for: %s", syncProvider.URI))
+		case syncProviderHTTP:
 			r.SyncImpl = append(
 				r.SyncImpl,
-				r.newHTTP(uri, logger),
+				r.newHTTP(syncProvider, logger),
 			)
-			rtLogger.Debug(fmt.Sprintf("using remote sync-provider for: %q", uri))
-		case regGRPC.Match(uriB):
+			rtLogger.Debug(fmt.Sprintf("using remote sync-provider for: %q", syncProvider.URI))
```
skyerus, line 96:
```suggestion
			rtLogger.Debug(fmt.Sprintf("using remote sync-provider for: %s", syncProvider.URI))
```

beeme1mr (commented):
I wasn't able to get the examples working. I built flagd using your branch and confirmed the config example flags were in the expected location.

```sh
./flagd start --sync-providers=\[{\"uri\":\"config/samples/example_flags.json\"\,\"provider\":\"file\"}\]

                 ______   __       ________   _______    ______      
                /_____/\ /_/\     /_______/\ /______/\  /_____/\     
                \::::_\/_\:\ \    \::: _  \ \\::::__\/__\:::_ \ \    
                 \:\/___/\\:\ \    \::(_)  \ \\:\ /____/\\:\ \ \ \   
                  \:::._\/ \:\ \____\:: __  \ \\:\\_  _\/ \:\ \ \ \  
                   \:\ \    \:\/___/\\:.\ \  \ \\:\_\ \ \  \:\/.:| | 
                    \_\/     \_____\/ \__\/\__\/ \_____\/   \____/_/ 

2023-03-01T14:01:42.843-0500    info    cmd/start.go:104        flagd version: dev (5ecdb19), built at: 2023-03-01T14:00:02Z    {"component": "start"}
2023-03-01T14:01:42.843-0500    fatal   cmd/start.go:144        no sync implementation set      {"component": "start"}
github.com/open-feature/flagd/cmd.glob..func1
        /home/beemer/code/openfeature/flagd/cmd/start.go:144
github.com/spf13/cobra.(*Command).execute
        /home/beemer/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:920
github.com/spf13/cobra.(*Command).ExecuteC
        /home/beemer/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:1044
github.com/spf13/cobra.(*Command).Execute
        /home/beemer/go/pkg/mod/github.com/spf13/cobra@v1.6.1/command.go:968
github.com/open-feature/flagd/cmd.Execute
        /home/beemer/code/openfeature/flagd/cmd/root.go:37
main.main
        /home/beemer/code/openfeature/flagd/main.go:30
runtime.main
        /home/beemer/sdk/go1.19.6/src/runtime/proc.go:250
```

docs/configuration/flagd_start.md
```diff
@@ -21,6 +21,7 @@ flagd start [flags]
   -d, --socket-path string                  Flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
   -y, --sync-provider string                DEPRECATED: Set a sync provider e.g. filepath or remote
   -a, --sync-provider-args stringToString   Sync provider arguments as key values separated by = (default [])
```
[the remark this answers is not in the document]
james-milligan: done

docs/configuration/flagd_start.md
```diff
@@ -21,6 +21,7 @@ flagd start [flags]
...
+  -s, --sync-providers string               JSON representation of an array of ProviderConfig objects. This object contains 2 required fields, uri (string) and provider (string). Documentation for this object can be found here:
```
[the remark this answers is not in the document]
james-milligan: resolved

pkg/sync/file/filepath_sync.go
```diff
@@ -17,10 +17,10 @@ import (
 )
 
 type Sync struct {
-	URI          string
-	Source       string
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
+	URI    string
+	Logger *logger.Logger
+	Config sync.ProviderConfig
```
Kavindu-Dodan, line 22: Do we use this Configuration inside ISync implementation?  🤔

pkg/sync/isync.go
```diff
@@ -59,3 +57,10 @@ type DataSync struct {
 	Source   string
 	Type
 }
+
+type ProviderConfig struct {
```
Kavindu-Dodan, line 61: should we focus on a per-provider configuration rather than a single one ? 🤔  For example, if we now introduce grpc specific configuration, then this get passed to all sync provider types

Kavindu-Dodan:
General observation while attempting to introduce grpc tls connection parameters. My view is that,

- we need provider-specific configuration so that it is easy to decode related configurations and build providers based on the specific configuration
- avoid parsing the configuration down to the implementation level. we should either simply use struct initialization or use a NewSyncX(configuration) mechanism.

beeme1mr:
> General observation while attempting to introduce grpc tls connection parameters. My view is that,
> 
> * we need provider-specific configuration so that it is easy to decode related configurations and build providers based on the specific configuration
> * avoid parsing the configuration down to the implementation level. we should either simply use struct initialization or use a NewSyncX(configuration) mechanism.

The seems reasonable to me.

toddbaert:
> General observation while attempting to introduce grpc tls connection parameters. My view is that,
> 
> * we need provider-specific configuration so that it is easy to decode related configurations and build providers based on the specific configuration
> * avoid parsing the configuration down to the implementation level. we should either simply use struct initialization or use a NewSyncX(configuration) mechanism.

Are you saying we should probably merge this (or something like it) before adding back TLS support to syncs?

pkg/sync/isync.go
james-milligan: You're right that the field will be passed to all provider types, but it would be a nil reference of the type. IMO I don't see any issue with using a shared data structure - however, if we feel there is a need to introduce one im happy to make these changes. 

Introducing per-sync configuration objects will also add complexity to the passing / parsing of the configurations as startup commands, which is the main argument for using a single shared type (unmarshal directly into a single shared slice, rather than having a []typeXYZ{} slice to accommodate every new data type, or storing all configurations as interfaces and inferring their types on startup)

james-milligan:
> avoid parsing the configuration down to the implementation level. we should either simply use struct initialization or use a NewSyncX(configuration) mechanism.

Do you mind explaining this last point

pkg/sync/isync.go
james-milligan: Another option is to map between the 'general' configuration structure and a sync specific one in the NewXYZ(cofiguration) methods, but this option feels inappropriate

Kavindu-Dodan:
@toddbaert 

> Are you saying we should probably merge this (or something like it) before adding back TLS support to syncs?

Yes, we need to finalize how we obtain configurations (needs this for the cert path argument). One option is to introduce a dedicated flag and later add this flag to the mechanism this PR introduces (sync provider specific configurations). But then we need to agree whether **we will support both**: i.e - Expose startup option (ex:-  bearer token, tls cert path ...) & provider specific option of this PR.

p.s - I will avoid the configuration changes for now and work on the implementation level and open a draft PR 

@james-milligan 

> Do you mind explaining this last point, this PR implements the NewSyncX(configuration) option

Sorry if this was not clear. What I said was to introduce a **New** function to build the ISync struct and avoid having the configuration structure **embedded** in ISync impls - something like `NewK8sWithConfig`. This is to avoid having unrelevant configuration options (ex:- a nil bearer token inside the file ISync provider). Plus, we can enforce validations through **New** function (fail fast)

Plus, having such a contract helps to improve how we handle configurations. Let's say we use a common config object and use the `New` function. Given that we do not bind this configuration object to internals, our next improvement can focus on per sync-specific configuration object (instead of the global one). From ISync implementation point of view, there will be no change. We just need to have the parsing mechanism, create sync specific config object (i.e- K8sSyncConfig, fileSyncConfig, grpcSyncConfig) and simply introduce a `New` function to handle this.  

Anyway, this is my opinion 😉  open for opinions or discussions. I do not want to delay this PR further.

docs/configuration/flagd_start.md
```diff
@@ -20,7 +20,8 @@ flagd start [flags]
   -k, --server-key-path string              Server side tls key path
   -d, --socket-path string                  Flagd socket path. With grpc the service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
   -y, --sync-provider string                DEPRECATED: Set a sync provider e.g. filepath or remote
-  -a, --sync-provider-args stringToString   Sync provider arguments as key values separated by = (default [])
+  -a, --sync-provider-args stringToString   DEPRECATED: Sync provider arguments as key values separated by = (default [])
+  -s, --sync-providers string               JSON representation of an array of ProviderConfig objects. This object contains 2 required fields, uri (string) and provider (string). Documentation for this object can be found here: https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md#sync-provider-customisation
```
toddbaert, line 24: I could go either way, but I'm wondering if `sources` is better language for this. I think it comes down to if we want to think of `sync-providers` as more of an internal detail to flagd or not. Currently, I think we use `sources` language in OFO. I'm interested in everyone's thoughts.

cmd/start.go
```diff
@@ -97,26 +105,44 @@ var startCmd = &cobra.Command{
 
 		rtLogger.Info(fmt.Sprintf("flagd version: %s (%s), built at: %s", Version, Commit, Date))
 
-		if viper.GetString(syncProviderFlagName) != "" {
```
toddbaert, line 100: Did we remove this deprecation?

cmd/start.go
```diff
@@ -97,26 +105,44 @@ var startCmd = &cobra.Command{
...
+		syncProviders, err := runtime.SyncProvidersFromURIs(viper.GetStringSlice(uriFlagName))
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		syncProvidersFromConfig := []sync.ProviderConfig{}
+		if cfgFile == "" && viper.GetString(syncProvidersFlagName) != "" {
+			syncProvidersFromConfig, err = runtime.SyncProviderArgPass(viper.GetString(syncProvidersFlagName))
+			if err != nil {
+				log.Fatal(err)
```
toddbaert, line 127: The feedback we get here seems fine. When I mucked with the JSON the errors were clear enough.

cmd/start.go
```diff
@@ -97,26 +105,44 @@ var startCmd = &cobra.Command{
...
+			}
+		} else {
+			err = viper.UnmarshalKey(syncProvidersFlagName, &syncProvidersFromConfig)
+			if err != nil {
+				log.Fatal(err)
```
toddbaert, line 132: The feedback we get here seems fine. When I mucked with the JSON the errors were clear enough.

cmd/start.go
```diff
@@ -97,26 +105,44 @@ var startCmd = &cobra.Command{
...
+			}
+		}
+		syncProviders = append(syncProviders, syncProvidersFromConfig...)
```
toddbaert, line 135: The feedback we get here seems fine. When I mucked with the JSON the errors were clear enough.

pkg/runtime/from_config.go
```diff
@@ -112,41 +125,88 @@ func (r *Runtime) newGRPC(uri string, logger *logger.Logger) *grpc.Sync {
...
 			zap.String("sync", "filepath"),
 		),
-		ProviderArgs: r.config.ProviderArgs,
-		Mux:          &msync.RWMutex{},
+		Config: config,
+		Mux:    &msync.RWMutex{},
+	}
+}
+
+func SyncProviderArgPass(syncProviders string) ([]sync.ProviderConfig, error) {
```
toddbaert, line 166:
```suggestion
func SyncProviderArgValid(syncProviders string) ([]sync.ProviderConfig, error) {
```

pkg/runtime/from_config.go
toddbaert: I'll leave that up to you. I'd prefer to keep the codecov happy, but if some of the additions are troublesome to tests it's fine with me. I think you've got most of the important stuff in this PR.

toddbaert (approved):
Approved with minor feedback after local testing. I do think I like @Kavindu-Dodan 's idea [here](https://github.com/open-feature/flagd/pull/448#issuecomment-1456580414) though, and think it's worth consideration if you agree.... We could also do it in a new PR I guess.

cmd/start.go
james-milligan: reintroduced, nice catch

pkg/runtime/from_config.go
james-milligan: renamed to SyncProviderArgParse

docs/configuration/flagd_start.md
james-milligan: ive introduced this change

cmd/start.go
```diff
@@ -58,6 +60,11 @@ func init() {
...
 		syncProviderFlagName, "y", "", "DEPRECATED: Set a sync provider e.g. filepath or remote",
 	)
+	flags.StringP(
+		sourcesFlagName, "s", "", "JSON representation of an array of ProviderConfig objects. This object contains "+
+			"2 required fields, uri (string) and provider (string). Documentation for this object can be found here: "+
+			"https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md#sync-provider-customisation",
+	)
 	flags.StringP(logFormatFlagName, "z", "console", "Set the logging format, e.g. console or json ")
 
 	_ = viper.BindPFlag(bearerTokenFlagName, flags.Lookup(bearerTokenFlagName))
```
Kavindu-Dodan, line 70: This must be removed right?

pkg/sync/http/http_sync.go
```diff
@@ -15,15 +15,13 @@ import (
...
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
-
-	ready bool
+	URI         string
+	Client      Client
+	Cron        Cron
+	LastBodySHA string
+	Logger      *logger.Logger
+	ready       bool
```
Kavindu-Dodan, line 23: Sorry for nitpicking, shall we separate **P**ublic and **p**rivate fields 

```suggestion
	URI         string
	Client      Client
	Cron        Cron
	LastBodySHA string
	Logger      *logger.Logger
	BearerToken string
	
	ready       bool
```

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -28,10 +28,10 @@ var (
 )
 
 type Sync struct {
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
-	URI          string
-
+	Logger     *logger.Logger
```
Kavindu-Dodan, line 31: Same with ordering 🥺  ?

Kavindu-Dodan:
Overal looks good 👍

docs/configuration/configuration.md
```diff
@@ -35,4 +34,29 @@ To use an existing FeatureFlagConfiguration custom resource, start flagD with th
...
+|------------|------------------------------------|
+| uri | required `string` |  |
+| provider   | required `string` (`file`, `kubernetes`, `http` or `grpc`) |
+| bearerToken     | optional `string`        |
+
+The `uri` field values do not need to follow the [URI patterns](#uri-patterns), the provider type is instead derived from the provider field.
+
+Example start command using a filepath sync provider and the equivalent config file definition:
+```sh
+./flagd start --sources=\[{\"uri\":\"config/samples/example_flags.json\"\,\"provider\":\"file\"}\]
```
Kavindu-Dodan, line 55: Not sure if this is only my machine (Mac zsh). I was able to use the start command simply without escaping. Used single quotes 🤔 

ex:- 
`go run main.go start --sources='[{"uri":"config/samples/example_flags.json","provider":"file"}]' --debug
`

docs/configuration/configuration.md
```diff
@@ -35,4 +34,29 @@ To use an existing FeatureFlagConfiguration custom resource, start flagD with th
 
 ```shell
 flagd start --uri core.openfeature.dev/default/my_example
+```
+
+### Source Configuration
+
+While a URI may be passed to flagd via the `--uri` flag, some implementations may require further configurations. In these cases the `--sources` flag should be used.
+The flag takes a string argument, which should be a JSON representation of an array of `ProviderConfig` objects. Alternatively, these configurations should be passed to
```
beeme1mr, line 42: Should we call them `SourceConfig` objects?

docs/configuration/configuration.md
```diff
@@ -35,4 +34,29 @@ To use an existing FeatureFlagConfiguration custom resource, start flagD with th
...
+flagd via config file, specified using the `--config` flag.
+
+| Field       | Type  | 
+|------------|------------------------------------|
+| uri | required `string` |  |
+| provider   | required `string` (`file`, `kubernetes`, `http` or `grpc`) |
+| bearerToken     | optional `string`        |
+
+The `uri` field values do not need to follow the [URI patterns](#uri-patterns), the provider type is instead derived from the provider field.
```
beeme1mr, line 51: What would happen if the `file` uri prefix was included? Can you give more examples below? I'm particularly interested in how the HTTP source config should look but including one for gRPC and Kubernetes would be useful as well.

docs/configuration/configuration.md
beeme1mr: single quotes worked for me in Ubuntu on WSL2.

cmd/start.go
```diff
@@ -106,17 +114,40 @@ var startCmd = &cobra.Command{
 			rtLogger.Warn("DEPRECATED: The --evaluator flag has been deprecated. " +
 				"Docs: https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md")
 		}
+
+		if viper.GetStringMapString(providerArgsFlagName) != nil {
+			rtLogger.Warn("DEPRECATED: The --sync-provider-args flag has been deprecated. " +
```
beeme1mr, line 119: I'm seeing this warning even when I don't pass `sync-provider-args`. Could you please double-check the logic? I would expect to only see this when it's actually being used.

beeme1mr (commented):
It's probably unrelated to this change but I'm seeing two WRITE events in the logs every time a flag configuration changes using the file source. Could you please confirm that this change isn't the root cause. If it's not, please create a follow up issue to investigate.

cmd/start.go
james-milligan: switched to a string check

docs/configuration/configuration.md
james-milligan: Added logic to parse `file:` and `core.openfeature.dev` prefixes, and added a reference to this in the docs

james-milligan:
> It's probably unrelated to this change but I'm seeing two WRITE events in the logs every time a flag configuration changes using the file source. Could you please confirm that this change isn't the root cause. If it's not, please create a follow up issue to investigate.

Please can you provide more info - im not able to replicate this behaviour:

```
2023-03-08T09:27:41.069Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:31.740Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json CHMOD {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:31.741Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json WRITE {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:31.741Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:36.256Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json CHMOD {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:36.256Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json WRITE {"component": "sync", "sync": "filepath"}
2023-03-08T09:28:36.256Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
```

I get repeated patterns of `CHMOD`, `WRITE` then a configuration `ALL` event using this startup command:
```
go run main.go start --sources='[{"uri":"config/samples/example_flags.json","provider":"file"}]' --debug 
```

cmd/start.go
```diff
@@ -58,6 +60,11 @@ func init() {
 	flags.StringP(
 		syncProviderFlagName, "y", "", "DEPRECATED: Set a sync provider e.g. filepath or remote",
 	)
+	flags.StringP(
+		sourcesFlagName, "s", "", "JSON representation of an array of SourceConfig objects. This object contains "+
+			"2 required fields, uri (string) and provider (string). Documentation for this object can be found here: "+
```
skyerus, line 65:
```suggestion
			"2 required fields, uri (string) and provider (string). Documentation for this object: "+
```

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -28,9 +28,9 @@ var (
 )
 
 type Sync struct {
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
-	URI          string
+	Logger *logger.Logger
+	Source string
```
skyerus, line 32: Is this definitely needed? Thought it was removed recently

pkg/sync/file/filepath_sync.go
```diff
@@ -17,9 +17,9 @@ import (
 )
 
 type Sync struct {
-	URI          string
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
+	URI    string
+	Logger *logger.Logger
+	Source string
```
skyerus, line 22: Needed?

cmd/start.go
james-milligan: deprecated :)

toddbaert:
> > It's probably unrelated to this change but I'm seeing two WRITE events in the logs every time a flag configuration changes using the file source. Could you please confirm that this change isn't the root cause. If it's not, please create a follow up issue to investigate.
> 
> Please can you provide more info - im not able to replicate this behaviour:
> 
> ```
> 2023-03-08T09:27:41.069Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:31.740Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json CHMOD {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:31.741Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json WRITE {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:31.741Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:36.256Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json CHMOD {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:36.256Z        info    file/filepath_sync.go:78        filepath event: config/samples/example_flags.json WRITE {"component": "sync", "sync": "filepath"}
> 2023-03-08T09:28:36.256Z        debug   file/filepath_sync.go:121       Configuration config/samples/example_flags.json:  ALL   {"component": "sync", "sync": "filepath"}
> ```
> 
> I get repeated patterns of `CHMOD`, `WRITE` then a configuration `ALL` event using this startup command:
> 
> ```
> go run main.go start --sources='[{"uri":"config/samples/example_flags.json","provider":"file"}]' --debug 
> ```

I think this is all old behavior, and not much can be done about it. It seems to be all about how particular editors and programs update files. Even in the most basic cases, programs actually remove/write the file. The only exception is if the file being watched is a symlink and that symlink is updated. That's the only thing that seems to be completely atomic.

Even if I run something like: `sed -i 's/"defaultVariant": "on"/"defaultVariant": "off"/g' config/samples/example_flags.json` which uses `sed`'s "in-place" option (`-i`), the file is actually removed and replaced with new contents (I had not idea it worked that way, but [apparently it does](https://github.com/fsnotify/fsnotify/issues/143))

toddbaert (approved):
Re-tested and re-approved.

skyerus (approved):

PR #455, opened 2023-03-02
docs: configuration merge docs

james-milligan:
## This PR

- adds documentation for merge events + resyncs

### Related Issues

https://github.com/open-feature/flagd/issues/452

PR #465, opened 2023-03-03
fix: set readiness once only

Kavindu-Dodan:
## This PR

The current readiness state can be changed because of an underlying sync provider error. This seems to be incorrect as **not ready state** make flagd to not receive traffic through K8s services [1]

For example, if HTTP endpoint-backed sync provider get an error, the current logic set readiness to `false`. Similarly, if grpc connectivity is lost for grpc provider, readiness goes to `false` state till connection establishment completes. 

This PR fix this behavior by setting the ready state **once only** per sync provider. This means if there is at least a single successful sync, sync provider sets its readiness to `true`.

Going further, we might need to decide if we need better readiness state handling. 

[1] - https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes

### DIFF
--- a/pkg/sync/http/http_sync.go
+++ b/pkg/sync/http/http_sync.go
@@ -62,6 +62,8 @@ func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 	if err != nil {
 		return err
 	}
+
+	// Set ready state
 	hs.ready = true
 
 	_ = hs.Cron.AddFunc("*/5 * * * *", func() {
@@ -149,18 +151,16 @@ func (hs *Sync) generateSha(body []byte) string {
 
 func (hs *Sync) Fetch(ctx context.Context) (string, error) {
 	if hs.URI == "" {
-		hs.ready = false
 		return "", errors.New("no HTTP URL string set")
 	}
 
 	body, err := hs.fetchBodyFromURL(ctx, hs.URI)
 	if err != nil {
-		hs.ready = false
 		return "", err
 	}
 	if len(body) != 0 {
 		hs.LastBodySHA = hs.generateSha(body)
 	}
-	hs.ready = true
+
 	return string(body), nil
 }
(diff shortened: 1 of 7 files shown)


thisthat:
The readiness state should change since this tells K8s when traffic can be scheduled to that Pod, but it depends on our semantics of ready state. With this change, we lost consistency in favor of availability and partition tolerance. Given that flagd is injected as a sidecar by OFO, I suggest going ahead with this change since it makes it more resilient and doesn't block traffic to the original pod.

toddbaert:
I may have misunderstood the first PR introducing health checks or misread the code:

> This PR fix this behavior by setting the ready state once only per sync provider. This means if there is at least a single successful sync, sync provider sets its readiness to true.

I thought this was already the way it worked :sweat_smile:

pkg/sync/http/http_sync.go
```diff
@@ -151,18 +153,16 @@ func (hs *Sync) generateSha(body []byte) string {
...
-		hs.ready = false
 		return "", err
 	}
 	if len(body) != 0 {
 		hs.LastBodySHA = hs.generateSha(body)
 	}
-	hs.ready = true
+
 	return string(body), nil
 }
```
toddbaert, line 168: I don't know how I missed this before :facepalm:

toddbaert:
Can we add some tests to establish expectations around this behavior? Perhaps it will be easier once [this](https://github.com/open-feature/flagd/pull/466) is merged?

beeme1mr:
Can you please include a section in the docs that describes the behavior in a follow up PR? An example of what I'm looking for can be found [here](https://github.com/open-feature/flagd/pull/455). Users should be able to understand what endpoints are available and the behavior to expect when using them.

skyerus (approved):
Nice find

pkg/sync/http/http_sync.go
Kavindu-Dodan: I also did not think about the implications of the ready state when we introduced the probe.

Kavindu-Dodan:
@toddbaert I added readiness state validation except for K8s

beeme1mr (approved):

PR #466, opened 2023-03-04
chore: refactoring and improve coverage for K8s Sync

### LINKED ISSUE #460: [Testing] Increase test coverate in K8s sync
As discussed [here](https://github.com/open-feature/flagd/pull/443#issuecomment-1449869155), we can increase coverage of this sync.

Try to get the coverage of this implementation at least as high as the rest of the repo average.

thisthat:
## This PR

- Refactor the code to make it easier to inject the K8s clients
- Add unit tests 

Fixes #460

### DIFF
--- a/pkg/sync/kubernetes/kubernetes_sync.go
+++ b/pkg/sync/kubernetes/kubernetes_sync.go
@@ -23,20 +23,22 @@ import (
 )
 
 var (
-	resyncPeriod = 1 * time.Minute
-	apiVersion   = fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version)
+	resyncPeriod                    = 1 * time.Minute
+	apiVersion                      = fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version)
+	featurFlagConfigurationResource = v1alpha1.GroupVersion.WithResource("featureflagconfigurations")
 )
 
 type Sync struct {
-	Logger       *logger.Logger
-	ProviderArgs sync.ProviderArgs
-	URI          string
+	Logger        *logger.Logger
+	ProviderArgs  sync.ProviderArgs
+	URI           string
+	ReadClient    client.Reader
+	DynamicClient dynamic.Interface
 
-	ready      bool
-	namespace  string
-	crdName    string
-	readClient client.Reader
-	informer   cache.SharedInformer
+	ready     bool
+	namespace string
+	crdName   string
+	informer  cache.SharedInformer
 }
 
 func (k *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error {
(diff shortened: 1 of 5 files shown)


pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -23,20 +23,22 @@ import (
 )
 
 var (
-	resyncPeriod = 1 * time.Minute
-	apiVersion   = fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version)
+	resyncPeriod                    = 1 * time.Minute
+	apiVersion                      = fmt.Sprintf("%s/%s", v1alpha1.GroupVersion.Group, v1alpha1.GroupVersion.Version)
+	featurFlagConfigurationResource = v1alpha1.GroupVersion.WithResource("featureflagconfigurations")
```
skyerus, line 28:
```suggestion
	featureFlagConfigurationResource = v1alpha1.GroupVersion.WithResource("featureflagconfigurations")
```

pkg/sync/kubernetes/kubernetes_sync_test.go
```diff
@@ -556,6 +559,281 @@ func TestSync_watcher(t *testing.T) {
...
+	fakeDynamicClient := fake.NewSimpleDynamicClient(s, ff)
+	validFFCfg := &v1alpha1.FeatureFlagConfiguration{
+		TypeMeta: Metadata,
+		ObjectMeta: v1.ObjectMeta{
+			Name:      name,
+			Namespace: ns,
+		},
+	}
+	fakeReadClient := newFakeReadClient(validFFCfg)
+	l, _ := logger.NewZapLogger(zapcore.DebugLevel, "")
```
skyerus, line 613: Can we error check here?

pkg/sync/kubernetes/kubernetes_sync_test.go
```diff
@@ -556,6 +559,281 @@ func TestSync_watcher(t *testing.T) {
...
+
+	t.Run("Happy Path", func(t *testing.T) {
```
skyerus, line 615: Could [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) be used here to reduce duplication/improve readability?

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -23,20 +23,22 @@ import (
...
-
-	ready      bool
-	namespace  string
-	crdName    string
-	readClient client.Reader
-	informer   cache.SharedInformer
+	Logger        *logger.Logger
+	ProviderArgs  sync.ProviderArgs
+	URI           string
+	ReadClient    client.Reader
```
skyerus, line 35: What's the reason for exporting this field?

pkg/sync/kubernetes/kubernetes_sync_test.go
thisthat: I think we can do that 👍

pkg/sync/kubernetes/kubernetes_sync.go
thisthat: I think refactoring the code a bit and introducing a `NewK8sSync(read, dynamic)` function we don't need to expose the two clients. At the moment they are exposed to construct the object in the runtime package

pkg/sync/kubernetes/kubernetes_sync.go
```diff
@@ -325,3 +311,37 @@ func k8sClusterConfig() (*rest.Config, error) {
 
 	return clusterConfig, nil
 }
+
+func NewK8sSync(
```
Kavindu-Dodan, line 315: Minor (personal preference) - I would like to see this on the top of the file so it is obvious that we have a New function to construct the structure.

Kavindu-Dodan (approved):
Approving. Left a minor suggestion.

I like the `New` function based struct construction. In furure, when all Sync providers have a similar construction, I think we should get rid of `Init()` function. Because, I would expect `New` function to construct an initialized Sync provider :)

james-milligan (approved):

toddbaert (approved):
`Merging... will increase coverage by 6.21%.` :sunglasses:

toddbaert (approved):

PR #477, opened 2023-03-08
feat: grpc tls connectivity (grpcs)

Kavindu-Dodan:
## This PR

Introduce TLS connectivity for GRPC sync provider.

TLS can be enabled using schema `grpcs://`. For example, 

`./flagd start --uri grpcs://localhost:8090`

Further, a self-sign certificate can be provided for TLS connectivity using configuration source field `certPath`

ex:- `./flagd start --sources='[{"uri":"grpcs://localhost:9090","provider":"grpc", "certPath":"<CA_CERT>"}]'`

### How to test 

Start mock server impl - https://github.com/Kavindu-Dodan/flagd-grpc-sync-server & then run flagd with grpc tls mode


beeme1mr:
This can be tested using the test server @Kavindu-Dodan created and generating self signed certs.

Generate CA cert

- CA Private Key: `openssl ecparam -name prime256v1 -genkey -noout -out ca.key`
- CA Certificate: `openssl req -new -x509 -sha256 -key ca.key -out ca.cert`

Generate Server certificate

- Create OpenSSL config: `echo "subjectAltName = DNS:localhost" > openssl.conf`
- Server private key:  `openssl ecparam -name prime256v1 -genkey -noout -out server.key`
- Server signing request:  `openssl req -new -sha256 -addext "subjectAltName=DNS:localhost" -key server.key -out server.csr`
- Server cert:  `openssl x509 -req -in server.csr -CA ca.cert -CAkey ca.key  -out server.crt -days 1000 -sha256 -extfile openssl.conf`

toddbaert (changes requested):
Largely this is working, but I _believe_ it caused a regression in our reconnect which causes flagd to crash when the server connection is lost. This tells me we need a test, e2e or otherwise, for this... I will create a new issue for that. I think the crash itself needs to be resolved in this PR though.

Additionally, `grpcs://` is not _really_ a valid scheme. I think it's fine to keep for the `URL` style configuration, but I think a boolean for `tls` might make more sense for the `--sources` style configuration. This more a matter of opinion though... I just feel wierd about relying on a non-standard scheme wherever we don't have to.

Kavindu-Dodan:
> Largely this is working, but I _believe_ it caused a regression in our reconnect which causes flagd to crash when the server connection is lost. This tells me we need a test, e2e or otherwise, for this... I will create a new issue for that. I think the crash itself needs to be resolved in this PR though.
> 
> Additionally, `grpcs://` is not _really_ a valid scheme. I think it's fine to keep for the `uri` style configuration, but I think a boolean for `tls` might make more sense for the `--sources` style configuration. This more a matter of opinion though... I just feel weird about relying on a non-standard scheme wherever we don't have to.
> 
> Edit: I created [this issue](https://github.com/open-feature/flagd/issues/483) for testing.

Good catch. I fixed this and added a unit test to validate this scenario 

I also have no hard opinion on the `grpcs://` schema usage. But if there are no other alternatives, this is fine :)

pkg/sync/grpc/grpc_sync.go
```diff
@@ -70,30 +83,28 @@ func (g *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
...
+	syncClient, err := g.client.SyncFlags(ctx, &v1.SyncFlagsRequest{ProviderId: g.ProviderID})
+	if err != nil {
+		return err
+	}
+
+	// Initial stream listening. Error will be logged and continue and retry connection establishment
+	err = g.handleFlagSync(syncClient, dataSync)
 	if err == nil {
-		return nil
+		// This should not happen as handleFlagSync expects to return with an error
```
toddbaert, line 106: "This should not happen as handleFlagSync expects to return with an error"

This _would_ happen if there's no problem on initial connection though, right? In that case we'd expect this?

pkg/sync/grpc/grpc_sync_test.go
```diff
@@ -615,6 +617,94 @@ func Test_ConnectWithRetry(t *testing.T) {
...
+	case <-tCtx.Done():
+		cancelFunc()
+		t.Error("timeout waiting for conditions to fulfil")
+	case rsp := <-syncChan:
+		if rsp.Type != expectType {
+			t.Errorf("expected response: %s, but got: %s", expectType, rsp.Type)
+		}
+	}
+}
+
```
toddbaert, line 707: Very cool, thanks.

toddbaert (approved):

toddbaert:
Confirmed reconnect works now.

pkg/sync/grpc/grpc_sync.go
Kavindu-Dodan: `handleFlagSync` is a blocking call. So if it returns, ideally there is an error. However, by any chance we return with a nil error, this check makes sure we do not panic by accessing error content.

toddbaert: Ahh yes, of course! I forgot it was blocking!

PR #495, opened 2023-03-13
feat: WIP Introduce flagd kube proxy

james-milligan:
## This PR

- introduced flagd Kubernetes proxy as new binary


kube-flagd-proxy/main.go
```diff
@@ -1,17 +1,31 @@
...
+	commit     = "HEAD"
+	date       = "unknown"
+	bannerText = `
+	{{ .AnsiColor.BrightRed }}	 ______   __       ________   _______    ______      
+	{{ .AnsiColor.BrightRed }}	/_____/\ /_/\     /_______/\ /______/\  /_____/\     
+	{{ .AnsiColor.BrightRed }}	\::::_\/_\:\ \    \::: _  \ \\::::__\/__\:::_ \ \    
+	{{ .AnsiColor.BrightRed }}	 \:\/___/\\:\ \    \::(_)  \ \\:\ /____/\\:\  \ \ \   
+	{{ .AnsiColor.BrightRed }}	  \:::._\/ \:\ \____\:: __  \ \\:\\_  _\/ \:\ \ \ \  
+	{{ .AnsiColor.BrightRed }}	   \:\ \    \:\/___/\\:.\ \  \ \\:\_\ \ \  \:\/.:| | 
+	{{ .AnsiColor.BrightRed }}	    \_\/     \_____\/ \__\/\__\/ \_____\/   \____/_/
```
beeme1mr, line 22:
```suggestion
	{{ .AnsiColor.BrightRed }}	    \_\/     \_____\/ \__\/\__\/ \_____\/   \____/_/ 
	{{ .AnsiColor.BrightRed }}	                                   Kubernetes Proxy  
```

kube-flagd-proxy/cmd/start.go
```diff
@@ -0,0 +1,83 @@
...
+	logFormatFlagName   = "log-format"
+	metricsPortFlagName = "metrics-port"
+	portFlagName        = "port"
+)
+
+func init() {
+	flags := startCmd.Flags()
+
+	// allows environment variables to use _ instead of -
+	flags.Int32P(portFlagName, "p", 8013, "Port to listen on")
```
beeme1mr, line 32: Will the proxy be able to do flag evaluation?

james-milligan: At least in this version no, it should only act as a simple proxy for reading configuration changes from FeatureFlagConfigurations without the required permissions

beeme1mr: Ah, I see. This is the port used for proxying the requests. Got it, thanks!

core/pkg/service/flag-evaluation/connect_metrics.go
skyerus, line 1: What's the reason for moving these?

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,235 @@
...
+		mu:           &sync.Mutex{},
+		SyncBuilder:  &SyncBuilder{},
+	}
+}
+
+func (s *SyncStore) FetchAllFlags(ctx context.Context, key interface{}, target string) (isync.DataSync, error) {
+	dataSyncChan := make(chan isync.DataSync, 1)
+	errChan := make(chan error, 1)
+	s.mu.Lock()
+	_, ok := s.syncHandlers[target]
```
skyerus, line 62: I'd assign the syncHandler to a var so the Unlock can be called directly after. Also this can be RLock/RUnlock unless I've misunderstood

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,235 @@
...
+			dataSync: make(chan isync.DataSync),
+			subs: map[interface{}]storedChannels{
+				key: {
+					errChan:  errChan,
+					dataSync: dataSync,
+				},
+			},
+		}
+		s.syncHandlers[target] = &syncHandler
+		fmt.Println("here", s.syncHandlers[target])
```
skyerus: Remove

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,235 @@
...
+	case <-time.After(5 * time.Second):
+		return isync.DataSync{}, errors.New("fetching all flags timed out after 5 seconds")
+	}
+}
+
+func (s *SyncStore) RegisterSubscription(ctx context.Context, target string, key interface{}, dataSync chan isync.DataSync, errChan chan error) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	// is there a currently active subscription for this target?
+	_, ok := s.syncHandlers[target]
```
skyerus, line 85: assign to var?

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,235 @@
...
+}
+
+func (s *SyncStore) watchResource(ctx context.Context, target string) {
+	ctx, cancel := context.WithCancel(ctx)
+	s.syncHandlers[target].cancelFunc = cancel
+	defer cancel()
+
+	go func() {
+		<-ctx.Done()
+		fmt.Println("deleting")
```
skyerus: remove

skyerus:
Some of the locking can be improved by using read only locks. Looks promising 😄

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/open-feature/flagd/core/pkg/logger"
+	isync "github.com/open-feature/flagd/core/pkg/sync"
+	"github.com/open-feature/flagd/core/pkg/sync/file"
+	"github.com/open-feature/flagd/core/pkg/sync/kubernetes"
+	"go.uber.org/zap"
+)
+
+var (
+	regCrd  *regexp.Regexp
+	regFile *regexp.Regexp
```
Kavindu-Dodan, line 20: Do we support file syncs ? 🤔 

If so we need to rethink file read overhead. K8s syncs are backed by informer caches. This is not the case with files.

core/pkg/service/sync/sync.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+type SyncServerConfiguration struct {
+	Port        uint16
+	MetricsPort uint16
+}
+
+func (s *SyncServer) Serve(ctx context.Context, svcConf iservice.Configuration) error {
+	mux := http.NewServeMux()
+	path, handler := rpc.NewFlagSyncServiceHandler(s)
```
Kavindu-Dodan, line 36: This is duplicated inside `setupServer()`

core/pkg/sync-store/sync_store.go
james-milligan: we do not, I've included this as a feature for easy debugging and won't be in the final product

kube-flagd-proxy/cmd/start.go
```diff
@@ -0,0 +1,83 @@
...
+			Configuration: syncServer.SyncServerConfiguration{
+				Port:        viper.GetUint16(portFlagName),
+				MetricsPort: viper.GetUint16(metricsPortFlagName),
+			},
+			Logger: logger,
+		}
+
+		go store.Cleanup()
+		go s.Serve(ctx, service.Configuration{
+			ReadinessProbe: func() bool { return true },
```
Kavindu-Dodan, line 76: I prefer to have both readiness and liveness, even if they do not correlate to any internal real state (always `true`). This allows us to wire them later when we really see the need.

core/pkg/service/sync/sync.go
```diff
@@ -0,0 +1,139 @@
+package sync
```
Kavindu-Dodan, line 1: I see several components here - a core similar to a `runtime`, `subscription handler` & `grpc/connect listener`. A refactoring here with good contracts will help with testing and maintenance. 

For example, `rpc.NewFlagSyncServiceHandler(s)`, parsing `SyncServer` as `s` is a little misleading.

kube-flagd-proxy/cmd/root.go
```diff
@@ -0,0 +1,76 @@
+package cmd
```
Kavindu-Dodan, line 1: Needs cleanup - several uses of `flagd` instead of `<NAME>`. Yes we need a NAME for this component. 

@toddbaert  & @beeme1mr  any thoughts?

core/pkg/sync-store/sync_store.go
Kavindu-Dodan: Nice, thanks :)

Kavindu-Dodan:
Overal looks good. Hope to see the refined version when it's ready 👏

core/pkg/service/sync/sync.go
james-milligan: removed

core/pkg/service/sync/sync.go
james-milligan: > For example, rpc.NewFlagSyncServiceHandler(s), parsing SyncServer as s is a little misleading.

This implementation is copied from `runtime`, happy to create an issue to refactor both, as they should follow a similar pattern

james-milligan: Happy to refactor this, could you elaborate on the structure you are suggesting?

kube-flagd-proxy/cmd/root.go
james-milligan: Yeah this was a direct copy and paste from the Flagg cmd, I agree this needs fixing, IMO im happy to call it `Kubernetes-flagd-proxy` here to match the other references

core/pkg/sync-store/sync_store.go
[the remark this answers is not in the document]
james-milligan: we already have a store package, I could introduce this into the same directory and next the existing one with another name?

kube-flagd-proxy/main.go
beeme1mr: Do you not like my suggestion? 😄

kube-flagd-proxy/cmd/root.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+var rootCmd = &cobra.Command{
+	Use:               "flagd",
+	Short:             "Flagd-kube-proxy allows flagd to subscribe to CRD changes without the required permissions.",
+	Long:              ``,
+	DisableAutoGenTag: true,
+	// Uncomment the following line if your bare application
+	// has an action associated with it:
+	// Run: func(cmd *cobra.Command, args []string) { },
```
Kavindu-Dodan, line 27: These can be removed and kept clean

core/pkg/service/sync/server.go
```diff
@@ -0,0 +1,106 @@
...
+			errChan <- err
+		}
+	}()
+
+	select {
+	case err := <-errChan:
+		return err
+	case <-ctx.Done():
+		return s.server.Shutdown(ctx)
+	}
```
Kavindu-Dodan, line 60: Potentially a refactoring suggestion - can be a follow-up 

Here, we start two servers (listeners) 1. Metrics 2. GRPC streams

Both can be wrapped in errgroup to handle errors and avoid the error chan. Further,  setupServer() is not needed in top level if we do this . Psecudo code look like,

```
group, gCtx := errgroup.WithContext(ctx)

group.go(startGrpcServer)

group.go(startMetricsServer)

group.wait()
```

This makes it easy to read the code as well as test inidividual server impls :)

james-milligan: nice suggestion, ill implement this tomorrow

kube-flagd-proxy/cmd/root.go
james-milligan: sure, will remove these lines from the flags cmd as well

core/pkg/service/sync/handler.go
```diff
@@ -0,0 +1,55 @@
...
+	dataSync := make(chan sync.DataSync)
+	l.syncStore.RegisterSubscription(ctx, req.Msg.GetProviderId(), req, dataSync, errChan)
+	for {
+		select {
+		case e := <-errChan:
+			return e
+		case d := <-dataSync:
+			if err := stream.Send(&syncv1.SyncFlagsResponse{
+				FlagConfiguration: d.FlagData,
+				State:             syncv1.SyncState(d.Type + 1),
```
Kavindu-Dodan, line 47: Needs to handle the sync type here. Best to write a simple helper `syncTypeToGrpcState(d.Type) SyncState`

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,277 @@
...
+		s.mu.RUnlock()
+		return
+	}
+	// sync ref is used to trigger a resync on a single channel when a new subscription is started
+	// but the associated SyncHandler already exists, i.e. this function is not run
+	sh.syncRef = sync
+	err = sync.Sync(ctx, sh.dataSync)
+	if err != nil {
+		s.logger.Error(fmt.Sprintf("error from sync for target %s: %s", target, err.Error()))
+		s.mu.RLock()
```
Kavindu-Dodan, line 214: `s.mu` is a store-wide mutex. Given that we maintain a single store, when we lock, it applies to all subroutines 🤔

This needs to be improved and cleaned up 

We need mutex for Store and then individual locks for subscription handlers

james-milligan: done

core/pkg/service/sync/handler.go
james-milligan: done

core/pkg/service/sync/server.go
james-milligan: done

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,279 @@
...
+		sh.mu.RUnlock()
+		return
+	}
+	// init sync, if this fails an error is broadcasted, and the defer results in cleanup
+	err = sync.Init(ctx)
+	if err != nil {
+		s.logger.Error(fmt.Sprintf("unable to initiate sync for target %s: %s", target, err.Error()))
+		sh.mu.RLock()
+		for _, ec := range sh.subs {
+			ec.errChan <- err
```
skyerus, line 205: Is there any worry of this blocking?

skyerus: There is a few instances of this logic, could be worth extracting into a func?

skyerus (approved):
LGTM, nice work

core/pkg/service/sync/handler.go
```diff
@@ -0,0 +1,59 @@
...
+type handler struct {
+	syncStore *syncStore.SyncStore
+	logger    *logger.Logger
+}
+
+func (l *handler) FetchAllFlags(ctx context.Context, req *connect.Request[syncv1.FetchAllFlagsRequest]) (
+	*connect.Response[syncv1.FetchAllFlagsResponse],
+	error,
+) {
+	data, err := l.syncStore.FetchAllFlags(ctx, nil, req.Msg.GetProviderId())
```
Kavindu-Dodan, line 22: Setting key to `nil`  here can result in a panic.

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,287 @@
...
+		return isync.DataSync{}, err
+	case <-time.After(5 * time.Second):
+		return isync.DataSync{}, errors.New("fetching all flags timed out after 5 seconds")
+	}
+}
+
+func (s *SyncStore) RegisterSubscription(
+	ctx context.Context,
+	target string,
+	key interface{},
```
Kavindu-Dodan, line 96: Do we need this `key` 🤔  I still couldn't figure out why we need it. May be I am missing the usage

core/pkg/sync-store/sync_store.go
```diff
@@ -0,0 +1,287 @@
...
+	sh.syncRef = sync
+	err = sync.Sync(ctx, sh.dataSync)
+	if err != nil {
+		s.logger.Error(fmt.Sprintf("error from sync for target %s: %s", target, err.Error()))
+		sh.writeError(s.logger, err)
+	}
+}
+
+func (h *syncHandler) writeError(logger *logger.Logger, err error) {
+	h.mu.RLock()
```
Kavindu-Dodan, line 207: Minor - A deferred unlock is better as it guarantees the unlock :) 

```suggestion
	h.mu.RLock()
	defer h.mu.RUnlock()
```

core/pkg/service/sync/handler.go
Kavindu-Dodan, line 22: And let's utilize `filter` field instead of `providerId` [1]. This makes the implementation generic :) 

[1] - https://github.com/open-feature/schemas/pull/89

Kavindu-Dodan (commented):
Looks good. Let's address the hilighted points and once done, this is good to go 👏

core/pkg/sync-store/sync_store.go
james-milligan: its a unique identifier for the subscription channel that we can use without allocating additional memory, its just in the subscription map -> it also gives us a clean trace ID if we choose to go that way

kube-flagd-proxy/main.go
```diff
@@ -1,17 +1,32 @@
...
 
-func main() {
-	l, err := logger.NewZapLogger(zapcore.DebugLevel, "console")
-	if err != nil {
-		log.Fatalf("initialize zap logger: %v", err)
-	}
+var (
+	version    = "dev"
+	commit     = "HEAD"
+	date       = "unknown"
```
beeme1mr, line 14: Should these values be hardcoded?

james-milligan: yes, they are overwritten on build there are the default values - mirrors the build command used in flagd

https://github.com/open-feature/flagd/blob/2f8e7efc69474bfd88e30ef66acfce7cdf23ccc3/flagd/main.go#L12

kube-flagd-proxy/cmd/start.go
```diff
@@ -0,0 +1,85 @@
...
+	metricsPortFlagName = "metrics-port"
+	portFlagName        = "port"
+)
+
+func init() {
+	flags := startCmd.Flags()
+
+	// allows environment variables to use _ instead of -
+	flags.Int32P(portFlagName, "p", 8013, "Port to listen on")
+	flags.Int32P(metricsPortFlagName, "m", 8014, "Metrics port to listen on")
```
beeme1mr, line 34: Should the proxy use different ports?

Kavindu-Dodan: Good point. Yes, this makes sense for example when running locally or for e2e tests

james-milligan: updated

kube-flagd-proxy/README.md
```diff
@@ -0,0 +1,60 @@
...
+<!-- markdownlint-disable MD033 -->
+<p align="center">
+    <img src="../images/kube-flagd-proxy.png" width="650">
+</p>
+<!-- markdownlint-enable MD033 -->
+On request, the flagd-kube-proxy will spawn a goroutine to watch the CR using the `core` package kubernetes sync. Each further request for the same resource will add a new new stream to the broadcast list. Once all streams have been closed, and there are no longer any listeners for a given resource, the sync will be closed.  
+
+The kube-flagd-proxy API follows the flagd grpc spec, found in the [buf schema registry](https://buf.build/open-feature/flagd), as such the existing grpc sync can be used to subscribe to the the CR changes.
+
+## Deplyoment
```
beeme1mr, line 15:
```suggestion
## Deployment
```

Kavindu-Dodan (approved):

kube-flagd-proxy/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: flagd
+spec:
+  containers:
+  - name: flagd
+    image: ghcr.io/open-feature/flagd:v0.4.5
```
beeme1mr, line 34: Should we have release please update this version?

beeme1mr (approved):

Issue #501, opened 2023-03-14
Expose "impressions" as OTel Metric


toddbaert:
Impressions (basically evaluations) are a key metric for feature flag admins. We should capture this count by flag key, and [variant](https://docs.openfeature.dev/docs/specification/glossary#variant) (if reasonable).

agardnerIT:
Can you provide some hints on how a contributor would begin tackling this or even pointing to the files on where to start?

thisthat:
Hey @agardnerIT, I created a small PoC to showcase how these metrics can be exposed: https://github.com/thisthat/flagd/pull/1

beeme1mr:
That's great @thisthat! :shipit:

PR #419, opened 2023-02-19
feat!: Use OTel to export Prometheus Metrics

thisthat:
### Related Issues

In #186, there was a discussion about going with OpenTelemetry or a direct Prometheus implementation. 
The agreement was to have a scraping endpoint, i.e., `pull mode`,   and not support the `push` mode.
This PR wants to replace the existing direct Prometheus implementation with the vendor-agnostic OpenTelemetry one, maintaining the same feature set. 

<details>
  <summary>Example of exposed metrics:</summary>

```
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 5.1e-05
go_gc_duration_seconds{quantile="0.25"} 0.000174874
go_gc_duration_seconds{quantile="0.5"} 0.000559251
go_gc_duration_seconds{quantile="0.75"} 0.000875708
go_gc_duration_seconds{quantile="1"} 0.001527791
go_gc_duration_seconds_sum 0.010100666
go_gc_duration_seconds_count 18
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 21
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.19.3"} 1
# HELP go_memstats_alloc_bytes Number of bytes allocated and still in use.
# TYPE go_memstats_alloc_bytes gauge
go_memstats_alloc_bytes 2.6836392e+07
# HELP go_memstats_alloc_bytes_total Total number of bytes allocated, even if freed.
# TYPE go_memstats_alloc_bytes_total counter
go_memstats_alloc_bytes_total 2.76925672e+08
# HELP go_memstats_buck_hash_sys_bytes Number of bytes used by the profiling bucket hash table.
# TYPE go_memstats_buck_hash_sys_bytes gauge
go_memstats_buck_hash_sys_bytes 6351
# HELP go_memstats_frees_total Total number of frees.
# TYPE go_memstats_frees_total counter
go_memstats_frees_total 2.563256e+06
# HELP go_memstats_gc_sys_bytes Number of bytes used for garbage collection system metadata.
# TYPE go_memstats_gc_sys_bytes gauge
go_memstats_gc_sys_bytes 7.975824e+06
# HELP go_memstats_heap_alloc_bytes Number of heap bytes allocated and still in use.
# TYPE go_memstats_heap_alloc_bytes gauge
[...]

PR #530, opened 2023-03-17
feat: refactor core module into multiple packages

thisthat:
Signed-off-by: Giovanni Liva <giovanni.liva@dynatrace.com>

## This PR

- Refactor the service_[connect|metrics].go into multiple files 
- Adds more unit tests when easy and feasible 

### Related Issues

Preparation for #501


toddbaert:
I like this refactor. In general it's my preference to see refactors done in separate PRs. It's helps to be able to approach a refactor PR knowing there's no functional changes. I'd like to see it separated if possible, but it's not necessary.

thisthat:
> I like this refactor. In general it's my preference to see refactors done in separate PRs. It's helps to be able to approach a refactor PR knowing there's no functional changes. I'd like to see it separated if possible, but it's not necessary.

That's a fair argument. I will remove the impressions and just tackle the refactoring with this PR 👍

core/pkg/otel/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+func (r MetricsRecorder) InFlightRequestEnd(ctx context.Context, attrs []attribute.KeyValue) {
+	r.httpRequestsInflight.Add(ctx, -1, attrs...)
+}
+
+func (r MetricsRecorder) Impressions(ctx context.Context, key, variant string) {
+	r.impressions.Add(ctx, 1, []attribute.KeyValue{
+		semconv.FeatureFlagKey(key),
+		semconv.FeatureFlagVariant(variant),
+		semconv.FeatureFlagProviderName("flagd"),
+	}...)
```
toddbaert, line 54: Seems right!

core/pkg/otel/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func NewOTelRecorder(exporter metric.Reader, serviceName string) *MetricsRecorder {
+	const requestDurationName = "http_request_duration_seconds"
+	const responseSizeName = "http_response_size_bytes"
+
+	// create a metric provider with custom bucket size for histograms
+	provider := metric.NewMeterProvider(
+		metric.WithReader(exporter),
+		metric.WithView(getDurationView(requestDurationName, serviceName, prometheus.DefBuckets)),
```
toddbaert, line 79: Do we have to use prometheus APIs instead of OTel for these metrics?

thisthat: This is for just keeping the same bucket size as it was before moving to OTel as SDK to report metrics (see https://github.com/open-feature/flagd/pull/419#issuecomment-1440806008)

core/pkg/otel/metrics.go
```diff
@@ -0,0 +1,107 @@
...
+		metric.WithView(getDurationView(responseSizeName, serviceName, prometheus.ExponentialBuckets(100, 10, 8))),
```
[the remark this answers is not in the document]
thisthat: I don't think a constant make sense here, but definitely a comment helps the reader to understand why we change the bucket size :)

core/pkg/otel/metrics.go
thisthat: As discussed before, I'll remove this and tackle the introduction of the new metric in a separate PR - this way the changelog is also nicer :)

core/pkg/otel/metrics.go
thisthat: Maybe to expand more on this subject: using OTel allows us to be vendor agnostic and provide metrics in different formats. The "output format" currently is wired inside the [`from_config` file](https://github.com/open-feature/flagd/pull/530/files#diff-60ffee6ef9753f574677d132f3e595feee457c5e52c1b1b88063c616a3c6cdf7R57)

core/pkg/otel/metrics_test.go
```diff
@@ -0,0 +1,168 @@
...
+				semconv.HTTPMethodKey.String(""),
+				semconv.HTTPStatusCodeKey.String(""),
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			rec := MetricsRecorder{}
+			res := rec.HTTPAttributes(tt.req.Service, tt.req.ID, tt.req.Method, tt.req.Code)
+			if len(res) != 4 {
```
bacherfl, line 78:
```suggestion
require.Equal(t, tt.want, res)
```

thisthat:
I need some help with the integration-tests @toddbaert 
Locally, they also don't work. The strange thing is that I can reach the binary under test via curl `curl -X POST "localhost:8013/schema.v1.Service/ResolveString" -d '{"flagKey":"string-flag","context":{}}' -H "Content-Type: application/json"`

thisthat:
> I need some help with the integration-tests @toddbaert Locally, they also don't work. The strange thing is that I can reach the binary under test via curl `curl -X POST "localhost:8013/schema.v1.Service/ResolveString" -d '{"flagKey":"string-flag","context":{}}' -H "Content-Type: application/json"`

For the record, connect-go requires an implementation of `Flush()` to work properly

bacherfl (approved):
lgtm

toddbaert (approved):

PR #536, opened 2023-03-20
chore: move credential builder for grpc sync into seperate component

bacherfl:
This PR moves the credential builder of the grpc sync into its own component to keep both components small and make unit testing easier. Also added a unit test for the `Init` method of the grpc sync component


james-milligan (approved):
LGTM

toddbaert (approved):
solid refactor

PR #538, opened 2023-03-20
fix: benchmark pipeline

### LINKED ISSUE #496: [BUG] Benchmarks no longer run after refactor
### Observed behavior

The nightly benchmark GitHub action no longer succeeds after the refactor.

Here's the step that fails:

```
Run set -o pipefail; go test -bench=Bench -short -benchtime=5s -benchmem ./... | tee output.txt
pattern ./...: directory prefix . does not contain main module or its selected dependencies
Error: Process completed with exit code 1.
```
And a link to a failed run:

https://github.com/open-feature/flagd/actions/runs/4400751617/jobs/7706332843

### Expected Behavior

The benchmarks should run successfully.

### Steps to reproduce

_No response_

odubajDT:
## This PR

- fixes benchmark pipeline
- ran `go mod tidy` on all directories

### Related Issues

Fixes #496


beeme1mr (approved):

flagd/go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/cucumber/godog v0.12.6
 	github.com/dimiro1/banner v1.1.0
 	github.com/mattn/go-colorable v0.1.13
+	github.com/open-feature/flagd/core v0.4.4
```
thisthat, line 9: This should not be here since we're using workspaces. 
Maybe we should not having the `go.work` files in gitignore 🤔 what's your opinion @james-milligan ?

kube-flagd-proxy/go.mod
```diff
@@ -5,32 +5,94 @@ go 1.19
 require (
 	github.com/dimiro1/banner v1.1.0
 	github.com/mattn/go-colorable v0.1.13
+	github.com/open-feature/flagd/core v0.4.5
```
thisthat, line 8: ditto

flagd/go.mod
james-milligan: I would still have expected the `go.mod` to contain `flagd/core`, all the go workspace is doing is handling `replace directives` (`=>`) under the covers

If the project is opened at a single module level, i.e. in `./flagd` there should be no issues / errors.

odubajDT: @thisthat

toddbaert (approved):

PR #550, opened 2023-03-22
fix: fixing silent lint failures

Kavindu-Dodan:
## This PR

Fix lint check & adds lint fixes for existing lint failures (reason for 16 file changes)

See this example [1], linting was failing silently as we rely on a `foreach` loop. 

Solution credits - https://stackoverflow.com/a/22214424/3197055  

[1] - https://github.com/open-feature/flagd/actions/runs/4471193065/jobs/7855746534#step:5:640


thisthat (approved):
Nice catch 🙌

james-milligan (approved):

toddbaert (approved):

PR #554, opened 2023-03-23
chore: refactor middleware setup in server

bacherfl:
This PR aims to clean up the way the middlewares for the server are set up.
Previously, the Metrics, CORS and H2C Middlewares were each set up in different ways, which made that section of the code a bit hard to comprehend at first.

Therefore, i introduced a `middleware` package, including sub packages for the metrics, cors, and h2c middleware. Each of those middlewares implement an `IMiddleware` interface, that can be registered to the server using the newly introduced `AddMiddleware` function


core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/open-feature/flagd/core/pkg/service/middleware"
```
thisthat, line 8: This will make the Linter fail with https://github.com/open-feature/flagd/pull/550

core/pkg/service/flag-evaluation/connect_service_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	middlewaremock "github.com/open-feature/flagd/core/pkg/service/middleware/mock"
```
thisthat, line 7: similar linting problem here

thisthat (approved):

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -94,30 +95,39 @@ func (s *ConnectService) setupServer(svcConf service.Configuration) (net.Listene
...
+		ReadHeaderTimeout: time.Second,
+		Handler:           handler,
+	}
+
+	go bindMetrics(s, svcConf)
+
+	// Add middlewares
+
+	metricsMiddleware := metricsmw.NewHTTPMetric(metricsmw.Config{
 		Service:        "openfeature/flagd",
```
Kavindu-Dodan, line 108: Since you are working on this : Service name `"openfeature/flagd"` is exposed through `service.Configuration`. We can improve it here :)

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -94,30 +95,39 @@ func (s *ConnectService) setupServer(svcConf service.Configuration) (net.Listene
 	path, handler := schemaConnectV1.NewServiceHandler(fes)
 	mux.Handle(path, handler)
 
-	mdlw := middleware.NewHTTPMetric(middleware.Config{
+	s.server = http.Server{
+		ReadHeaderTimeout: time.Second,
+		Handler:           handler,
+	}
+
+	go bindMetrics(s, svcConf)
```
Kavindu-Dodan, line 103: Since you are working on this (I am 100% agreeing to the refactoring)-

IMO this must be moved from `setupServer()` and be part of `serve()` - Because I expect services to be up and running when we actually serve :) 

See this implementation [1] for kube-proxy part which IMO is easy to understand 

And going further, we can avoid passing `ConnectService` to `bindMetrics()` as it's only usage is logging. I think we can pass a fresh logger for metrics which is even an improvement :) 

Do you think you can work on this too? Otherwise it could be a follow-up 

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/sync/server.go#L44-L45

Kavindu-Dodan:
Awesome work - this IMO is a long overdue refactoring to clean up.

I left two improvements of which one is essential and the other is a suggestion.

core/pkg/service/flag-evaluation/connect_service.go
thisthat: I agree with the first part @Kavindu-Dodan, great suggestion 👍  
With the suggestion for `bindMetrics()`, this function uses only the port field of the service configuration, not the logger. The function provides a binding for opening a fresh HTTP server for probes and the Prometheus metric page

bacherfl: Good suggestion, I have moved the `bindMetrics` into the `Serve` method, and renamed it to `startMetricsServer` to be consistent with the `sync` server's naming. Another improvement would be to also move the logic of the start of the server handling the business logic into its own method, but I would do this in a small follow up PR after this one is merged :)

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -65,6 +66,8 @@ func (s *ConnectService) Serve(ctx context.Context, eval eval.IEvaluator, svcCon
 		close(errChan)
 	}()
 
+	go s.startMetricsServer(svcConf)
```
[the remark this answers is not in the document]
Kavindu-Dodan: yes and I think @bacherfl already working on this through a dedicated PR [1]

And using errgroup will help here :)

core/pkg/service/flag-evaluation/connect_service.go
Kavindu-Dodan: Thank you :) look forward for the next PR

beeme1mr (approved):

PR #558, opened 2023-03-24
feat: Introduce kube-proxy-metrics

james-milligan:
## This PR

- adds framework for exposing metrics for the `flagd-kube-proxy`
- exposes `sync_active_streams`, for use in auto scaling the proxy in deployments


core/pkg/service/sync/sync_metrics.go
```diff
@@ -0,0 +1,35 @@
+package sync
+
+import (
+	"context"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
```
thisthat, line 7: can we use OTel for exposing metrics and configure prometheus as the default exporter? this way we can change the backend implementation more easily if necessary.

core/pkg/service/sync/sync_metrics.go
```diff
@@ -0,0 +1,42 @@
...
+
+	_, err = meter.RegisterCallback(func(_ context.Context, o api.Observer) error {
+		o.ObserveInt64(syncGuage, s.handler.syncStore.GetActiveSubscriptionsInt64())
+		return nil
+	}, syncGuage)
+	if err != nil {
+		return err
+	}
+
+	<-ctx.Done()
```
skyerus, line 40: Why is this needed?

skyerus (approved):

PR #560, opened 2023-03-24
chore!: unify sources configuration handling

Kavindu-Dodan:
## This PR

This PR is best reviewed by going through the new documentation [1]

Restructured flagd startup configuration handling with respect to sync providers. Documentation is updated [1] & handling of prefixes is cleared (ex:- removal of grpc(s):// prefix for `--source` based sync provider startup argument)

## Breaking change

Contains a breaking change with respect to how sync provider URIs are handled

- **what** -source configuration URI parsing
- **why** - align source configuration handling among different sources & avoid mix uage of prefixes to aling with documented behavior
- **how** - source configuration no longer validate or parse URI prefixes. This means, existing source configuration strings/files may need an upgrade if they used prefixes in URIs. The upgrade is to simply remove any prefixes from source configuration URIs 

For example, 

**old**

``` yaml
sources:
- uri: grpcs://my-flag-source:8080 # mixed usage of schema shared with URI prefixes ❌ 
  provider: grpc
  certPath: /certs/ca.cert
```

**new**

``` yaml
sources:
- uri: my-flag-source:8080
  provider: grpc
  tls: true. # new way to convey secure connection requirement ✅ 
  certPath: /certs/ca.cert
```
____

Is a pre-requisite of https://github.com/open-feature/open-feature-operator/issues/389 

[1] - https://github.com/open-feature/flagd/blob/2a4783ef61eb3756e7a3f99aae596392906cd765/docs/configuration/configuration.md


core/pkg/runtime/from_config.go
```diff
@@ -218,24 +221,19 @@ func ParseSources(sourcesFlag string) ([]SourceConfig, error) {
...
 	}
-	for i, sp := range syncProvidersParsed {
+	for _, sp := range syncProvidersParsed {
 		if sp.URI == "" {
 			return syncProvidersParsed, errors.New("sync provider argument parse: uri is a required field")
 		}
 		if sp.Provider == "" {
 			return syncProvidersParsed, errors.New("sync provider argument parse: provider is a required field")
 		}
-		switch uriB := []byte(sp.URI); {
```
james-milligan, line 228: this switch should be added back in, it prevents errors if a developer was to use a prefix. While this isn't expected behaviour it shouldn't break if they do.

james-milligan: it should also be expanded to include the grpc(s) regex, which would prevent this from being a breaking change

Kavindu-Dodan: I disagree. 

If we allow such a fall-back mechanism, then we again have mixed usage of source configurations. And those lead to confusion (both users and developers). Not to mention documentation and potential backward compatibility issues.

My take is - we must be strict on what we support and make a clear separation between  `--uri` vs `--sources` based configurations. And document the correct behavior. Otherwise, we will end up with discussions like this [1]. 

Regarding the breaking change - It's best to do this now rather delaying further. 

[1] - https://github.com/open-feature/open-feature-operator/pull/408#discussion_r1143085223

docs/configuration/configuration.md
```diff
@@ -1,87 +1,143 @@
...
+| bearerToken | optional `string`  | Used for http sync and token get appended to `Authorization` header with [bearer schema](https://www.rfc-editor.org/rfc/rfc6750#section-2.1) |
+| providerID  | optional `string`  | Value binds to grpc connection's providerID field. GRPC server implementations may use this to identify connecting flagd instance            |
+| selector    | optional `string`  | Value binds to grpc connection's selector field. GRPC server implementations may use this to filter flag configurations                      |
+| grpcSecure  | optional `boolean` | Used to enable secure TLS connectivity for grpc sync. Default(ex:- if unset) is false, which will use insecure grpc connection               |
+| certPath    | optional `string`  | Used for grpcs sync when TLS certificate is needed. If not provided, system certificates will be used for TLS connection                     |
+
+The `uri` field values **do not** need to follow the [URI patterns](#uri-patterns), the provider type is instead derived
```
james-milligan, line 112: the phrasing here could be stronger, 'do not need' feels as though it would still work, whereas flagd would error if you do

Kavindu-Dodan: Thanks, yes will update this :)

Kavindu-Dodan: Updated the documentation to emphasize the difference between uri patterns and source configuration URI

core/pkg/runtime/from_config_test.go
```diff
@@ -34,55 +34,55 @@ func TestParseSource(t *testing.T) {
...
 					URI:      "config/samples/example_flags.json",
-					Provider: "file",
+					Provider: syncProviderFile,
 				},
 				{
 					URI:         "http://test.com",
-					Provider:    "http",
+					Provider:    syncProviderHTTP,
 					BearerToken: ":)",
 				},
```
thisthat, line 43: maybe worth adding another entry to the in/out array with `https` and `grpcs`

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
+| uri         | required `string`  | Flag configuration source of the provider                                                                                                    |
+| provider    | required `string`  | Provider type - `file`, `kubernetes`, `http` or `grpc`                                                                                       |
+| bearerToken | optional `string`  | Used for http sync and token get appended to `Authorization` header with [bearer schema](https://www.rfc-editor.org/rfc/rfc6750#section-2.1) |
+| providerID  | optional `string`  | Value binds to grpc connection's providerID field. GRPC server implementations may use this to identify connecting flagd instance            |
+| selector    | optional `string`  | Value binds to grpc connection's selector field. GRPC server implementations may use this to filter flag configurations                      |
+| grpcSecure  | optional `boolean` | Used to enable secure TLS connectivity for grpc sync. Default(ex:- if unset) is false, which will use insecure grpc connection               |
```
toddbaert, line 119: Is there a reason you chose `grpcSecure` instead of something like `tls` which is more generic and might even apply to yet-to-be-added syncs?

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
+## Sync providers
 
-| Sync       | Pattern                               | Example                               |
-|------------|---------------------------------------|---------------------------------------|
-| Kubernetes | `core.openfeature.dev/namespace/name` | `core.openfeature.dev/default/my-crd` |
-| Filepath   | `file:path/to/my/flag`                | `file:etc/flagd/my-flags.json`        |
-| Remote     | `http(s)://flag-source-url`           | `https://my-flags.com/flags`          |
-| Grpc       | `grpc(s)://flag-source-url`           | `grpc://my-flags-server`              |
+Sync providers are a core part of flagd. They are the sources of feature flag configurations for evaluations. flagd
+currently support the following sync providers,
```
toddbaert, line 28:
```suggestion
Sync providers are a core part of flagd; they are the abstraction that enables different sources for feature flag configurations.
flagd currently support the following sync providers:
```

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
-Custom sync providers can be used to provide flag evaluation logic.
+* [Kubernetes provider](#kubernetes-provider)
+* [Filepath Configuration](#filepath-provider)
+* [Remote Configuration](#remote-provider)
+* [GRPC Configuration](#grpc-provider)
 
 ### Kubernetes provider
 
-The Kubernetes provider allows flagD to connect to a Kubernetes cluster and evaluate flags against a specified FeatureFlagConfiguration resource as defined within the [open-feature-operator](https://github.com/open-feature/open-feature-operator/blob/main/apis/core/v1alpha1/featureflagconfiguration_types.go) spec.
+The Kubernetes sync provider allows flagD to connect to a Kubernetes cluster and evaluate flags against a specified
```
toddbaert, line 37:
```suggestion
The Kubernetes sync provider allows flagd to connect to a Kubernetes cluster and evaluate flags against a specified
```

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
 flagd start --uri core.openfeature.dev/default/my_example
 ```
 
-## Source Configuration
+In this example, `default/my_example` expected to be a valid FeatureFlagConfiguration resource, where `default` is the
+namespace and `my_example` being the resource name.
+
+### Filepath provider
+
+File path sync provider reads and watch the source file for updates(ex:- changes and deletions).
```
toddbaert, line 53:
```suggestion
The file path sync provider reads and watch the source file for updates(ex:- changes and deletions).
```

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
+
+```shell
+flagd start --uri file:etc/featureflags.json
+```
+
+In this example, `etc/featureflags.json` is a valid feature flag configuration file accessible by the flagd runtime.
+
+### Remote provider
+
+Remote sync provider fetch flags from a remote source and periodically poll the source for flag configuration updates.
```
toddbaert, line 63:
```suggestion
The HTTP sync provider fetch flags from a remote source and periodically poll the source for flag configuration updates.
```

We've generally tried to replace "remote" to "http". "grpc" is also possibly "remote" so "http" is a better name.

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
+```shell
+flagd start --uri https://my-flag-source.json
+```
+
+In this example, `https://my-flag-source.json` is a remote endpoint responding valid feature flag configurations when
+invoked with **HTTP GET** request.
+
+### GRPC provider
+
+GRPC sync provider stream flag configurations from a grpc sync provider implementation. This stream connection is ruled
```
toddbaert, line 74:
```suggestion
The GRPC sync provider stream flag configurations from a grpc sync provider implementation. This stream connection is ruled
```

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
 
 ```shell
 flagd start --uri core.openfeature.dev/default/my_example
 ```
 
-## Source Configuration
+In this example, `default/my_example` expected to be a valid FeatureFlagConfiguration resource, where `default` is the
+namespace and `my_example` being the resource name.
+
+### Filepath provider
```
toddbaert, line 51: Thanks for adding blurbs about each provider, these are great!

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
...
+    provider: http
+    bearerToken: bearer-dji34ld2l
+  - uri: default/my-flag-config
+    provider: kubernetes
+  - uri: my-flag-source:8080
+    provider: grpc
+  - uri: my-flag-source:8080
+    provider: grpc
+    certPath: /certs/ca.cert
+    grpcSecure: true
```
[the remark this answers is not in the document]
Kavindu-Dodan: > What happens if `certPath` is set but `secure` is not

Right now, this results in an `insecure` connection [1] . I am not against making this an error. Other option is to log a warning and continue :) 

[1] - https://github.com/open-feature/flagd/blob/8501dc0648e6c48954d871cf62f46016aa4cc490/core/pkg/sync/grpc/credentials/builder.go#L24

docs/configuration/configuration.md
toddbaert: I don't think it needs to be done in this PR, but the HTTP sync might be improved if it was made more consistent with GRPC. It should also be able to use `tls` and `certpath` (I think). IMHO that's a reason to use `tls` right now. `certPath` is sot of the same - it's grpc specific right now but the same concept applies to HTTP.

toddbaert: Agred

docs/configuration/configuration.md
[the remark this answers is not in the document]
thisthat: IMHO we should not fail, since the connection might still work. I am also in favor of a warning log.

docs/configuration/configuration.md
Kavindu-Dodan: Updated to use `tls` [1]

[1] - https://github.com/open-feature/flagd/blob/2a4783ef61eb3756e7a3f99aae596392906cd765/docs/configuration/configuration.md#source-configuration

core/pkg/runtime/from_config_test.go
Kavindu-Dodan: Resolving after adding required tests :)

docs/configuration/configuration.md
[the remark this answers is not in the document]
Kavindu-Dodan: @toddbaert @thisthat I went ahead with an error. The justification here is to fail fast.

If cert path is provided for an insecure (`tls:false`) sync, then this is a wrong configuration. The check is done at grpc credentials module [1] as this is the only place we currently use this config option. 


[1] - https://github.com/open-feature/flagd/blob/2a4783ef61eb3756e7a3f99aae596392906cd765/core/pkg/sync/grpc/credentials/builder.go#L24-L28

docs/configuration/configuration.md
```diff
@@ -1,87 +1,166 @@
 # Configuration
 
-`flagd` supports configuration via config file, environment variables and flags.
-In cases of conflict, flags have the
-highest priority, followed by environment variables and finally config file.
+<!-- TOC -->
```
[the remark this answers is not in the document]
Kavindu-Dodan: Used idea md plugin [1]. I am not super familiar with this tool so I will create an issue out of this suggestion :) 

[1] - https://www.jetbrains.com/help/idea/markdown.html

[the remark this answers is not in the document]
Kavindu-Dodan: Added a dedicated issue to improve this - https://github.com/open-feature/flagd/issues/570

docs/configuration/configuration.md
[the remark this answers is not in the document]
toddbaert: Works for me!

toddbaert (approved):
LGTM!

beeme1mr (approved):
Nice change, thanks!

PR #566, opened 2023-03-27
chore: move startServer functions into errGroups

bacherfl:
This PR is a follow up of #554 to make use of `errGroups` for starting the server and metrics server of the flag evaluation component, as discussed in https://github.com/open-feature/flagd/pull/554#discussion_r1147179222


core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -40,40 +46,42 @@ func (s *ConnectService) Serve(ctx context.Context, eval eval.IEvaluator, svcCon
...
-	}()
-
-	go s.startMetricsServer(svcConf)
-
-	select {
-	case err := <-errChan:
+		return nil
+	})
+	err := g.Wait()
+	if err != nil {
```
thisthat, line 81: nit
```suggestion
	if err := g.Wait(); err != nil {
```

thisthat (approved):
LGTM only a minor NIT

Kavindu-Dodan (approved):

PR #578, opened 2023-03-30
ci: improve homebrew release process

beeme1mr:
## This PR

- adds a base `release-type` as required by [Release Please](https://github.com/google-github-actions/release-please-action/blob/main/action.yml#L20-L22)
- update homebrew commit message to not include the flagd prefix

### Notes

Homebrew uses the first line of the commit message as the PR title and the rest of the message as the PR body. Without a body, Homebrew will flag the PR as invalid.

Addresses: https://github.com/Homebrew/homebrew-core/pull/127098#issuecomment-1490624290

### DIFF
--- a/.github/workflows/release-please.yaml
+++ b/.github/workflows/release-please.yaml
@@ -22,6 +22,7 @@ jobs:
       - uses: google-github-actions/release-please-action@ee9822ec2c397e8a364d634464339ac43a06e042 # v3
         id: release
         with:
+          release-type: "go"
           command: manifest
           token: ${{secrets.GITHUB_TOKEN}}
           default-branch: main
@@ -35,7 +36,7 @@ jobs:
         uses: actions/github-script@98814c53be79b1d30f795b907e553d8679345975 # v6
         id: items-to-publish
         env:
-          CHANGED_ITEMS: '${{ steps.release.outputs.paths_released }}'
+          CHANGED_ITEMS: "${{ steps.release.outputs.paths_released }}"
         with:
           # Release please outputs a string representation of an array under the key paths_released
           # This script parses the output, and filters each of the changed items (provided as paths to the root of the package, e.g. flagd or core)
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yaml
```diff
@@ -214,19 +215,20 @@ jobs:
...
+    # Only run on non-forked flagd releases
+    if: ${{ github.repository_owner == 'open-feature' && needs.release-please.outputs.flagd_tag_name }}
+    steps:
+      - uses: mislav/bump-homebrew-formula-action@v2
+        with:
+          formula-name: flagd
+          tag-name: ${{ needs.release-please.outputs.flagd_tag_name }}
+          download-url: https://github.com/${{ github.repository }}.git
+          commit-message: |
+            {{formulaName}} ${{ needs.release-please.outputs.flagd_version }}
```
Kavindu-Dodan, line 230: According to the documentation of this action [1], commit-message expect the `{{version}}` placeholder. 

And we set this here [2] . Am I missing something 🤔 

Besides, what went wrong here should be the missing new line 

> It's recommended that commit-message has both subject and body, i.e. that it contains a subject line followed by a blank line followed by body text. 


[1] - https://github.com/mislav/bump-homebrew-formula-action#action-inputs
[2] - https://github.com/open-feature/flagd/blob/ef341f3895f9df75bff795e984415c7d7f0dadbd/.github/workflows/release-please.yaml#L161

beeme1mr: `{{version}}` uses the git tag we're passing in. That is why the PR title was `flagd flagd/v0.5.0` instead of `flagd 0.5.0`. This should address that by not using templating offered by the Homebrew action and instead setting it ourselves.

We explicitly set the flagd version output [here](https://github.com/open-feature/flagd/blob/main/.github/workflows/release-please.yaml#L54). You can see it in action [here](https://github.com/open-feature/flagd/actions/runs/4566549553/jobs/8059489382) under "Dump Release Please Output". The VERSION environment variable you linked is set defined in the `release-go-binaries` job.

Kavindu-Dodan: Thank you. I understand it now :) 

I missed the `$` in the replacement (`${{ needs.release-please.outputs.flagd_version }}`) which is an interpolation.

Kavindu-Dodan (approved):

Issue #575, opened 2023-03-29
OTel Traces for evaluation in flagd


beeme1mr:
- traces for flag evaluation
  - initially, we should probably sample everything
  - subsequently, we may want to investigate optimizing a sampling algorithm to sample only SOME evaluations
  - see semantic conventions for [gRPC](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md) and [HTTP](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md), and possibly [feature flags](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/feature-flags/)
- try to continue trace for existing context (incoming request) if possible

beeme1mr:
https://buf.build/blog/connect-opentelemetry-go/

Kavindu-Dodan:
Given that I already worked on #563, I would like to continue on this topic :)  @toddbaert  or @thisthat any opinion ?

Kavindu-Dodan:
>  Start or extend a trace for a change event (e.g. file, HTTP, grpc, kubernetes watcher)

Investigated the distributed traces between grpc & grpc server implementation (ex:- flagd-proxy) for grpc change event. `otelgrpc` interceptor spans work for the full stream (i.e - stream start to stram end). This is not suitable for flagd use cases as we have long-running grpc streams.

PR #586, opened 2023-04-03
feat: flagd OTEL collector

### LINKED ISSUE #563: Support OTel collector in flagd
In order to get data into OTel, we need to be able to configure flagd to communicate with a collector. This entails:

- a new CLI arg for flagd, specifying the collector (taking the collector URI)
  - if CLI arg is set, configure OTLP in the OTel SDK (see example [here](https://github.com/keptn/lifecycle-toolkit/blob/main/operator/controllers/common/otel_utils.go#L60-L73))
  - deploy [collector](https://opentelemetry.io/docs/collector/getting-started/) + configure it with Jaeger and confirm works as expected 
-  documentation and tests for the above

Kavindu-Dodan:
## This PR

Fixes #563

Introduce configurations for flagd to connect to OTEL collector. Also, improves how telemetry configurations are handled.

Changes include,

- Package renaming - otel to telemetry - This avoids import conflicts 

- Introduce a telemetry builder - The intention is to have a central location to handle telemetry configurations and build telemetry components 

- Introduce a span processor builder - Provide groundwork for https://github.com/open-feature/flagd/issues/575 (needs fallback mechanism)

## How to test?

Consider following this guide - (doc generated from commit)  https://github.com/open-feature/flagd/blob/81c66b3c89540b475fe0a46ac89869800f7b74ae/docs/configuration/flagd_telemetry.md

In short, 

- create configuration files (docker-compose yaml, collector config yaml, Prometheus yaml  )  
- start collector setup (docker-compose up)
- start flagd with otel collector override for metrics (`flagd start --uri file:/flags.json --metrics-exporter otel --otel-collector-target localhost:4317`)

Metrics will be available at Prometheus(http://localhost:9090/graph). Traces are still missing as we have to implement them.


core/pkg/telemetry/global.go
```diff
@@ -0,0 +1,150 @@
...
+	responseSizeName    = "http_response_size_bytes"
+	metricsExporterOtel = "otel"
+	exportInterval      = 2 * time.Second
+)
+
+// SetupGlobal setup global telemetry configurations based on provided configurations. This needs to be run before
+// obtaining any global.Meter or otel.Tracer based on global providers. Note that, metrics exporter works independent
+// of the collector target. However, overriding metrics exporter require collector target as we only support otel
+// collector
+func SetupGlobal(ctx context.Context, serviceName string, metricsExporter string, collectorTarget string) error {
```
Kavindu-Dodan, line 34: @thisthat  need your input on this approach of setting tracer and meter providers globally

thisthat: I don't like global objects since they introduce race conditions when executing tests in parallel (besides all other antipatterns). I suggest creating a file similar to what we have for metrics. Please also take a look at my comment on that file.

core/pkg/telemetry/metrics.go
```diff
@@ -90,7 +80,7 @@ func NewOTelRecorder(exporter metric.Reader, serviceName string) *MetricsRecorde
 	hsize, _ := meter.Float64Histogram(
 		responseSizeName,
 		instrument.WithDescription("The size of the HTTP responses"),
-		instrument.WithUnit(unit.Bytes),
+		instrument.WithUnit("By"),
```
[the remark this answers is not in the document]
Kavindu-Dodan: Unit is deprecated. See - https://pkg.go.dev/go.opentelemetry.io/otel/metric/unit#pkg-types

core/pkg/telemetry/metrics.go
```diff
@@ -69,19 +69,9 @@ func getDurationView(svcName, viewName string, bucket []float64) metric.View {
 	)
 }
 
-func NewOTelRecorder(exporter metric.Reader, serviceName string) *MetricsRecorder {
-	const requestDurationName = "http_request_duration_seconds"
-	const responseSizeName = "http_response_size_bytes"
+func NewOTelRecorder(serviceName string) *MetricsRecorder {
+	meter := global.Meter(serviceName)
```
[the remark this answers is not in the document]
Kavindu-Dodan: Thanks, yes that's a fair argument. 

The advantage of global provider is that we can use the provider through `global`, similar to the usage here in any location where we need metrics. However, yes drawbacks are there.

Instead of the global approach, I will try to unify startup parameter handling and yield a reader instead. That way, individual metric handlers can build their provider based on the reader.

core/pkg/telemetry/metrics.go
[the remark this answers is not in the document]
Kavindu-Dodan: See this for the removal - https://github.com/open-telemetry/opentelemetry-go/pull/3814 :)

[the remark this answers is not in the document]
thisthat: Thank you 👍

core/pkg/telemetry/metrics.go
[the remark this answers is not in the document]
thisthat: Something worth mentioning: since we are the Application code, we should set the global providers. This way, libraries that have OTel API support can expose traces and metrics :)

core/pkg/runtime/from_config.go
```diff
@@ -126,6 +129,24 @@ func FromConfig(logger *logger.Logger, config Config) (*Runtime, error) {
 	}, nil
 }
 
+// buildTelemetryRecorder is a helper to build telemetry.MetricsRecorder based on configurations
+func buildTelemetryRecorder(logger *logger.Logger, config telemetry.Config) (*telemetry.MetricsRecorder, error) {
+	// Build metric reader based on configurations
+	mReader, err := telemetry.BuildMetricReader(context.TODO(), config)
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to setup metric reader: %v", err))
```
bacherfl, line 137: since any error that occurs during the setup will by logged in [start.go](https://github.com/open-feature/flagd/blob/c9aba7ffd226c8eca363209528f56fe6dc575740/flagd/cmd/start.go#L163), I would rather return a wrapped error here:

```
return nil, fmt.Errorf("failed to setup metric reader: %w", err)
```

docs/configuration/flagd_start.md
```diff
@@ -15,7 +15,9 @@ flagd start [flags]
   -e, --evaluator string                    DEPRECATED: Set an evaluator e.g. json, yaml/yml.Please note that yaml/yml and json evaluations work the same (yaml/yml files are converted to json internally) (default "json")
   -h, --help                                help for start
   -z, --log-format string                   Set the logging format, e.g. console or json  (default "console")
+  -t, --metrics-exporter string             Set the metrics exporter. Default(if unset) is Prometheus. Can be override to otel - OpenTelemetry metric exporter. Overriding to otel require otelCollectorTarget to be present
```
thisthat, line 18: nit
```suggestion
  -t, --metrics-exporter string             Set the metrics exporter. Default (if unset) is Prometheus. Can be override to otel - OpenTelemetry metric exporter. Overriding to otel require otelCollectorTarget to be present
```

docs/configuration/flagd_start.md
```diff
@@ -15,7 +15,9 @@ flagd start [flags]
...
   -m, --metrics-port int32                  Port to serve metrics on (default 8014)
+  -o, --otel-collector-target string        Set the grpc target of the otel collector for flagd runtime. If unset, collector setup will be ignored resulting in ignored traces(NoopTracerProvider)
```
thisthat, line 20:
```suggestion
  -o, --otel-collector-uri string        Set the grpc URI of the OpenTelemetry collector for flagd runtime. If unset, the collector setup will be ignored and traces will not be exported.
```

core/pkg/service/flag-evaluation/flag_evaluator_test.go
```diff
@@ -96,7 +100,8 @@ func TestConnectService_ResolveAll(t *testing.T) {
 				t.Errorf("ConnectService.ResolveAll() error = %v, wantErr %v", err.Error(), tt.wantErr.Error())
 				return
 			}
-			data, err := exp.Collect(context.TODO())
+			var data metricdata.ResourceMetrics
```
bacherfl, line 103: out of scope for this PR, but I believe we could improve here by creating an interface for the metrics recorder which can be mocked here. Right now we are also testing functionality of that component here, which exceeds the scope of the flag evaluation component in my opinion.

thisthat (approved):

docs/configuration/flagd_start.md
Kavindu-Dodan: went for `target` as this is kind of the grpc terminology at API level [1] , but yes let's go for `uri`

[1] - https://pkg.go.dev/google.golang.org/grpc#Dial

core/pkg/runtime/from_config.go
Kavindu-Dodan: Good suggestion. I improve error handling in this function based on the feedback :) 

Will update with the next commit

core/pkg/service/flag-evaluation/flag_evaluator_test.go
Kavindu-Dodan: Yes, we could introduce a mock for `MetricsRecorder`. But, given that we validate recorded data, mock will need functionality similar to the real implementation.

And regarding the change here - this is an api change to have out variable assignment instead of a return

core/pkg/telemetry/metrics.go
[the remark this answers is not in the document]
Kavindu-Dodan: Updated documentation to have current telemetry overview [1]

[1] - https://github.com/open-feature/flagd/blob/301e74f1571b67a6dd7cc87ef8b86025d698314e/docs/configuration/flagd_telemetry.md

bacherfl (approved):

toddbaert (approved):
I'm not an OTel expert, but this looks fine to me and I trust @bacherfl and @thisthat 's expertise.

Issue #590, opened 2023-04-06
[BUG] flagd-proxy locking issue


james-milligan:
### Observed behavior

During load testing it was found that there is a misconfiguration in the locks for the sync store, resulting in fatal errors under high load.

This only seems to be triggered +1000 connections

### Expected Behavior

flags-proxy should be able to handle simultaneous connections without failure

### Steps to reproduce

This project can be used to stress test the flags-proxy, resulting in the fatal error:
https://github.com/james-milligan/flagd-proxy-profiling

PR #592, opened 2023-04-06
fix: flagd-proxy locking bug fix

james-milligan:
## This PR

- resolves issue with locks found during load testing

### Related Issues

https://github.com/open-feature/flagd/issues/590

### Notes

dependent upon https://github.com/open-feature/flagd/pull/589


core/pkg/sync-store/sync_store.go
```diff
@@ -132,13 +131,16 @@ func (s *SyncStore) RegisterSubscription(
...
 			dataSync: dataSync,
 		}
-
 		// access pointer + trigger resync passing the dataSync
 		if sh.syncRef != nil {
 			go func() {
-				s.logger.Debug(fmt.Sprintf("sync handler exists for target %s, triggering a resync", target))
-				if err := sh.syncRef.ReSync(ctx, dataSync); err != nil {
-					errChan <- err
+				s.mu.Lock()
```
bacherfl, line 137: can we use `RLock/RUnlock` here? It seems like the syncHandlers are not modified heare, so a read lock might be sufficient

toddbaert: Ya, agreed with @bacherfl , this seems like only read-ops.

bacherfl (approved):

beeme1mr (approved):

PR #598, opened 2023-04-10
feat: otel traces for flag evaluation

Kavindu-Dodan:
## This PR

Partially address #575 by introducing **traces** for flag evaluations & otlp **batch exporter** based on **otlp grpc** with aim to use by otel collector.

Contents,

- Refactor telemetry builder to register `TraceProvider` [1].
- Tracers [2] for `flag_evaluator.go` & `json_evaluator.go` with passing context to derive span [2] tree

```
flagEvaluationService
|__ jsonEvaluator
```

Note that, setting `otel-collector-uri` is essential to export collected traces. However, flagd can still be run **without** collector uri -  <del>this results in an empty(nil) exporter with no-op behavior</del> this results in no provider registration, making traces to use `NoopTracerProvider`

## How to test 

1. Set up flagd with otel collector - follow documentation [4]
2. Perform flag evaluations  [5]
3. Visit Jaeger UI (typically - http://127.0.0.1:16686/) for traces 

overview
trace

[1] - https://opentelemetry.io/docs/reference/specification/trace/api/#tracerprovider 
[2] - https://opentelemetry.io/docs/reference/specification/trace/api/#tracer
[3] - https://opentelemetry.io/docs/reference/specification/trace/api/#span
[4] - https://github.com/open-feature/flagd/blob/main/docs/configuration/flagd_telemetry.md#export-to-otel-collector
[5] - https://github.com/open-feature/flagd/blob/main/docs/configuration/fractional_evaluation.md#example


core/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -269,6 +259,42 @@ func (s *FlagEvaluationService) ResolveObject(
...
+	logger *logger.Logger,
+	resolver func(context context.Context, reqID, flagKey string, ctx *structpb.Struct) (T, string, string, error),
+	flagKey string,
+	evaluationContext *structpb.Struct,
+	resp response[T],
+	metrics *telemetry.MetricsRecorder,
+) error {
+	reqID := xid.New().String()
+	defer logger.ClearFields(reqID)
+
```
thisthat, line 274: as a follow-up, we could do a bit of refactoring to the telemetry pkg to expose a utility function to create the semantic convention for FF instead of having that embedded into the metric part. 
This way we can add an event to the span when the evaluation is performed as the semantic conventions suggest ([docs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md#convention)).

core/pkg/telemetry/builder.go
```diff
@@ -23,48 +25,59 @@ const (
...
+	var err error
+
 	if cfg.CollectorTarget == "" {
-		return nil, fmt.Errorf("otel collector target is required for span processor")
+		logger.Warn("collector target is not set. Falling back to empty(nil) trace exporter performing no action")
+	} else {
+		exporter, err = buildOtlpExporter(ctx, cfg.CollectorTarget)
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
Kavindu-Dodan: Thanks, good suggestion :) 

However, I had to change the logic here as the nil span processor caused panic. Hence, the logic is now changed and clearer :) Please review again and let me know.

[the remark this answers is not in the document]
Kavindu-Dodan: done :)

core/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -50,13 +56,18 @@ func (s *FlagEvaluationService) ResolveAll(
 ) (*connect.Response[schemaV1.ResolveAllResponse], error) {
```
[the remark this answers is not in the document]
Kavindu-Dodan: Will resolve the conversation and let's revisit this topic if we see the need :)

beeme1mr (approved):
This looks great to me and works well. Thanks for putting this together!

As a follow-up, ut would be nice if the service name and version were defined at the service level. You can see the semantic convention for that [here](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/).

Kavindu-Dodan:
> This looks great to me and works well. Thanks for putting this together!
> 
> As a follow-up, ut would be nice if the service name and version were defined at the service level. You can see the semantic convention for that [here](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/).

Thank you for the feedback and review 

Service name is already available in both metrics and traces. I will add the service version with the next commit.

core/pkg/service/flag-evaluation/flag_evaluator.go
Kavindu-Dodan: I hope you are referring to this section,

`A flag evaluation SHOULD be recorded as an Event on the span during which it occurred.`

Isn't this highlighting the feature flag evaluation side (client sdk or server sdk) of the evaluation? If you can explain this a bit, I can follow up on the implementation :)

thisthat: Yes, this is the section I am referring to :) 
Currently, while recording the metric we add as a dimension the three attributes `provider`, `key`, and `variant`. We could refactor that piece of code to expose the creation of these attributes and then in this piece of code we could add the following after the call to `resolver`:
```go
span := trace.SpanFromContext(ctx)
span.SetAttributes(telemetry.FFAttributes(flagKey, variant))
```
`SpanFromContext` returns the current active Span in the context. If no Span is active in the context, a noop span implementation is returned. `telemetry.FFAttributes(string,string)` is the function that builds the attribute for the FF semantic convention. 
I hope this clarifies a bit more. Let me know if something is still unclear :)

odubajDT (approved):
Nice work, thanks! :)

core/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -50,13 +56,18 @@ func (s *FlagEvaluationService) ResolveAll(
...
+
+	sCtx, span := s.flagEvalTracer.Start(ctx, "resolveAll")
+	defer span.End()
+
 	res := &schemaV1.ResolveAllResponse{
 		Flags: make(map[string]*schemaV1.AnyFlag),
 	}
-	values := s.eval.ResolveAllValues(reqID, req.Msg.GetContext())
+	values := s.eval.ResolveAllValues(sCtx, reqID, req.Msg.GetContext())
+	span.SetAttributes(attribute.Int("count", len(values)))
```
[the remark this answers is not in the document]
Kavindu-Dodan: Yup, it is an overkill :) [1]

Resolving the conversation

[1] - https://github.com/open-feature/flagd/blob/24558321e185fa47907a283ee3b7c287a057802f/core/pkg/service/flag-evaluation/flag_evaluator.go#L69

core/pkg/service/flag-evaluation/flag_evaluator.go
Kavindu-Dodan: Thanks for the explanation. Yes, it makes sense to add span attributes as you suggest. 

Will refactor and add this with the next commit.

Kavindu-Dodan: @thisthat please check the proposed improvement [1] 

And here is the span with attributes 

image

[1] - https://github.com/open-feature/flagd/blob/8eb1402b15a02230fdd9a8e4dd9e8fe681aec267/core/pkg/telemetry/utils.go

core/pkg/eval/json_evaluator.go
```diff
@@ -57,6 +65,12 @@ func (je *JSONEvaluator) GetState() (string, error) {
 }
 
 func (je *JSONEvaluator) SetState(payload sync.DataSync) (map[string]interface{}, bool, error) {
+	_, span := je.jsonEvalTracer.Start(
+		context.Background(),
+		"setState",
+		trace.WithAttributes(attribute.String("source", payload.Source)))
```
thisthat, line 71: to avoid confusion with the [messaging semconv](https://github.com/open-telemetry/opentelemetry-specification/blob/25f513dbea2810be1c6f94d314d1a8d7e7b592c9/semantic_conventions/trace/messaging.yaml#L59), I suggest the following
```suggestion
		trace.WithAttributes(attribute.String("feature_flag.source", payload.Source)))
```

thisthat (approved):
Great addition 🚀

PR #599, opened 2023-04-11
feat: introduce flagd-proxy profiling tool

james-milligan:
## This PR

- introduces a profiling tool for the flagd-proxy to test the effects of concurrent connections alongside the size of flag definitions and their effects upon response time, from configuration change to broadcast completion


beeme1mr:
Would you consider this a feature that should trigger a new version?

james-milligan:
> Would you consider this a feature that should trigger a new version?

changed type to test

flagd-proxy/tests/loadtest/go.mod
```diff
@@ -0,0 +1,18 @@
+module github.com/open-feature/flagd/flagd-proxy/tests/loadtest
+
+go 1.20
```
Kavindu-Dodan, line 3: minor - all other modules are at 1.19. should we aim the same here?

flagd-proxy/tests/loadtest/pkg/handler/handler.go
```diff
@@ -0,0 +1,179 @@
...
+	finishedSuccess := make(chan bool, 1)
+
+	errChan := make(chan error, 1)
+
+	fmt.Printf("starting %d watchers...\n", watchers)
+	var c syncv1grpc.FlagSyncServiceClient
+	var err error
+
+	for i := 0; i < watchers; i++ {
+		if i%250 == 0 {
```
Kavindu-Dodan, line 105: I assume we are having connection issues when reaching ~250 connections ?

Kavindu-Dodan (approved):
approving with minor remarks :) no major concerns and given this is a test suite, we can always improve/change

flagd-proxy/tests/loadtest/pkg/handler/handler.go
james-milligan: yes this is the default maximum number of streams per client, this can also be enforced on the client side, but I felt it safer to stay within the default bounds for this tool 
https://github.com/golang/net/blob/9001ca7de9d738aed247bf005debccb5b186a965/http2/server.go#L58

PR #616, opened 2023-04-14
refactor: introduce additional linting rules + fix discrepancies 

james-milligan:
## This PR

- extends the scope of linting
- fixes all linting issues for the new scope


beeme1mr (approved):

Kavindu-Dodan (approved):
Approving - I like the error wrapping as it gives more context in logs

Issue #630, opened 2023-04-25
[BUG] file path sync is not working in windows


Kavindu-Dodan:
### Observed behavior

A community user reached and noted that file path sync is not working for flagd when used in Windows OS.  According to the explanation, flag definitions were not updated when the flag source file is changed.

### Expected Behavior

flag definitions must get updated correctly from file source

toddbaert:
[fsnotify](https://github.com/fsnotify/fsnotify) is supposed to support windows, but I've had trouble getting it working, as has @heckelmann .

Each OS uses a different underlying mechanism for seeing fs changes, so the behaviors might not be exactly the same.

Kavindu-Dodan:
Thank you @toddbaert  

@RealAnna  will investigate further on this and let's try to fix OR document this behavior for Windows

RealAnna:
Tested by building main directly on windows 11. 
The only issue I had was I needed to allow networking when the windows defender popped up. After that, I could run flagd on a command prompt and modify the file. A few seconds are needed for the update of the flag to be visible.

The curl command in the README only works within bash, since windows curl has a different syntax.

![image](https://user-images.githubusercontent.com/89971034/234795295-53e747dd-c9d0-45ea-ba85-0560639e3a72.png)

```
$ curl -X POST "localhost:8013/schema.v1.Service/ResolveString" -d '{"flagKey":"myStringFlag","context":{}}' -H "Content-Type: application/json"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    93  100    54  100    39  42959  31026 --:--:-- --:--:-- --:--:-- 
93000{"value":"val12", "reason":"STATIC", "variant":"key1"}

 ~
$ curl -X POST "localhost:8013/schema.v1.Service/ResolveString" -d '{"flagKey":"myStringFlag","context":{}}' -H "Content-Type: application/json"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    95  100    56  100    39  28169  19617 --:--:-- --:--:-- --:--:-- 
95000{"value":"val1333", "reason":"STATIC", "variant":"key1"}
```

beeme1mr:
@Kavindu-Dodan can we close this issue now that a workaround is documented?

Kavindu-Dodan:
[...]

PR #632, opened 2023-04-27
chore: add instructions for windows and fix failing unit tests

RealAnna:
## This PR

- fixes unit tests that failed on windows machines
- clarifies usage of flagd on a windows machine

### Related Issues

https://github.com/open-feature/flagd/issues/630

### Notes

The learning from this pr were:

1.  Always prefer using the go test suite tempdir instead of relying on os library to allocate and delete test folders, this simplifies code and avoids os-specific issues
2.  Make sure to always close file pointers, since leaving dangling ones tends to cause access errors in Windows aka error file already open. 
3.  When running via docker one must make sure that the file is in the same filesystem where the docker container runs than the watcher will work. AKA on Windows it is good as long as everything comes from WSL filesystem

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,8 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
     ```
 
     Or use docker:
-   (for Windows use WSL both for the file path and the docker commands. This flagd on docker command WILL NOT WORK on Windows due to the mixed filesystems [#8479](https://github.com/docker/for-win/issues/8479)!! See also [this medium article](https://levelup.gitconnected.com/docker-desktop-on-wsl2-the-problem-with-mixing-file-systems-a8b5dcd79b22) )
+    _Note - In Windows, use WSL system for both the file location and Docker runtime. Mixed file systems does not
+    work and this is a limitation of Docker (https://github.com/docker/for-win/issues/8479)_
 
     ```sh
     docker run \
@@ -74,7 +75,7 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
       --uri file:./example_flags.flagd.json
     ```
 
-    Or use docker (!! this does not work on Windows without using WSL !!):
+    Or use docker ( _Note - In Windows, this require WSL system for both the file location and Docker runtime_):
 
     ```sh
     docker run \
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
 
     Or use docker:
     _Note - In Windows, use WSL system for both the file location and Docker runtime. Mixed file systems does not
-    work and this is a [limitation of Docker] (https://github.com/docker/for-win/issues/8479)_
+    work and this is a [limitation of Docker](https://github.com/docker/for-win/issues/8479)_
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -98,6 +98,13 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
       -d '{"flagKey":"myStringFlag","context":{}}' -H "Content-Type: application/json"
     ```
 
+    For Windows use bash or use the following:
```
thisthat, line 101:
```suggestion
    For Windows we recommend using a [WSL](https://learn.microsoft.com/en-us/windows/wsl/install) terminal. 
    Otherwise, use the following with `cmd` or `powershell` 
```

RealAnna: powershell wont work with this curl, but I will commit the other suggestions 👍

README.md
```diff
@@ -49,6 +49,7 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
     ```
 
     Or use docker:
+   (for Windows use WSL both for the file path and the docker commands. This flagd on docker command WILL NOT WORK on Windows due to the mixed filesystems [#8479](https://github.com/docker/for-win/issues/8479)!! See also [this medium article](https://levelup.gitconnected.com/docker-desktop-on-wsl2-the-problem-with-mixing-file-systems-a8b5dcd79b22) )
```
Kavindu-Dodan, line 52: I would reword this section to something like the below (italic text & in a new line), 

```
    _Note - In Windows, use WSL system for both the file location and Docker runtime. Mixed file systems does not 
   work and this is a limitation of Docker (https://github.com/docker/for-win/issues/8479)_
```

I have mixed feelings on linking to the medium article as this is a premium access only. Plus, rewording to highlight this comes from Docker rather than flagd :)

README.md
```diff
@@ -73,7 +74,7 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
       --uri file:./example_flags.flagd.json
     ```
 
-    Or use docker:
+    Or use docker (!! this does not work on Windows without using WSL !!):
```
Kavindu-Dodan, line 77: Moving this to italic text & in a new line with a `Note`

And rather than saying `does not work`, reword to say something like "In Windows, this requires WSL"

README.md
toddbaert: agreed.

toddbaert (commented):
LGTM but I agree with @Kavindu-Dodan 's suggestions.

README.md
```diff
@@ -50,7 +50,7 @@ Experiment with flagd in your browser using [the Killercoda tutorial](https://ki
 
     Or use docker:
     _Note - In Windows, use WSL system for both the file location and Docker runtime. Mixed file systems does not
-    work and this is a limitation of Docker (https://github.com/docker/for-win/issues/8479)_
+    work and this is a [limitation of Docker] (https://github.com/docker/for-win/issues/8479)_
```
thisthat, line 53:
```suggestion
    work and this is a [limitation of Docker](https://github.com/docker/for-win/issues/8479)_
```

Kavindu-Dodan (approved):
lgtm

Issue #650, opened 2023-05-11
[FEATURE] Add feature flag key as a dimension on the reason metric


beeme1mr:
## Overview

The reasons metric captures a count of the various [Reasons](https://openfeature.dev/specification/types#resolution-details) a flag was evaluated the way it was. It can be used to see how many times an invalid flag key is used or how after a starting rule matches. Currently, it's not possible to filter by specific flag keys to see the reason to count that flag. This was to prevent an explosion of dimensions if the flag key was invalid. However, using an option dimension, this value could be set only if the flag key is valid.

## Requirements

- `feature_flag.key` is added as a dimension on the `reasons` metric only for valid feature flags. Invalid flag keys do NOT include this dimension.

Kavindu-Dodan:
We should be able to do this fairly easily through adding an conditional attribute(dimension) at out logic [1]

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/telemetry/metrics.go#L80-L85

Issue #646, opened 2023-05-09
OTel Traces for Flag Source Change Events in flagd


Kavindu-Dodan:
The runtime package in flagd should be updated to create spans with each message to the dataSync channel. This will allow us to see flag change events in OTel.

Kavindu-Dodan:
PR https://github.com/open-feature/flagd/pull/653 extends the already available span for flag syncs.

PR #653, opened 2023-05-11
feat: telemetry improvements

Kavindu-Dodan:
## This PR

note - Includes minor refactoring, review commit by commit for for more clarity 

Fixes both #646 &  #650

Enrich flag sync span with more attributes. Renamed span name from `setState` to `flagSync`. 

Additional attributes,

- `feature_flag.sync_type` : Mapping to internal sync type (All, Add, Update, Delete) for better visibility
- `feature_flag.change_count`: Number of flag configurations changed through this sync 

image

The `reasons` metric now contains a flag key for valid (non-error) evaluations

image

### DIFF
--- a/core/pkg/telemetry/metrics.go
+++ b/core/pkg/telemetry/metrics.go
@@ -60,28 +60,32 @@ func (r MetricsRecorder) InFlightRequestEnd(ctx context.Context, attrs []attribu
 	r.httpRequestsInflight.Add(ctx, -1, metric.WithAttributes(attrs...))
 }
 
+func (r MetricsRecorder) RecordEvaluation(ctx context.Context, err error, reason, variant, key string) {
+	if err == nil {
+		r.Impressions(ctx, reason, variant, key)
+	}
+	r.Reasons(ctx, key, reason, err)
+}
+
 func (r MetricsRecorder) Impressions(ctx context.Context, reason, variant, key string) {
 	r.impressions.Add(ctx,
 		1,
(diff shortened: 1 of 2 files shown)


thisthat (approved):
:shipit:

core/pkg/eval/json_evaluator.go
```diff
@@ -67,29 +69,40 @@ func (je *JSONEvaluator) GetState() (string, error) {
...
-		return je.store.Add(je.Logger, payload.Source, newFlags.Flags), false, nil
+		events = je.store.Add(je.Logger, payload.Source, newFlags.Flags)
 	case sync.UPDATE:
-		return je.store.Update(je.Logger, payload.Source, newFlags.Flags), false, nil
+		events = je.store.Update(je.Logger, payload.Source, newFlags.Flags)
 	case sync.DELETE:
-		return je.store.DeleteFlags(je.Logger, payload.Source, newFlags.Flags), true, nil
+		events = je.store.DeleteFlags(je.Logger, payload.Source, newFlags.Flags)
 	default:
 		return nil, false, fmt.Errorf("unsupported sync type: %d", payload.Type)
```
bacherfl, line 99: should we also record an error for the span in this case?

bacherfl (approved):

core/pkg/eval/json_evaluator.go
bacherfl: thanks for the explanation - that sounds reasonable to me :)

Kavindu-Dodan: Cool, I will resolve this conversation

toddbaert (approved):

PR #658, opened 2023-05-17
feat: add `starts_with` and `ends_with` json evaluators

bacherfl:
This PR adds additional jsonLogic evaluators for `starts_with` and `ends_with`, which check if the given property starts/ends with the given value. Both evaluators return a boolean, indicating whether the condition has been met.

The following example shows how these evaluators can be used in a flag targeting rule:

```json
{
  "flags": {
    "headerColor": {
      "state": "ENABLED",
      "defaultVariant": "red",
      "variants": {
        "blue": "#0000FF",
        "green": "#00FF00",
        "red": "#FF0000",
        "yellow": "#FFFF00"
      },
      "targeting": {
        "if": [
          {
            "starts_with": [
              {
                "var": "email"
              },
              "user@faas"
            ]
          },
          "red",
          "green"
        ]
      },
      "source": ""
    }
  }
}
```

This rule will resolve to the `red` variant of the `headerColor` flag, if the `email` property within the evaluationContext starts with `user@faas`.


core/pkg/eval/string_comparison_evaluation.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// will have been resolved to
+//
+// ["user@faas.com", "user@faas"]
+//
+// at the time this function is reached.
+func parseStringComparisonEvaluationData(values interface{}) (string, string, error) {
+	parsed, ok := values.([]interface{})
+	if !ok {
+		return "", "", errors.New("starts_with evaluation is not an array")
```
thisthat, line 92: I don't have strong opinion how the log line looks like but I think we should provide info about "ends_with". 
If I start playing with the targeting rule using "ends_with" making a mistake and in the log I see "starts_with", I would be completely lost 😅 
```suggestion
		return "", "", errors.New("[start/end]s_with evaluation is not an array")
```

bacherfl: agree, will adapt that :)

core/pkg/eval/json_evaluator.go
```diff
@@ -59,6 +59,8 @@ func NewJSONEvaluator(logger *logger.Logger, s *store.Flags) *JSONEvaluator {
 		jsonEvalTracer: otel.Tracer("jsonEvaluator"),
 	}
 	jsonlogic.AddOperator("fractionalEvaluation", ev.fractionalEvaluation)
+	jsonlogic.AddOperator("starts_with", ev.startsWithEvaluation)
```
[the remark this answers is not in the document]
beeme1mr: I just double-checked JSON logic, and it looks like they use snake case. As a follow-up, it may be worth adding an alias for the `fractionEvaluation` operator to also follow this convention.

beeme1mr (commented):
The functionality looks good, but it needs to be documented.

core/pkg/eval/string_comparison_evaluation.go
```diff
@@ -0,0 +1,110 @@
...
+//			]
+//	}
+//
+// This rule can be applied to the following data object, where the evaluation will resolve to 'true':
+//
+// { "email": "user@faas.com" }
+//
+// Note that the 'starts_with'  evaluation rule must contain exactly two items, which both resolve to a
+// string value
+func (je *JSONEvaluator) startsWithEvaluation(values, _ interface{}) interface{} {
```
Kavindu-Dodan, line 28: I think we can get rid of the pointer receiver here. Yes, `fractionalEvaluation` also uses the same. But, the only usage is to access the logger and IMO this is not necessary - we should be able to inject a dedicated logger and make this independent (low coupling)

And going further, all custom operators can be aggregated similar to what we have here :) 

I think this is a good refactoring and if preferred can be a follow-up (including fractionalEvaluation )

core/pkg/eval/string_comparison_evaluation.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// at the time this function is reached.
+func parseStringComparisonEvaluationData(values interface{}) (string, string, error) {
+	parsed, ok := values.([]interface{})
+	if !ok {
+		return "", "", errors.New("[start/end]s_with evaluation is not an array")
+	}
+
+	if len(parsed) != 2 {
+		return "", "", errors.New("starts_with evaluation must contain a value and a comparison target")
```
Kavindu-Dodan, line 96: Minor - I think this needs correction on all error logs below

```suggestion
		return "", "", errors.New("[start/end]s_with evaluation must contain a value and a comparison target")
```

bacherfl:
> The functionality looks good, but it needs to be documented.

the documentation has now been added as well

thisthat (approved):

core/pkg/eval/string_comparison_evaluation.go
Kavindu-Dodan: @bacherfl Yes, this is what I was proposing. And we can do the same for `fractionalEvaluation` where we can inject the logger and refer implementation where needed (similar to`StringComparisonEvaluator` usage )

Yeah, agree with the point on the follow-up. This will improve the readability & testability.

Kavindu-Dodan (approved):

beeme1mr (approved):

core/pkg/eval/string_comparison_evaluation.go
beeme1mr: https://github.com/open-feature/flagd/issues/667

PR #660, opened 2023-05-17
refactor: introduce isyncstore interface

james-milligan:
## This PR

- decouples the sync store from sync server startup, allowing for alternative sync store implementations to be introduced

### DIFF
--- a/core/pkg/sync-store/interface.go
+++ b/core/pkg/sync-store/interface.go
@@ -1,0 +1,26 @@
+package store
+
+import (
+	"context"
+
+	isync "github.com/open-feature/flagd/core/pkg/sync"
+)
+
+// ISyncStore defines the interface for the sync store
+type ISyncStore interface {
+	FetchAllFlags(
+		ctx context.Context,
+		key interface{},
+		target string,
+	) (isync.DataSync, error)
+	RegisterSubscription(
+		ctx context.Context,
+		target string,
+		key interface{},
+		dataSync chan isync.DataSync,
+		errChan chan error,
+	)
+
+	// metrics hooks
+	GetActiveSubscriptionsInt64() int64
+}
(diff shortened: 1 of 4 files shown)


core/pkg/sync-store/interface.go
```diff
@@ -0,0 +1,26 @@
+package store
+
+import (
+	"context"
+
+	isync "github.com/open-feature/flagd/core/pkg/sync"
+)
+
+// ISyncStore defines the interface for the sync store
+type ISyncStore interface {
```
odubajDT, line 10: PR looks good, nice job! I am just not 100% sure with the naming of the interface `ISyncStore`. I think the `I` prefix comes from Java and is not the best way to use in GO. @bacherfl @thisthat WDYT?

thisthat: the I prefix is used elsewhere in this codebase. See this two examples:
- https://github.com/open-feature/flagd/blob/93e708e979d5e0f454d4131a43787f4f0012f3ab/core/pkg/service/middleware/interface.go#L7
- https://github.com/open-feature/flagd/blob/93e708e979d5e0f454d4131a43787f4f0012f3ab/core/pkg/sync/isync.go#L38

I am not aware of a specific recommendation from the Go language itself. From the standard library, I think I saw mainly `er` as suffix, i.e., this would be `SyncStorer` 
Nevertheless, given the remaining code base use already `I`, I would not mind having the suggested naming

odubajDT (approved):

core/pkg/sync-store/interface.go
james-milligan: Yes, I agree that its not a naming convention associated with Go, but wanted to keep the name in line with the other interfaces in the project

toddbaert: I also don't like the `I` prefix in general, but I think consistency is more important.

If we really feel like getting rid of all the `I` prefixes, it would probably be done in a standalone PR.

toddbaert (approved):

PR #670, opened 2023-05-24
fix: fix connect error code handling for disabled flags

### LINKED ISSUE #668: [BUG] Disabled flags results in an incorrect error message
### Observed behavior

When feature flag is `DISABLED`, flag return an incorrect error message 

`error code: PROVIDER_NOT_READY: connection not made`

### Expected Behavior

The error message should highlight the fact that flag is disabled. This must be corrected at code level [1]

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/flag_evaluator.go#L325C1-L326

### Steps to reproduce

_No response_

Kavindu-Dodan:
## This PR

fixes #668 

Fixes incorrect error code mapping, which resulted in error message `error code: PROVIDER_NOT_READY: connection not made`

With fix, the error code is set to unknown (as there is no appropriate error code for this error) , and results in message `openfeature: GENERAL: unknown: FLAG_DISABLED`

### DIFF
--- a/core/pkg/service/flag-evaluation/flag_evaluator.go
+++ b/core/pkg/service/flag-evaluation/flag_evaluator.go
@@ -322,11 +322,9 @@ func errFormat(err error) error {
 		return connect.NewError(connect.CodeNotFound, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.TypeMismatchErrorCode:
 		return connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
-	case model.FlagDisabledErrorCode:
-		return connect.NewError(connect.CodeUnavailable, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.ParseErrorCode:
 		return connect.NewError(connect.CodeDataLoss, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
-	case model.GeneralErrorCode:
+	case model.FlagDisabledErrorCode, model.GeneralErrorCode:
 		return connect.NewError(connect.CodeUnknown, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	}
 
--- a/core/pkg/service/flag-evaluation/flag_evaluator.go
+++ b/core/pkg/service/flag-evaluation/flag_evaluator.go
@@ -318,13 +318,13 @@ func formatContextKeys(context *structpb.Struct) []string {
 
 func errFormat(err error) error {
 	switch err.Error() {
-	case model.FlagNotFoundErrorCode:
+	case model.FlagNotFoundErrorCode, model.FlagDisabledErrorCode:
 		return connect.NewError(connect.CodeNotFound, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.TypeMismatchErrorCode:
 		return connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.ParseErrorCode:
 		return connect.NewError(connect.CodeDataLoss, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
-	case model.FlagDisabledErrorCode, model.GeneralErrorCode:
+	case model.GeneralErrorCode:
 		return connect.NewError(connect.CodeUnknown, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	}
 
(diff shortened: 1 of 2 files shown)


core/pkg/service/flag-evaluation/flag_evaluator_test.go
```diff
@@ -903,3 +903,50 @@ func getMetricReader() (*telemetry.MetricsRecorder, metric.Reader) {
 	rs := resource.NewWithAttributes("testSchema")
 	return telemetry.NewOTelRecorder(exp, rs, "testSvc"), exp
 }
+
+// TestFlag_Evaluation_ErrorCodes test validate error mapping from known errors to connect.
+// Code and avoid accidental changes. This is essential as SDK implementations rely on connect.
```
beeme1mr, line 908: > Code and avoid accidental changes.
What does that mean?

core/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -322,11 +322,9 @@ func errFormat(err error) error {
 		return connect.NewError(connect.CodeNotFound, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.TypeMismatchErrorCode:
 		return connect.NewError(connect.CodeInvalidArgument, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
-	case model.FlagDisabledErrorCode:
-		return connect.NewError(connect.CodeUnavailable, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
 	case model.ParseErrorCode:
 		return connect.NewError(connect.CodeDataLoss, fmt.Errorf("%s, %s", ErrorPrefix, err.Error()))
-	case model.GeneralErrorCode:
+	case model.FlagDisabledErrorCode, model.GeneralErrorCode:
```
beeme1mr, line 327: You could argue that a disabled flag should behave like it's not there. If that's the case, the `FLAG_NOT_FOUND` error code would be more appropriate. What do you think @toddbaert?

beeme1mr: We could also consider adding a `FLAG_DISABLED` error code but I think that would be a breaking spec change.

Kavindu-Dodan: Good point, I don't have a strong opinion here. I am fine with the current approach (error with FLAG_DISABLED message section) or the proposed change.

toddbaert: Personally I'm of the opinion that flags disabled in flagd should behave EXACTLY as flags that aren't there. Basically they're "unreleased" flags.

core/pkg/service/flag-evaluation/flag_evaluator_test.go
```diff
@@ -903,3 +903,50 @@ func getMetricReader() (*telemetry.MetricsRecorder, metric.Reader) {
...
+// Code to differentiate GRPC errors vs Flag errors. For any change, we must change respective SDK changes
```
toddbaert, line 909:
```suggestion
// TestFlag_Evaluation_ErrorCodes test validate error mapping from known errors to connect.
// Code and avoid accidental changes. This is essential as provider implementations rely on connect.
// Code to differentiate gRPC errors vs Flag errors. Changes must be absorbed by flagd providers.
```

toddbaert: The capitalization here is a bit odd. Maybe a block comment would be better?

Kavindu-Dodan: Fixed with the latest commit. Thank you pointing this out

core/pkg/service/flag-evaluation/flag_evaluator.go
Kavindu-Dodan: Reason will be an error and we can do the following, 

At flagd, mark disabled flags with connect `CodeNotFound` with reason (error message) `FLAG_DISABLED`

This from Go SDK will get converted to following error message, 

`openfeature: FLAG_NOT_FOUND: not_found: FlagdError:, FLAG_DISABLED`

cc @beeme1mr  @toddbaert

Kavindu-Dodan: @toddbaert  thanks and done - updated with the latest commit . Latest error log look like below,

`openfeature: FLAG_NOT_FOUND: not_found: FlagdError:, FLAG_DISABLED`

Kavindu-Dodan:
Latest error message - `openfeature: FLAG_NOT_FOUND: not_found: FlagdError:, FLAG_DISABLED`

toddbaert (approved):

beeme1mr (approved):

PR #675, opened 2023-05-25
Feat/semver json evaluator

bacherfl:
This PR adds additional jsonLogic evaluatorsfor `sem_ver`, which checks if the given property is a valid semantic version string that fulfills a certain criteria (e.g. if the version is lower/greater, equal to or not equal to a certain version) .The evaluator return a boolean, indicating whether the condition has been met.

The following example shows how the evaluators can be used in a flag targeting rule:

```json
{
  "flags": {
    "headerColor": {
      "state": "ENABLED",
      "defaultVariant": "red",
      "variants": {
        "blue": "#0000FF",
        "green": "#00FF00",
        "red": "#FF0000",
        "yellow": "#FFFF00"
      },
      "targeting": {
        "if": [
          {
            "sem_ver": [
              {
                "var": "version"
              },
              ">",
              "1.0.0"
            ]
          },
          "red",
          "green"
        ]
      },
      "source": ""
    }
  }
}
```

This rule will resolve to the `red` variant of the `headerColor` flag, if the `version` property within the evaluationContext is a semantic version string that is greater than `1.0.0`.

### DIFF
--- a/core/pkg/eval/semver_evaluation_test.go
+++ b/core/pkg/eval/semver_evaluation_test.go
@@ -1,0 +1,482 @@
+package eval
+
+import (
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/model"
+	"github.com/open-feature/flagd/core/pkg/store"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/types/known/structpb"
+	"testing"
+)
+
+func TestSemVerOperator_Compare(t *testing.T) {
+	type args struct {
+		v1 string
+		v2 string
+	}
+	tests := []struct {
+		name    string
+		svo     SemVerOperator
+		args    args
+		want    bool
+		wantErr bool
+	}{
+		{
+			name: "invalid operator",
+			svo:  "",
+			args: args{
+				v1: "v0.0.1",
+				v2: "v0.0.2",
+			},
+			want:    false,
+			wantErr: true,
+		},
+		{
+			name: "less",
+			svo:  Less,
+			args: args{
+				v1: "v0.0.1",
+				v2: "v0.0.2",
+			},
+			want:    true,
+			wantErr: false,
+		},
+		{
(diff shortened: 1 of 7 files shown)


beeme1mr (approved):

docs/configuration/sem_ver_evaluation.md
```diff
@@ -0,0 +1,86 @@
...
+The `sem_ver` evaluation checks if the given property matches a semantic versioning condition.
+It returns 'true', if the value of the given property meets the condition, 'false' if not.
+
+## SemVer Evaluation Configuration
+
+The `sem_ver` evaluation can be added as part of a targeting definition.
+Note that the 'sem_ver' evaluation rule must contain exactly three items:
+
+1. Target property: this needs which both resolve to a semantic versioning string
+2. Operator: One of the following: `=`, `!=`, `>`, `<`, `>=`, `<=`, `~`, `^`
```
toddbaert, line 18:
```suggestion
2. Operator: One of the following: `=`, `!=`, `>`, `<`, `>=`, `<=`, `~` (match minor version), `^` (match major version)
```

toddbaert (approved):
Looks great to me! I really like the ~ and ^ additions, thanks. I think semver is a common-enough use case that this is easily justified, many vendors support it in rules.

core/pkg/eval/semver_evaluation_test.go
```diff
@@ -0,0 +1,622 @@
...
+			context: &structpb.Struct{Fields: map[string]*structpb.Value{
+				"email": {Kind: &structpb.Value_StringValue{
+					StringValue: "user@faas.com",
+				}},
+			}},
+			expectedVariant: "green",
+			expectedValue:   "#00FF00",
+			expectedReason:  model.TargetingMatchReason,
+		},
+	}
```
thisthat, line 593: I would add an entry to check for the correct handling of `~` and `^` to protect against regressions

thisthat (approved):
LGTM 👍

PR #681, opened 2023-05-30
docs: extended the flag configuration documentation

beeme1mr:
# This PR

- Includes a description for the full flag configuration
- Expands on the targeting rule conditions, properties, and functions.

### Related Issues

Fixes #680

PR #684, opened 2023-06-01
chore: refactor json logic evaluator to pass custom operators as options

### LINKED ISSUE #667: Refactor JSONEvaluator
I just committed my suggestion - what I don't particularly like about it is that the `StringComparisonEvaluator` is initialized within the `NewJSONEvaluator` function - here it maybe would make sense to do a follow up refactoring where the individual evaluators (i.e. starts/ends_with, fractional can be added as options to the `NewJSONEvaluator` function, to further decouple the `JSONEvaluator` from the other components by not needing it to be aware of all custom evaluators itself. what do you think?

_Originally posted by @bacherfl in https://github.com/open-feature/flagd/pull/658#discussion_r1201894000_

bacherfl:
Closes #667

This PR moves the registration of the custom JSON evaluators out of the `NewJSONEvaluator` method, and allows to pass the custom evaluators as options to the `New` function. This way, the `JSONEvaluator` does not need to be aware of the custom evaluations and reduces its responsibility to just providing the means to execute jsonLogic evaluations, based on its store and the flag configurations.


core/pkg/eval/json_evaluator.go
```diff
@@ -49,7 +49,17 @@ const (
 	Disabled = "DISABLED"
 )
 
-func NewJSONEvaluator(logger *logger.Logger, s *store.Flags) *JSONEvaluator {
+type JSONEvaluatorOption func(je *JSONEvaluator)
+
+func WithEvaluator(aliases []string, evalFunc func(interface{}, interface{}) interface{}) JSONEvaluatorOption {
```
thisthat, line 54: This is a slight behavioral change since previously we only had a single alias for an operator.
```suggestion
func WithEvaluator(aliases string, evalFunc func(interface{}, interface{}) interface{}) JSONEvaluatorOption {
```

bacherfl: the possibility for providing a list of aliases was done as preparation for introducing an alias for the `fractionalEvaluation` (see https://github.com/open-feature/flagd/pull/658#discussion_r1196717542), but we can also introduce this later when we work on that specific change

bacherfl: as discussed, changed it to `string` instead of `[]string`

beeme1mr: Should I create a follow-up for adding an alias? It would be worth it since `fractionalEvaluation` has a different casing than the others. It looks a bit out of place in the docs I was working on.

https://github.com/open-feature/flagd/pull/681

core/pkg/runtime/from_config.go
```diff
@@ -140,6 +141,30 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
...
+			eval.NewStringComparisonEvaluator(logger).EndsWithEvaluation,
+		),
+		eval.WithEvaluator(
+			"sem_ver",
+			eval.NewSemVerComparisonEvaluator(logger).SemVerEvaluation,
+		),
+	)
+	return evaluator
+}
+
```
toddbaert, line 167: I like this much better. Thanks

toddbaert (approved):

toddbaert:
FYI I think the lint error that's appeared is a bug/false positive.

Kavindu-Dodan (approved):

beeme1mr (approved):

Kavindu-Dodan:
Unfortunately cannot reopen the PR. This is discussed here : https://github.com/isaacs/github/issues/361 

let's open a new PR to get this through

beeme1mr:
Yeah, my bad. I thought closing and reopening the PR would fix the GitHub action issue.

PR #731, opened 2023-07-05
Feat/zero downtime

### LINKED ISSUE #728: [FEATURE] Support graceful shutdown
## Goal

FlagD shall be upgraded without any downtime and run with multiple replicas.

### Details

- FlagD should be able to handle upgrades without any downtime.
For this, we should make sure that it handles the [pre-stop signals of K8s](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/).
- When the SIGTERM and SIGINT signals are sent to the container, FlagD should close any open connection with all connected SDKs.
- SDKs should try to reconnect again and should hit a ready replica of FlagD.
- Furthermore, FlagD should run as PID 1 in the container, otherwise it won't receive the signals emitted by K8s.

### DoD

- FlagD runs as PID1
- FlagD handles termination gracefully
- SDK tries to reconnect if a FlagD is upgraded
- Readiness probe is "disabled" as soon as the termination signal is received
- Tests to cover this behavior are provided

odubajDT:
## This PR

- implements graceful shutdown of flagD, which leads to zero-downtime

### Related Issues

Fixes #728 

### Follow-up Tasks
 - zero downtime test will be part of a follow up PR


config/deployments/flagd/deployment.yaml
```diff
@@ -0,0 +1,44 @@
...
+            path: /healthz
+            port: 8014
+          initialDelaySeconds: 5
+          periodSeconds: 60
+        ports:
+          - containerPort: 8013
+        args:
+          - start
+          - --uri
+          - https://raw.githubusercontent.com/open-feature/flagd/main/samples/example_flags.flagd.json
```
Kavindu-Dodan, line 43: Nit - shall we avoid this reference to flags and replace it with volume mounted file path reference [1]

ex:

```
          args: [ "start", "--uri", "file:/etc/flagd/config.json" ]
          volumeMounts:
            - name: config-volume
              mountPath: /etc/flagd
      volumes:
        - name: config-volume
          configMap:
            name: open-feature-flags
            items:
              - key: flags
                path: config.json
                
---

# ConfigMap for Flagd OpenFeatuer provider
apiVersion: v1
kind: ConfigMap
metadata:
  name: open-feature-flags
data:
  flags: |
{{Flag JSON }}
```

The reason is to decouple tests with a resource living in the main. Given that test's evaluation rely on flag content, we should avoid circular dependency on repo main 

[1] - https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md#filepath-provider

Kavindu-Dodan:
Nice work 👏 

There's one more aspect we need to consider. Flagd maintains a grpc backed eventing mechanism between its SDK components [1] [2]. 

My suggestion is to introduce a shutdown message to eventing through this PR. Later on, each SDK implementation can follow up on acting on this new event.

Proposed event - `provider_shutdown`
Action - The event listener at SDK should shutdown and trigger a connection loop with back offs 

cc - @thisthat & @toddbaert 

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/eventing.go
[2] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L542

config/deployments/flagd/deployment.yaml
odubajDT: Also a posibility, will adapt!

odubajDT:
> Nice work clap
> 
> There's one more aspect we need to consider. Flagd maintains a grpc backed eventing mechanism between its SDK components [1] [2].
> 
> My suggestion is to introduce a shutdown message to eventing through this PR. Later on, each SDK implementation can follow up on acting on this new event.
> 
> Proposed event - `provider_shutdown` Action - The event listener at SDK should shutdown and trigger a connection loop with back offs
> 
> cc - @thisthat & @toddbaert
> 
> [1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/eventing.go [2] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L542

Could be a nice. Maybe a small discussion in the team would be good, so the whole team would align on the same solution.

core/pkg/runtime/runtime.go
```diff
@@ -37,8 +37,7 @@ func (r *Runtime) Start() error {
 	if r.Evaluator == nil {
 		return errors.New("no evaluator set")
 	}
-	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
-	defer cancel()
```
thisthat, line 41: please, keep the cancel function since some function down the line could listen to the `done()` event

thisthat:
> > Nice work clap
> > There's one more aspect we need to consider. Flagd maintains a grpc backed eventing mechanism between its SDK components [1] [2].
> > My suggestion is to introduce a shutdown message to eventing through this PR. Later on, each SDK implementation can follow up on acting on this new event.
> > Proposed event - `provider_shutdown` Action - The event listener at SDK should shutdown and trigger a connection loop with back offs
> > cc - @thisthat & @toddbaert
> > [1] - [`main`/core/pkg/service/flag-evaluation/eventing.go](https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/eventing.go?rgh-link-date=2023-07-06T05%3A21%3A00Z) [2] - [open-feature/go-sdk-contrib@`main`/providers/flagd/pkg/provider.go#L542](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go?rgh-link-date=2023-07-06T05%3A21%3A00Z#L542)
> 
> Could be a nice. Maybe a small discussion in the team would be good, so the whole team would align on the same solution.
> 
> But AFAIK, grpc sends a EOF message defaultly when closing the stream/connection. Therefore SDKs can react on this message and just re-connect. We used this approach when implementing ZD for flagD-proxy, where when a replica of flagD-proxy is terminated, is sends EOF, where the client (for example flagD) reacts on this by re-connecting to the newly created replica. Therefore there is maybe no reason to create a custom event when shutting down, as the client should just try to re-connect, when the connection is closed. For the client it's not important that it looses the connection, but that it can immediately re-connect to other replica

@odubajDT mind that a service connects to FlagD via HTTP. See the provider implementation for Go [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/service/client.go#L77) as a reference. 

I support the idea of @Kavindu-Dodan 👍

toddbaert:
> @odubajDT mind that a service connects to FlagD via HTTP. See the provider implementation for Go [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/service/client.go#L77) as a reference.

Just to avoid any confusion here, the flagd go provider doesn't use "raw" HTTP - it uses the _Connect protocol_ which is a protocol for gRPC interop _over_ HTTP.

In fact, we could modify the section of code referenced by @thisthat to use raw gRPC:

```
c.client = schemaConnectV1.NewServiceClient(
			&http.Client{
				Transport: &http.Transport{
					TLSClientConfig: tlsConfig,
					DialContext:     dialContext,
				},
			},
			url,
			connect.WithGRPC(),  // I added this this would use raw gRPC
		)
```

TbH, I'm not sure why we are using connect in this case (maybe @Kavindu-Dodan or @james-milligan know, maybe there's a good reason).

test/zero-downtime/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This will create a flagd deployment `flagd-dev` namespace.
+
+To run the test, execute:
+
+```shell
+IMG=your-flagd-image IMG_ZD=your-flagd-image2 make run-zd-test
+```
+
+Please be aware, you need to build your two custom images with differetn tags for flagD first.
```
bacherfl, line 19:
```suggestion
Please be aware, you need to build your two custom images with different tags for flagD first.
```

Kavindu-Dodan:
> > @odubajDT mind that a service connects to FlagD via HTTP. See the provider implementation for Go [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/service/client.go#L77) as a reference.
> 
> Just to avoid any confusion here, the flagd go provider doesn't use "raw" HTTP - it uses the _Connect protocol_ which is a protocol for gRPC interop _over_ HTTP.
> 
> In fact, we could modify the section of code referenced by @thisthat to use raw gRPC:
> 
> ```
> c.client = schemaConnectV1.NewServiceClient(
> 			&http.Client{
> 				Transport: &http.Transport{
> 					TLSClientConfig: tlsConfig,
> 					DialContext:     dialContext,
> 				},
> 			},
> 			url,
> 			connect.WithGRPC(),  // I added this; it would use raw gRPC instead of Connect
> 		)
> ```
> 
> TbH, I'm not sure why we are using connect in this case (maybe @Kavindu-Dodan or @james-milligan know, maybe there's a good reason). I also don't know if this has an impact on the discussion above.

@toddbaert 

I think this is slightly connected. Using connect protocol is not the same as grpc 🤔  so we might need to consider adding grpc connectivity if we want to rely on grpc options (ex:- timeouts)

@odubajDT 

The reason I am proposing a shutdown message is to avoid any eventing related errors/logs as sdk side. For example, I have seen the following intermittent error in Java SDK when using flagD

> ERROR [dev.openfeature.contrib.providers.flagd.EventStreamObserver] event stream io.grpc.StatusRuntimeException: UNAVAILABLE: Network closed for unknown reason

And I am not sure about the claim that grpc connection. The stream can have a connection timeout [1], which allows a proper connection termination based on grpc internals. But this is/might not true for server shutdown as there is no trigger for such shutdown. And this is where my concern comes from. 

Besides, shutdown is one part, which allows sdk logic to build/refactor connection retry attempts. And later on we can enforce connection timeouts for connection load balancing (avoid long-lived streams and allow load balancing) based on the same logic. 

[1] - https://pkg.go.dev/google.golang.org/grpc/keepalive#ServerParameters

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -152,6 +155,14 @@ func (s *ConnectService) AddMiddleware(mw middleware.IMiddleware) {
 	s.server.Handler = mw.Handler(s.server.Handler)
 }
 
+func (s *ConnectService) Shutdown() {
+	s.readinessEnabled = false
+	s.eventingConfiguration.emitToAll(service.Notification{
```
Kavindu-Dodan, line 160: +1 for the shutdown event so that eventing subscriptions can react

test/zero-downtime/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This will create a flagd deployment `flagd-dev` namespace.
+
+To run the test, execute:
+
+```shell
+IMG=your-flagd-image IMG_ZD=your-flagd-image2 make run-zd-test
+```
+
+Please be aware, you need to build your two custom images with different tags for flagD first.
```
Kavindu-Dodan, line 19: Just for my understanding, is the two image requirement (two tags) come from K8s (ex:- RollingUpdate config) ? Or can this be tests with one image

Kavindu-Dodan (approved):
LGTM

test/zero-downtime/README.md
odubajDT: It can be the same image, but needs to have 2 different tags, so that the Rollingupdate of the Deployment is actually executed. If we would use only one image with one tag, no update would be executed (comes from k8s behavior) and therefore the test results are not valid

odubajDT:
> LGTM
> 
> Sidenote - ZD support for flagd-proxy [1] is also a good follow-up task if you are interested
> 
> [1] - https://github.com/open-feature/flagd/tree/main/flagd-proxy

I can look at it thank you :) But first I wanted to make the ZD test part of the pipeline https://github.com/open-feature/flagd/issues/732

toddbaert:
> Nice work clap
> 
> There's one more aspect we need to consider. Flagd maintains a grpc backed eventing mechanism between its SDK components [1] [2].
> 
> My suggestion is to introduce a shutdown message to eventing through this PR. Later on, each SDK implementation can follow up on acting on this new event.
> 
> Proposed event - `provider_shutdown` Action - The event listener at SDK should shutdown and trigger a connection loop with back offs
> 
> cc - @thisthat & @toddbaert
> 
> [1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/eventing.go [2] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L542

I think we do need to add a shutdown event to the actual spec/SDKs eventually, @Kavindu-Dodan . I think the use-cases are narrow but it's probably important. I'm not sure if you meant this, or the flagd providers themselves.

toddbaert (approved):
LGTM. Is it worth it to automate this test in a kind cluster as part of the CI? I know we have unit tests, which may be enough.

I'll let others decide. Either way, we can create another issue/PR for that if we want.

odubajDT:
> > Nice work clap
> > There's one more aspect we need to consider. Flagd maintains a grpc backed eventing mechanism between its SDK components [1] [2].
> > My suggestion is to introduce a shutdown message to eventing through this PR. Later on, each SDK implementation can follow up on acting on this new event.
> > Proposed event - `provider_shutdown` Action - The event listener at SDK should shutdown and trigger a connection loop with back offs
> > cc - @thisthat & @toddbaert
> > [1] - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/eventing.go [2] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L542
> 
> I think we do need to add a shutdown event to the actual spec/SDKs eventually, @Kavindu-Dodan . I think the use-cases are narrow but it's probably important. I'm not sure if you meant this, or the flagd providers themselves.

I think the SDKs need to react on this `shutdown` event properly (this part is not implemented yet). But I guess that should be a follow-up ticket for each SDK, or ?

odubajDT:
> LGTM. Is it worth it to automate this test in a kind cluster as part of the CI? I know we have unit tests, which may be enough.
> 
> I'll let others decide. Either way, we can create another issue/PR for that if we want.

We already have an issue for that https://github.com/open-feature/flagd/issues/732 but it's true it was not discussed in the team, if it's needed. Maybe not is the right time to look into this. @thisthat @bacherfl @RealAnna @Kavindu-Dodan @mowies WDYT?

beeme1mr (approved):

PR #761, opened 2023-07-21
feat: grpc selector as scope

Kavindu-Dodan:
## This PR

Introduce flag evaluation metadata to flagd.

For this initial change, metadata named "scope" is added. This maps to flagd gRPC source's `selector` [^1]

Change process:

- schema update  [^2] & release [^3]
- json flag store was updated to store source details (the current detail being the selector)
- updated flag resolvers to return metadata structure
- lots of reference changes 

end to end result:

See below Prometheus screenshot for scope attribute, derived from metadata 

image

[^1]: https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md#source-configuration 
[^2]: https://github.com/open-feature/schemas/pull/97
[^3]: https://github.com/open-feature/schemas/pull/98


beeme1mr (approved):
Looks good, thanks.

core/pkg/eval/json_evaluator.go
```diff
@@ -243,54 +245,55 @@ func (je *JSONEvaluator) ResolveObjectValue(
...
+	variant string, variants map[string]interface{}, reason string, metadata map[string]interface{}, err error,
+) {
+	metadata = map[string]interface{}{}
+
 	flag, ok := je.store.Get(flagKey)
 	if !ok {
 		// flag not found
 		je.Logger.DebugWithID(reqID, fmt.Sprintf("requested flag could not be found: %s", flagKey))
-		return "", model.ErrorReason, errors.New(model.FlagNotFoundErrorCode)
+		return "", flag.Variants, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
```
thisthat, line 285: I think this won't work. since `ok` is `false` `flag` is empty

toddbaert: Ya I think that's certainly true.

toddbaert, line 285:
```suggestion
		return "", map[string]interface{}{}, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
```

toddbaert:
@thisthat I replaced the `Variants` with an equivalent zero value.

As for the conflict, not matter how I resolve it, your recent test `TestErrorIntercepted` seems to fail. I can't quite make sense of it, these changes don't seem related to yours :sweat_smile:

toddbaert:
> @thisthat I replaced the `Variants` with an equivalent zero value.
> 
> As for the conflict, not matter how I resolve it, your recent test `TestErrorIntercepted` seems to fail. I can't quite make sense of it, these changes don't seem related to yours sweat_smile

I resolved this, I think it was a not-related race.

toddbaert (approved):
I addressed a few issues pointed out by others, and fixed an unrelated race condition.

@beeme1mr could you please do one last manual check? Then we can merge.

PR #782, opened 2023-08-01
docs: initial docs commit

agardnerIT:
## This PR
- Adds docs in the `/web-docs` folder for the new flagd website
- Adds a GitHub action to build the docs using mkdocs and store the static files into the /build directory

The docs use mkdocs and the readthedocs theme. The intention is to structure into a happy path:

1. All new users will first read the concepts section which explains feature flags and flagd
1. Users then "choose their journey" and start down either the kubernetes path or the non kubernetes path.

--------------

## web-docs vs. docs

I believe the GitHub `docs` folder and the `web-docs` serve different audiences. The former being "more technical" and the latter being "less technical". The language used, technical depth to go to and topics are all entirely different.

With that in mind, I believe it makes sense to keep the two separate.

Simply porting over the `docs` content to the website would not be wise and would lead to a bad user experience and potentially impact adoption.

It is also to be expected that the public facing website will take time to "get right" in terms of content, navigation and feel.

This also:

- Allows flagd developers to document things in `docs` alongside the code that are fast moving, not yet concrete or are in a state of change.
- Doesn't break backwards compatibility - if someone visits the repo they can still navigate.
- DRY principle (don't repeat yourself): Developers write the docs once, in the appropriate place.

------------

### Great, but that's now two doc stacks

What this **DOES NOT** lead to is double documentation / two places / split brain docs:

- If the content is suitable for the website, it is documented ONCE in `web-docs`. A placeholder page inside `docs` MAY be created but ONLY as a pointer to the "proper" website documentation.
- If the content is suitable for the `docs` folder (ie. it discusses a highly technical topic that only a subset of audience would reasonably find appropriate) then it is documented ONCE in `docs`. A placeholder page in `web-docs` MAY be created by ONLY as a pointer to the "full" readme on GitHub.

I realise that this PR violates this rule because I can't edit the `docs` files to point to the website (yet). Once the website is live, issues will be raised and will make excellent and easy `Good First Issues` for new contributors as you're effectively asking people to remove content and point instead to a URL. Thus helping to bring us in line with this rule and **encouraging first time flagd contributors** and **building community**.

### Why not just "live retrieve" from GitHub?

This model would see the documentation website built from the "source" material on GitHub. Hopefully, the above explanation regarding different audiences explains why (I believe) this to be a bad idea.

If there is ever a need to get something live that quickly, a signpost page in `web-docs` (temporarily) pointing people to the GitHub markdown would suffice until we can

------

## Locally serving the docs (for development)
You'll need Python installed then `pip install -r requirements.txt` to install `mkdocs`.

Then from the root: `mkdocs serve -a localhost:8000`. Any changes you make to the markdown files will be automatically updated on the live site.

--------------

### Ownership and Maintenance Statement

I hereby state that I'm happy to be considered a "docs maintainer" to keep this content up to date and relevant.


web-docs/index.md
```diff
@@ -0,0 +1,30 @@
+# flagd
+
+Flagd is a feature flag daemon. It is a ready-made, open source, [OpenFeature](https://openfeature.dev) compliant feature flag backend system.
+
+- OpenFeature compliant and speaks your language.
+- Easy to extend to new languages.
+- Supports multiple data sources simultaneously.
+- Feature Flag updates occur in near real-time.
+- Contains a powerful and flexible rule targeting engine and deterministic percentage-based rollouts.
+- Flag evaluation statistics and metrics are exposed and compatible with Prometheus.
```
Kavindu-Dodan, line 10: IMO OpenTelemetry gives more prominence here 

```suggestion
- Flag evaluation statistics and metrics are exposed and compatible with OpenTelemetry.
```

web-docs/index.md
```diff
@@ -0,0 +1,30 @@
...
+Start your flagd learning journey here: [flagd concepts](concepts)
+
+## Running flagd
+
+### Kubernetes or Non-Kubernetes?
+
+Now that you know the concepts, it's time to decide how to run flagd.
+
+flagd can run on Kubernetes and in non-k8s environments.
+Choose your operating environment to learn more:
```
Kavindu-Dodan, line 27:
```suggestion
Choose your operating mode to learn more:
```

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
+# flagd basics
+
+Your flagd journey will start by defining your feature flags.
+
+flagd will then read those feature flags and make them available to your application.
+
+Your application will interact with flagd via the [OpenFeature SDK](https://example.com) to retrieve flag values via the flagd API.
+
+![flagd architecture](../images/flagd-logical-architecture.jpg)
```
Kavindu-Dodan, line 9: minor - do we need the architecture diagram again here?

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
...
+## Targeting rules
+
+Imagine you are introducing a new feature. You create a flag with two possible variants: `on` and `off`. You want to safely roll out the feature.
+Therefore the flags `defaultValue` is set to `off` for all users.
+
+In other words, the new feature is disabled by default.
+
+Now imagine you want to enable the feature, but only when both of these conditions are true:
+
+- Logged in users
```
Kavindu-Dodan, line 82:
```suggestion
- User is logged in
```

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
...
+- The user email ends in `@example.com`
```
Kavindu-Dodan, line 83:
```suggestion
- The user's email ends in `@example.com`
```

web-docs/concepts/index.md
Kavindu-Dodan, line 82: The targeting rule does not cover this scenario and only validates the email sent in the context

IMO, we can reword the example to give the idea of "enabling new feature to logged in users who have their emails ending with a specific domain - @example.com"

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
...
+            }
+         }
+    }
+}
+```
+
+### Pseudo-code of application passing context
+
+```js
+// The second parameter is the default in case flagd is unavailable
```
Kavindu-Dodan, line 117:
```suggestion
// Here, we provide an empty context, hence the flag evaluates to false value which is the defaultVariant
```

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
...
+featureAvailable = openFeature.getBooleanValue("isFeatureEnabled", false, {}) // false
+
+// isFeatureEnabled for a logged in user with an email example@gmail.com
```
Kavindu-Dodan, line 120:
```suggestion
// Here, we provide email for the flag evaluation. Still flag evaluates to defaultVariant of false as email does not end with desired domain
```

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
...
+featureAvailable = openFeature.getBooleanValue("isFeatureEnabled", false, {"email": "example@gmail.com"}) // false
+
+// isFeatureEnabled for a logged in user with an email someone@example.com
```
Kavindu-Dodan, line 123: // Here, the flag is evaluated with targeting rule matching, hence the value of true

beeme1mr:
I'll review what you have and wire it up to netlify as soon as possible.

web-docs/concepts/index.md
Kavindu-Dodan, line 123:
```suggestion
// Here, the flag is evaluated with targeting rule matching, hence the value of true
```

web-docs/concepts/index.md
```diff
@@ -0,0 +1,190 @@
+# flagd basics
+
+Your flagd journey will start by defining your feature flags.
+
+flagd will then read those feature flags and make them available to your application.
+
+Your application will interact with flagd via the [OpenFeature SDK](https://example.com) to retrieve flag values via the flagd API.
```
Kavindu-Dodan, line 7: Need to fix the link

Kavindu-Dodan:
PR build fails and this is due to markdown linitng 

Please run `make markdownlint` to list markdown errors & use `make markdownlint-fix` to auto fix some of them

web-docs/concepts/index.md
beeme1mr: Technically, the user would need to be logged for the email matching to work but I agree the phrasing makes it sound like the targeting rule is checking both criteria specifically.

web-docs/concepts/index.md
agardnerIT: What is the most appropriate link for this?

web-docs/concepts/index.md
agardnerIT: I think so. People may land on this page instead of the homepage.

toddbaert:
> I hereby state that I'm happy to be considered a "docs maintainer" to keep this content up to date and relevant.

We could consider using [component_owners](https://github.com/dyladan/component-owners) to help with this.

agardnerIT:
I believe I've fixed all markdown issues and this failure is something different. It looks like a performance test which is only just being violated.

What does this performance test, test? If it is response time, perhaps compressing the images would help?

Side note: We need to clearly document what all is going on in these PR actions.

web-docs/concepts/opentelemetry.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Metrics
+
+flagd expose following metrics,
+
+* `http_request_duration_seconds`
+* `http_response_size_bytes`
+* `http_requests_inflight`
+* `impressions`
+* `reasons`
```
beeme1mr, line 20:
```suggestion
* `http.server.duration`
* `http.server.response.size`
* `http.server.active_requests`
* `feature_flag.flagd.impression`
* `feature_flag.flagd.evaluation.reason`
```
I've updated the original docs also to reflect this change.

https://github.com/open-feature/flagd/edit/main/docs/configuration/flagd_telemetry.md

beeme1mr (commented):
I'll give this a more thorough review tomorrow but it looks great from my initial scan. Thanks for putting this together. Better documentation is sorely needed.

beeme1mr (approved):

PR #789, opened 2023-08-02
fix: metric descriptions match the otel spec

beeme1mr:
## This PR

- updates the metric description to match the OTel spec

### Notes

The metric descriptions must match what's defined in the [OTel spe ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#metric-httpserverduration) or the collector will complain.

Signed-off-by: Michael Beemer <michael.beemer@dynatrace.com>

### DIFF
--- a/core/pkg/service/flag-evaluation/flag_evaluator_test.go
+++ b/core/pkg/service/flag-evaluation/flag_evaluator_test.go
@@ -711,11 +711,8 @@ func BenchmarkFlag_Evaluation_ResolveInt(b *testing.B) {
 	tests := map[string]resolveIntArgs{
 		"happy path": {
 			evalFields: resolveIntEvalFields{
-				result: 12,
-				evalCommons: evalCommons{
-					variant: "on",
-					reason:  model.DefaultReason,
-				},
+				result:      12,
+				evalCommons: happyCommon,
 			},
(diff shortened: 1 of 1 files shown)


core/pkg/telemetry/metrics.go
```diff
@@ -131,27 +131,27 @@ func NewOTelRecorder(exporter msdk.Reader, resource *resource.Resource, serviceN
 	// we can ignore errors from OpenTelemetry since they could occur if we select the wrong aggregator
 	hduration, _ := meter.Float64Histogram(
 		httpRequestDurationMetric,
-		metric.WithDescription("The latency of the HTTP requests"),
+		metric.WithDescription("Measures the duration of inbound HTTP requests."),
```
toddbaert, line 134: no change these are exported from the semconv package, is there?

toddbaert (approved):

core/pkg/telemetry/metrics.go
Kavindu-Dodan: Seems not 🤔  there's only attribute key value exposed

Kavindu-Dodan (approved):

Issue #787, opened 2023-08-02
[BUG] Missing value with ResolveBoolean when variant is off 


L3o-pold:
### Observed behavior

Using `flagd:v0.6.2` the `/schema.v1.Service/ResolveBoolean` route respond without a `value` field. The PHP SDK then think it's an [error](https://github.com/open-feature/php-sdk-contrib/blob/main/providers/Flagd/src/http/FlagdResponseValidator.php#L24).

### Expected Behavior

I guess `value` field is [mandatory](https://github.com/open-feature/flagd/blob/main/docs/usage/evaluation_examples.md#resolve-a-boolean-value-with-evaluation-context). 

### Steps to reproduce

Use this kind of config file.

```json
{
    "flags": {
        "test": {
            "state": "ENABLED",
            "variants": {
                "on": true,
                "off": false
            },
            "defaultVariant": "on",
            "targeting": {
                "if": [
                    {
                        "$ref": "ipWithoutTest"
                    },
                    "off",
                    "on"
                ]
            }
        }
    },
    "$evaluators": {
        "ipWithoutTest": {
            "in": [
                {
                    "var": [
                        "ip"
                    ]
                },
                [
                    "10.0.0.1",
                    "10.0.0.2"
                ]
            ]
        }
    }
}
```

Then query with an IP present in the evaluator:

```shell
curl --location 'http://localhost:8003/schema.v1.Service/ResolveBoolean' \
--header 'Content-Type: application/json' \
--data '{
    "flagKey": "test",
    "context": {
        "ip": "10.0.0.1"
    }
}'
```

In the response the `value` field is missing

```json
{
    "reason": "TARGETING_MATCH",
    "variant": "off",
    "metadata": {}
}
```

If you do the same query with an IP that's not matching the evaluator rule, the `value` attribute is present.

```json
{
    "value": true,
    "reason": "TARGETING_MATCH",
    "variant": "on",
    "metadata": {}
}
```

beeme1mr:
Hey @L3o-pold, sorry you ran into this issue. This is the default behavior of proto3 JSON encoded output. We've documented the behavior [here](https://github.com/open-feature/flagd/blob/main/docs/help/omitted_value_from_response.md).

It looks like it may be possible to override this default behavior.

https://protobuf.dev/programming-guides/proto3/#json-options

I think this may be worth exploring. FYI @toddbaert @Kavindu-Dodan

L3o-pold:
[...]

PR #792, opened 2023-08-03
feat: include falsy json fields

toddbaert:
- adds custom JSON config which returns all falsy json values in responses
- adds associated tests
- removes now irrelevant doc

Responses that used to look like these:

`{"reason":"DEFAULT", "variant":"red", "metadata":{}}`
`{"reason":"STATIC", "variant":"off", "metadata":{}}`

Now look like these:

`{"value":0, "reason":"DEFAULT", "variant":"red", "metadata":{}}`
`{"value":false, "reason":"STATIC", "variant":"off", "metadata":{}}`

Fixes: https://github.com/open-feature/flagd/issues/787


core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -119,7 +120,14 @@ func (s *ConnectService) setupServer(svcConf service.Configuration) (net.Listene
 		s.eventingConfiguration,
 		s.metrics,
 	)
-	path, handler := schemaConnectV1.NewServiceHandler(fes, svcConf.Options...)
+
+	marshalOpts := WithJSON(
+		// json parsing configuration - we emit "unpopulated" fields (falsy fields are not dropped)
+		protojson.MarshalOptions{EmitUnpopulated: true},
+		protojson.UnmarshalOptions{DiscardUnknown: true},
```
beeme1mr, line 127: Does this change the unmarshaling behavior from what we had?

beeme1mr (approved):
Awesome, great job.

core/pkg/service/flag-evaluation/connect_service.go
toddbaert: Only in the way we want. `DiscardUnknown` was the previous behavior, see [here](https://github.com/bufbuild/connect-go/blob/main/codec.go#L176).

core/pkg/service/flag-evaluation/json_codec.go
```diff
@@ -0,0 +1,95 @@
...
+		connect.WithCodec(&jsonCodec{name: "json; charset=utf-8", marshal: marshal, unmarshal: unmarshal}),
+	)
+}
+
+type jsonCodec struct {
+	name      string
+	marshal   protojson.MarshalOptions
+	unmarshal protojson.UnmarshalOptions
+}
+
```
Kavindu-Dodan, line 32:
```suggestion
var _ connect.Codec = (*jsonCodec)(nil)
```

This is a compile time check to see if `jsonCodec` implements `connect.Codec` interface

Kavindu-Dodan (approved):
Nice work 🎉

core/pkg/service/flag-evaluation/json_codec.go
Kavindu-Dodan: Connect internals also do the same - https://github.com/bufbuild/connect-go/blob/main/codec.go#L146C1-L146C37

core/pkg/service/flag-evaluation/json_codec_test.go
```diff
@@ -0,0 +1,139 @@
...
+	for name, tt := range tests {
+		t.Run(name, func(t *testing.T) {
+			jsonMap := map[string]interface{}{}
+			message, err := tt.message()
+			if err != nil {
+				t.Errorf("Got error creating message: %v", err.Error())
+			}
+			bytes, err := tt.marshaller(message)
+			if tt.wantErr && err == nil {
+				t.Errorf("Expected error but not none")
```
bacherfl, line 104: optional suggestion: To keep consistent with the other assertions, you could also use `require.NotNil(t, err)`, or `require.NotNilf(t, err, "Expected error but not none")`

bacherfl (approved):
left one optional suggestion, but LGTM otherwise!

core/pkg/service/flag-evaluation/json_codec_test.go
toddbaert: ty!

toddbaert: (I meant to type "got none" also)

Issue #843, opened 2023-08-16
Improvements to fractional custom evalutor


toddbaert:
The fractional evaluation currently uses xxh3 algo as part of it's bucketing algorithm (see [here](https://github.com/open-feature/flagd/blob/main/core/pkg/eval/fractional_evaluation.go)). In order to support [in-memory implementations](https://github.com/open-feature/flagd/issues/842) in other languages, we should use a more popular algorithm, such as `murmur3` (this is popular with other feature flag SDKs as well). Additionally, the algorithm doesn't currently take into account the flag key. This means that, for different flags, bucketing will have a predictable ordinality. The algo should be updated to use both the selected context property AND the flag key as factors to the hash. We also should update the evaluator to be able to use nested context props (not just top level props).

Definition of done:

- [x] fractional evaluator updated to use murmur3 or md5, or other more popular hash than xxh3 (take into account support in various other langues, such as Java, JS, etc)
- [x] flag key added to context and used as input to hash (implemented https://github.com/open-feature/flagd/pull/847#pullrequestreview-1590024606)
- [ ] ability to use nested props as key
  - see: https://github.com/open-feature/flagd/pull/848#discussion_r1301661915 
    - allow the use a JSON logic expression to get the bucket value (ie: `{"var":"some.nested.prop"}`)
    - if no value is provided, use `{"cat": [{"var":"$flagd.flagKey"}, {"var":"targetingKey"}]}` (basically what we have now, but expressed as a JSON logic expression)

craigpastro:
👋  @toddbaert. I took a quick look at this. The first step is easy. For part 2, flag key also used as input to hash, one way I imagine this could be accomplished is by adding the flag key to the context/data when calling the functional evaluator. Is that what you were thinking too? Happy to make a PR for this if so.

I didn't understand what step 3 is asking: "ability to use nested props as key (user.email)".

toddbaert:
Hey @craigpastro ! If you look at the [fractional evaluation doc](https://github.com/open-feature/flagd/blob/main/core/pkg/eval/fractional_evaluation.go#L54C11-L54C18), the first argument is the property we use for the deterministic bucketing (in addition to the key, which we will add, as you mention). The problem is, it seems to me this property MUST be a top level property in the map.

It would be great if we could express the desire to use a nested property, something like `user.email` if it exists in the context. JSON logic allows for the declaration and extractions of variables, so we may be able to leverage that.

[...]

PR #833, opened 2023-08-14
docs: Move fractional evaluation pages to website

agardnerIT:
## This PR
- Moves fractional evaluation information to website

PR #846, opened 2023-08-17
refactor: replace xxh3 with murmur3

craigpastro:
## This PR

Replace xxh3 with murmur3 in fractional evaluation's bucketing algorithm.

### Related Issues

Part 1 of #843.


core/pkg/eval/fractional_evaluation.go
```diff
@@ -58,7 +58,7 @@ func parseFractionalEvaluationData(values, data interface{}) (string, []fraction
 
 	valueToDistribute, ok := v.(string)
 	if !ok {
-		return "", nil, fmt.Errorf("var: %s isn't of type string", bucketBy)
+		return "", nil, fmt.Errorf("var: %s isn't of type string", v)
```
toddbaert, line 61: nice catch.

toddbaert (approved):
Thanks!

beeme1mr:
Hey @craigpastro, thanks for working on this! Could you please update the technical section in the docs to reference to murmur3 instead of xxh3?

https://github.com/open-feature/flagd/blob/main/docs/configuration/fractional_evaluation.md#fractional-evaluation-technical-description

craigpastro:
👋  @ beeme1mr. Thank you. I've updated the docs and linked to the reference implementation. I don't know if that is the right thing to link to, but I didn't find anything better.

Kavindu-Dodan (approved):
Changes looks good. Seems the test is now evenly spreadout (contain all four colours)

@craigpastro  yes you have updated the docs per current main branch status. However, this will conflict with change proposed through PR https://github.com/open-feature/flagd/pull/833 

@agardnerIT fyi

craigpastro:
👋 @Kavindu-Dodan. Thank you for the review and for noticing that the documentation will conflict. I can rebase if #833 is merged first.

core/pkg/eval/fractional_evaluation.go
thisthat, line 61: maybe it is better to use the default format since we know it is not a string
```suggestion
		return "", nil, fmt.Errorf("var: %v isn't of type string", v)
```

thisthat (approved):
LGTM 👍 I left a minor comment for you to consider @craigpastro

beeme1mr (approved):

toddbaert:
I think [this](https://github.com/open-feature/flagd/pull/849) is causing the build failure.

PR #847, opened 2023-08-17
feat: add flag key to hash in fractional evaluation

craigpastro:
## This PR

This PR adds the flag key as input to the hash in fractional evaluation. Since the evaluation function only has access to user-supplied logic and context, which almost certainly will not contain the flag key, in order to get this data to the evaluation function I add a `flagdProperties` struct to the context via a namespaced key `$flagd`. The `flagdProperties` struct contains the flag key, which is then used as input to the hash.

### Related Issues

Part 2 of #843.


core/pkg/eval/fractional_evaluation_test.go
```diff
@@ -63,46 +63,99 @@ func TestFractionalEvaluation(t *testing.T) {
 		expectedReason  string
 		expectedError   error
 	}{
-		"test@faas.com": {
+		"rachel@faas.com": {
```
craigpastro, line 66: I changed the emails so that we get all the colours. Probably a waste of time as it will change if #846 gets merged 🙃

toddbaert: > I changed the emails so that we get all the colours. Probably a waste of time as it will change if #846 gets merged 🙃

I like the idea though, lol

core/pkg/eval/json_evaluator.go
```diff
@@ -24,7 +24,11 @@ import (
 	"go.uber.org/zap"
 )
 
-const SelectorMetadataKey = "scope"
+const (
+	SelectorMetadataKey = "scope"
+
+	contextFlagKey = "$flagd.__flag_key__"
```
toddbaert, line 30: Two thoughts about this:

- I like your `$flagd.` proposal, I think it provides enough namespacing that we don't need the `__`... so `$flagd.flag_key` seems adequate.
- We could make `$flagd` a property, with nested values (just `flag_key` for now), so to access it in a rule, we'd do `{"var" : "$flagd.flag_key"}`. If we don't do it this way, I think we'd need to escape the `.` in the name, which would feel clumsy, I think.
  - If for some reason we don't want to use nesting, we should NOT use `.` as a delimeter. Maybe `__` instead.

core/pkg/eval/json_evaluator.go
```diff
@@ -306,6 +310,15 @@ func (je *JSONEvaluator) evaluateVariant(reqID string, flagKey string, context m
...
 		}
 
+		if context == nil {
+			context = map[string]any{}
+		}
+
+		// Don't overwrite this value if a user has somehow set it.
+		if _, ok := context[contextFlagKey]; !ok {
+			context[contextFlagKey] = flagKey
+		}
```
toddbaert, line 320: I think we should do the opposite, and overwrite values in the `$flagd` property, but maybe warn when we do it.

toddbaert:
@craigpastro 

Could we add another test that just directly references `$flagd.flag_key` in a rule? Something that returns a value if `flagd.flag_key` === {some-known-value}. This will assure that we are actually populating the value as expected.

core/pkg/eval/json_evaluator.go
beeme1mr: I agree with everything you said. Having these values accessible in a user's configuration would be very helpful. It may be interesting once we allow users to access nested properties to be able to do something like this:

```json
{
	"fractionalEvaluation": [
		{"cat": [{"var":"$flagd.flag_key"}, {"var":"user.email"}]},
		[ "red", 25 ],
		[ "blue", 25 ],
		[ "green", 25 ],
		[ "yellow", 25]
	]
}
```

Automatically setting the current timestamp and other properties sounds like a good idea. We should create a follow up issue to track the ideas and implementation.

toddbaert: > I agree with everything you said. Having these values accessible in a user's configuration would be very helpful. It may be interesting once we allow users to access nested properties to be able to do something like this:
> 
> ```json
> {
> 	"fractionalEvaluation": [
> 		{"cat": [{"var":"$flagd.flag_key"}, {"var":"user.email"}]},
> 		[ "red", 25 ],
> 		[ "blue", 25 ],
> 		[ "green", 25 ],
> 		[ "yellow", 25]
> 	]
> }
> ```
> 
> Automatically setting the current timestamp and other properties sounds like a good idea. We should create a follow up issue to track the ideas and implementation.

I could be confused, but I think we'd do the `cat` operation ourselves, right? So the users would still just mention the prop they want to bucket on, and we'd `cat` it in the evaluator implementation. The only thing missing is the prop user's specify can't currently be a nested prop: `some.nested.prop".

toddbaert:
Related: https://github.com/open-feature/flagd/issues/851

core/pkg/eval/json_evaluator.go
toddbaert: @craigpastro:

@beeme1mr and I were talking more about this, and we came up with a proposal we'd like your input on:

We're thinking about changing the `fractionalEvaluation` in a couple ways:

- Support a version that doesn't require an explicit bucketing prop, and by default uses `flag_key` + `targetingKey` as discussed. This would look like :

```json
{
	"fractionalEvaluation": [
		[ "red", 25 ],
		[ "blue", 25 ],
		[ "green", 25 ],
		[ "yellow", 25 ]
	]
}
```

Which would probably be what people want in 90% of use-cases
- Allow an a json logic expression for the first param to derive a custom bucketing value, if you dont want to do above, which would look like:

```json
{
	"fractionalEvaluation": [
		{"cat": [{"var":"$flagd.flag_key"}, {"var":"user.email"}]},
		[ "red", 25 ],
		[ "blue", 25 ],
		[ "green", 25 ],
		[ "yellow", 25]
	]
}
```

- change the name to `fractional` (`fractionalEvaluation` is redundant), and keep the current implementation as is, but deprecate it so we can remove it with 1.0.

craigpastro: 1. The first proposal sounds nice and simplifies things a bit, which is always nice. Where does `targetingKey` come from in this case? Will you need to specify it in `targeting`?
```
"targeting": {
  "key": "email",
  "fractionalEvaluation": [...]
}
```

2. This is also nice! In this case it means that `jsonlogic` will handle the logic as long as the correct stuff is added to the context, is that right?

craigpastro: > I like your $flagd. proposal,
 
 That is @beeme1mr's proposal from https://github.com/open-feature/flagd/issues/843#issuecomment-1682718897.

toddbaert: `targetingKey` is a property we can assume will be present in the context in this case (and if it's not, we can log a warning and default the rule). This is a property we can assume to be defined. It's something we mention in the OpenFeature spec: https://openfeature.dev/specification/sections/evaluation-context#requirement-311

craigpastro: Ah, I didn't know that. Thank you. If we can assume that `targetingKey` is in the context, then we don't need to add it from the JSON evaluator do we? That is, this PR could be simplified quite a bit?

core/pkg/eval/json_evaluator.go
craigpastro: Good point. Updated. Thank you!

craigpastro:
> Could we add another test that just directly references $flagd.flag_key in a rule? Something that returns a value if flagd.flag_key === {some-known-value}. I know nobody would often write a rule like this in real life, but it will assure that we are actually populating the value as expected. See json logic do [here](https://jsonlogic.com/operations.html#var).

I've added a test. Is it what you had in mind?

core/pkg/eval/fractional_evaluation_test.go
craigpastro: We still get all the colours :metal:

core/pkg/eval/json_evaluator_test.go
```diff
@@ -1207,3 +1207,45 @@ func TestFlagStateSafeForConcurrentReadWrites(t *testing.T) {
...
+			"flags": {
+				"welcome-banner": {
+					"state": "ENABLED",
+					"variants": {
+						"true": true,
+						"false": false
+					},
+					"defaultVariant": "false",
+					"targeting": {
+						"==": [ { "var": "$flagd.flag_key" }, "welcome-banner" ]
```
bacherfl, line 1225: I think this needs to be changed to match the field name of the `flagdProperties` struct
```suggestion
						"==": [ { "var": "$flagd.FlagKey" }, "welcome-banner" ]
```

core/pkg/eval/json_evaluator.go
```diff
@@ -336,6 +349,37 @@ func (je *JSONEvaluator) evaluateVariant(reqID string, flagKey string, context m
...
+}
+
+func getFlagdProperties(context map[string]any) (flagdProperties, bool) {
+	properties, ok := context[flagdPropertiesKey]
+	if !ok {
+		return flagdProperties{}, false
+	}
+
+	var p flagdProperties
+	if err := mapstructure.Decode(properties, &p); err != nil {
```
bacherfl, line 376: here we could also use a `json.Marshal/json.Unmarshal`, if we do not want to introduce a new dependency to github.com/mitchellh/mapstructure

core/pkg/eval/json_evaluator_test.go
toddbaert: Yes, I think this is true... however, I'm not sure we want to use this casing. All implementations will have to populate this prop, and in that case, I would advocate for something more in line with start JSON casing (ie: camelCase), so `flagKey`.

What do you folks think? @craigpastro @bacherfl @beeme1mr

toddbaert (changes requested):
Once [this](https://github.com/open-feature/flagd/pull/847/files#r1301178728) and [this](https://github.com/open-feature/flagd/pull/847/files#r1301186781) are resolved, I think I can approve.

core/pkg/eval/json_evaluator.go
toddbaert: I'm not sure you'll want to add this to this PR (though you can if you'd like).

core/pkg/eval/json_evaluator_test.go
bacherfl: I agree, `flagKey` would be more in line with typical json casing

craigpastro: Ah, silly me. I updated last night and failed to update the test. I also like camelCase for JSON fields so I'll update to `flagKey`.

core/pkg/eval/json_evaluator.go
craigpastro: Oh yeah, that is much nicer. Thank you! Updated.

core/pkg/eval/fractional_evaluation.go
```diff
@@ -41,7 +41,7 @@ func parseFractionalEvaluationData(values, data interface{}) (string, []fraction
 		return "", nil, errors.New("fractional evaluation data has length under 2")
 	}
 
-	bucketBy, ok := valuesArray[0].(string)
+	targetingKey, ok := valuesArray[0].(string)
```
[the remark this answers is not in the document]
craigpastro: I see. Thank you. Reverted to `bucketBy`.

craigpastro:
The length of `evaluateVariant` is failing the linter. I don't see any nice way to restructure the function. I can make it pass by removing whitespace but that doesn't seem like a good idea, and if more properties are added to `flagdProperties` it will just fail again in the future. Added a `// nolint:funlen` is an option.

core/pkg/eval/json_evaluator.go
```diff
@@ -34,7 +33,7 @@ const (
 var regBrace *regexp.Regexp
 
 type flagdProperties struct {
-	FlagKey string
+	FlagKey string `json:"flagKey"`
```
toddbaert, line 36: I swear, this single feature might be the best part of Go.

toddbaert (approved):
LGTM. I think we can differ documenting this feature until we decide to add other properties (such as `current_timestamp`).

We can tackle the changes described [here](https://github.com/open-feature/flagd/pull/848#discussion_r1301661915) in a separate PR.

toddbaert:
> The length of `evaluateVariant` is failing the linter. I don't see any nice way to restructure the function. I can make it pass by removing whitespace but that doesn't seem like a good idea, and if more properties are added to `flagdProperties` it will just fail again in the future. Added a `// nolint:funlen` is an option.

I've added the `nolint`. A lot of length in this func is coming from debug messages and helpful comments. It doesn't read as overly complex to me.

beeme1mr (approved):

PR #849, opened 2023-08-21
chore: pin golangci-lint

toddbaert:
golangci requires go 1.20 now :grimacing: . 

We can upgrade to 1.20, but for now the change in this tool is causing lint failures since we use `@latest` when we install it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,8 @@ uninstall:
 	rm /etc/systemd/system/flagd.service
 	rm -f $(DESTDIR)$(PREFIX)/bin/flagd
 lint:
-	go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@latest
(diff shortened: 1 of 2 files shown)


thisthat (approved):

Makefile
```diff
@@ -60,7 +60,8 @@ uninstall:
 	rm /etc/systemd/system/flagd.service
 	rm -f $(DESTDIR)$(PREFIX)/bin/flagd
 lint:
-	go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+# pinned to @v1.50.1 until we migrate to go 1.20
```
beeme1mr, line 63: Can you please create an issue to upgrade to 1.20 and reference this PR?

beeme1mr (approved):

PR #854, opened 2023-08-22
ci: ensure unit tests are run in PR checks

### LINKED ISSUE #853: [BUG] Unit tests are not executed in PR checks
### Observed behavior

Currently, unit tests are not executed properly in the PR checks.
As can be seen in e.g. https://github.com/open-feature/flagd/actions/runs/5927342581/job/16070661947,
when the `make test` command is being executed the step immediately exits successfully with the following message:

```
make: 'test' is up to date.
``` 

### Expected Behavior

Unit tests should be executed in the PR checks to ensure no regression is introduced.

### Steps to reproduce

1. Create a PR
2. Let the PR checks run
3. Inspect the `test` job - there no actual unit test will be executed

bacherfl:
Closes #853 

This PR fixes the behavior described in the referenced issue by makeing the `test` target in the makefile a `.PHONY`.

The issue was occuring because in the same directory as the Makefile, there was a directory with the same name as the `test` target, which caused the target to not run - For more details, see [this thread](https://stackoverflow.com/questions/3931741/why-does-make-think-the-target-is-up-to-date) on StackOverflow

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,7 @@ build: workspace-init # default to flagd
 	make build-flagd
 build-flagd:
 	go build -ldflags "-X main.version=dev -X main.commit=$$(git rev-parse --short HEAD) -X main.date=$$(date +%FT%TZ)" -o ./bin/flagd ./flagd
+.PHONY: test
 test: # default to core
 	make test-core
 test-core:


odubajDT (approved):

beeme1mr (approved):

toddbaert:
ooof, I even noticed this, but I was running `make test-core` to work around it. I should have realized it was impacting the CI.

Issue #831, opened 2023-08-11
[FEATURE] gRPC Health Check must be implemented in case flagd is accessed via AWS ALB


kworkbee:
### Requirements

To configure flagd to be accessed via AWS ALB when deployed as standalone rather than as OpenFeature Operator, it must be specified the health check related properties of ALB, which must be implemented as '/AWS.ALB/healthcheck' path in flagd.

toddbaert:
@Kavindu-Dodan @bacherfl @thisthat is this as simple as making `healthz/readyz` configurable?

Kavindu-Dodan:
@kinyoklion  I don't fully understand your requirement.

Regardless of the application you are deploying (here you are referencing to flagd), AWS allows your to configure health check endpoint of your ALB.

Please check these steps to configure this [^1] in your EC2 instance.

flagd expose readiness & liveness probes as described in our documentation [^2]. So you can configure health check endpoints for them. 

[^1]: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/target-group-health-checks.html#modify-health-check-settings
[^2]: https://github.com/open-feature/flagd/blob/main/web-docs/concepts/architecture.md#readiness--liveness-probes

toddbaert:
I think you at-ed @kinyoklion when you meant to at @kworkbee

Kavindu-Dodan:
> @Kavindu-Dodan I think you at-ed @kinyoklion when you meant to at @kworkbee 😅

yeah, my bad 😞 . updated the comment :D

kworkbee:
@toddbaert @Kavindu-Dodan Hello, Thanks for your answer.

I tried it, but due to AWS ALB Ingress Controller setting, if backend protocol version uses gRPC, it seems that only gRPC is available as healthcheck. You can specify the (Not HTTP) Status Code only as the status code of the gRPC.

beeme1mr:
@Kavindu-Dodan, would it be possible to add support for gRPC-based health checks in addition to HTTP?

toddbaert:
I can look into this. It seems that we do indeed need a gRPC healthcheck that returns a value between 0-99 (probably 12) in order to satisfy AWS' requirements.

toddbaert:
I think this is a legimitate issue and we should fix it by adding a standard gRPC health endpoint, which conforms to [this standard proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto).

toddbaert:
[...]

PR #861, opened 2023-08-24
feat: add grpc health, deprecate old health

toddbaert:
* adds a standard gRPC health check
* moves HTTP health checks off the metrics port, to the app port (we previously hadn't documented the health-checks are controlled by the metrics port config)
* deprecates old health checks on metrics port (logs warning)

fixes: https://github.com/open-feature/flagd/issues/831

### DIFF
--- a/config/deployments/flagd/deployment.yaml
+++ b/config/deployments/flagd/deployment.yaml
@@ -29,13 +29,13 @@ spec:
           readinessProbe:
             httpGet:
               path: /readyz
-              port: 8014
+              port: 8013
             initialDelaySeconds: 5
             periodSeconds: 5
           livenessProbe:
             httpGet:
               path: /healthz
-              port: 8014
+              port: 8013
             initialDelaySeconds: 5
             periodSeconds: 60
           ports:
--- a/core/pkg/service/flag-evaluation/connect_service.go
+++ b/core/pkg/service/flag-evaluation/connect_service.go
@@ -220,10 +220,12 @@ func (s *ConnectService) startMetricsServer(svcConf service.Configuration) error
 	s.metricsServer.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
 		case "/healthz":
-			s.logger.Warn(fmt.Sprintf("call to deprecated /healthz endpoint on port %d, use port %d instead", svcConf.MetricsPort, svcConf.Port))
+			s.logger.Warn(
+				fmt.Sprintf("/healthz endpoint on port %d is deprecated, use port %d instead", svcConf.MetricsPort, svcConf.Port))
 			w.WriteHeader(http.StatusOK)
 		case "/readyz":
-			s.logger.Warn(fmt.Sprintf("call to deprecated /readyz endpoint on port %d, use port %d instead", svcConf.MetricsPort, svcConf.Port))
+			s.logger.Warn(
+				fmt.Sprintf("/readyz endpoint on port %d is deprecated, use port %d instead", svcConf.MetricsPort, svcConf.Port))
 			if s.readinessEnabled && svcConf.ReadinessProbe() {
 				w.WriteHeader(http.StatusOK)
 			} else {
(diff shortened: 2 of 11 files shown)


core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -129,16 +130,28 @@ func (s *ConnectService) setupServer(svcConf service.Configuration) (net.Listene
 
 	path, handler := schemaConnectV1.NewServiceHandler(fes, append(svcConf.Options, marshalOpts)...)
 	mux.Handle(path, handler)
+	mux.Handle(grpchealth.NewHandler(grpchealth.NewStaticChecker(
```
toddbaert, line 133: Standard implementation, recommended by buf/conntect.

You can test this locally with `grpcurl` (below command assumes you've downloaded the [health proto](https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto) to `~/Downloads`:

```bash
grpcurl -import-path '~/Downloads/'  -proto health.proto -plaintext localhost:8013 grpc.health.v1.Health/Check
```

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -129,16 +130,28 @@ func (s *ConnectService) setupServer(svcConf service.Configuration) (net.Listene
...
+		} else {
+			w.WriteHeader(http.StatusPreconditionFailed)
+		}
+	}))
 
 	s.serverMtx.Lock()
 	s.server = &http.Server{
 		ReadHeaderTimeout: time.Second,
-		Handler:           handler,
+		Handler:           mux,
```
toddbaert, line 150: The `mux` created above was actually unused - instead we just used the one handler originally.

Instead of adding some conditions based on paths our other ready handler, I've used the `mux` which simplifies things a bit (this is what mux is for).

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -206,8 +219,10 @@ func (s *ConnectService) startMetricsServer(svcConf service.Configuration) error
 	s.metricsServer.Handler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
```
toddbaert, line 220: In my new implementation, I've used `mux` instead of this kind of switch statement.

core/pkg/service/flag-evaluation/connect_service.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	schemaConnectV1 "buf.build/gen/go/open-feature/flagd/bufbuild/connect-go/schema/v1/schemav1connect"
+	"connectrpc.com/grpchealth"
```
toddbaert, line 14: this package is a connect-compatible implementation of a gRPC healthcheck.

config/deployments/flagd/deployment.yaml
```diff
@@ -29,13 +29,13 @@ spec:
           readinessProbe:
             httpGet:
               path: /readyz
-              port: 8014
```
thisthat, line 32: please, adapt also the ZD tests manifest to use the new port :)

thisthat (commented):
We might consider adding a test using https://github.com/grpc-ecosystem/grpc-health-probe to check the gRPC probes are correct. Currently, the ZD tests use the HTTP probes.

toddbaert:
> We might consider adding a test using https://github.com/grpc-ecosystem/grpc-health-probe to check the gRPC probes are correct. Currently, the ZD tests use the HTTP probes.



> We might consider adding a test using https://github.com/grpc-ecosystem/grpc-health-probe to check the gRPC probes are correct. Currently, the ZD tests use the HTTP probes.

great idea!

Kavindu-Dodan (approved):
overal looks good 🙌 

approving and hope there will be improvements proposed by this comment - https://github.com/open-feature/flagd/pull/861#pullrequestreview-1595033315

config/deployments/flagd/deployment.yaml
toddbaert: @thisthat flagd-proxy uses the sync package from core, NOT the `flag-evaluation` package. The health-checks of flagd-proxy are not impacted by this change (which may or may not be good, since things will be a bit inconsistent.

Problem is, I'm not 100% sure I can add HTTP 1.1 health-check support to the sync package since it doesn't use `connect`.

toddbaert:
Closed in favor of https://github.com/open-feature/flagd/pull/863

PR #863, opened 2023-08-25
feat: add gRPC healthchecks

toddbaert:
* adds a gRPC healthcheck to the "metrics" port
* supports both gRPC and HTTP on "metrics" port
* adds relevant doc

-----

@thisthat @Kavindu-Dodan @beeme1mr 

I've closed the previous PR and gone about this a different way... Primarily because I realized it's a bad idea to put the healthcheck on the "main" socket, since that can be bound not just to a TCP socket, but to a unix socket, which would be very problematic for a healthcheck :sweat_smile: .

I've instead added the new gRPC health checks to the "metrics" ports of both `flagd` and `flagd-proxy` with some basic home-grown multiplexing code. **_I strongly recommend we re-name this port from "metrics" to "management" in a subsequent PR, since we are not just using it for metrics_** (we can probably use viper to make this non-breaking, ie use "metricsPort" if "managementPort" is missing).

@thisthat another bonus is this approach doesn't require and new deps, and therefore no proto conflicts!

fixes: https://github.com/open-feature/flagd/issues/831

### DIFF
--- a/core/pkg/service/sync/server.go
+++ b/core/pkg/service/sync/server.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"strings"
 	"time"
 
 	rpc "buf.build/gen/go/open-feature/flagd/grpc/go/sync/v1/syncv1grpc"
@@ -13,8 +14,12 @@ import (
 	iservice "github.com/open-feature/flagd/core/pkg/service"
 	syncStore "github.com/open-feature/flagd/core/pkg/sync-store"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/health"
+	"google.golang.org/grpc/health/grpc_health_v1"
 )
 
 type Server struct {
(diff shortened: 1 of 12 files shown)


flagd/build.Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=bind,source=./core,target=./core \
     --mount=type=bind,source=./flagd,target=./flagd \
-    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -a -ldflags "-X main.version=${VERSION} -X main.commit=${COMMIT} -X main.date=${DATE}" -o /bin/flagd-build flagd/main.go
+    CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -a -ldflags "-X google.golang.org/protobuf/reflect/protoregistry.conflictPolicy=ignore -X main.version=${VERSION} -X main.commit=${COMMIT} -X main.date=${DATE}" -o /bin/flagd-build flagd/main.go
```
Kavindu-Dodan, line 32: Given we do not have any conflicts , this should be able to remove (also check profile.Dockerfile)

Kavindu-Dodan:
Overal looks good 🙌

thisthat (commented):
Beside the comment of @Kavindu-Dodan, LGTM 👍

flagd/build.Dockerfile
toddbaert: nice catch, thought I got all of these.

toddbaert:
@Kavindu-Dodan @thisthat I've fixed that issue. Please re-review.

thisthat (approved):

Kavindu-Dodan (approved):

core/pkg/service/sync/server.go
```diff
@@ -101,26 +106,38 @@ func (s *Server) startServer() error {
...
+		} else {
+			mux.ServeHTTP(writer, request)
+			return
 		}
 	})
+
+	s.metricsServer = &http.Server{
+		Addr:              fmt.Sprintf(":%d", s.config.MetricsPort),
+		ReadHeaderTimeout: 3 * time.Second,
+		Handler:           h2c.NewHandler(handler, &http2.Server{}), // we need ot use h2c to support plaintext HTTP2
```
beeme1mr, line 139:
```suggestion
		Handler:           h2c.NewHandler(handler, &http2.Server{}), // we need to use h2c to support plaintext HTTP2
```

docs/other_resources/high_level_architecture.md
```diff
@@ -35,14 +35,19 @@ process gets pushed to event subscribers.
...
-With default
-start-up configurations, these probes are exposed at the following URLs,
+With default start-up configurations, these probes are exposed on the metrics port (default: 8014) at the following URLs,
 
 - Liveness: <http://localhost:8014/healthz>
 - Readiness: <http://localhost:8014/readyz>
 
+### gRPC
+
+Flagd exposes a [standard gRPC liveness check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) on the metrics port (default: 8014).
```
beeme1mr, line 49: Can you include a sample path like we have for the HTTP probe?

web-docs/concepts/architecture.md
```diff
@@ -33,14 +33,19 @@ process gets pushed to event subscribers.
...
-With default
-start-up configurations, these probes are exposed at the following URLs,
+With default start-up configurations, these probes are exposed on the metrics port (default: 8014) at the following URLs,
 
 - Liveness: <http://localhost:8014/healthz>
 - Readiness: <http://localhost:8014/readyz>
 
+### gRPC
+
+Flagd exposes a [standard gRPC liveness check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) on the metrics port (default: 8014).
```
beeme1mr, line 47: Same comment as above.

beeme1mr (approved):

web-docs/concepts/architecture.md
toddbaert: See: https://github.com/open-feature/flagd/pull/863#discussion_r1308806528

docs/other_resources/high_level_architecture.md
Kavindu-Dodan: I also agree with @toddbaert here. 

We cannot specify gRPC URL without the schema. IMO having the link to health-check documentation [^1] should be sufficient. But may be we can add the line (extracted from reference doc)

> "A client can query the server’s health status by calling the Check method, and a deadline should be set on the rpc"

And say we do not use/support `service` filtering  

[^1]: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

toddbaert: I am resolving this for now @beeme1mr . I'll open a new PR if you think additional doc is needed.

PR #848, opened 2023-08-21
docs: add specification for in-process flagd provider implementations

bacherfl:
Closes #842

This PR adds the specification for in-process flagd providers. For now I have added this to the `other_ressources` section of the docs, but please let me know if another directory would be more fitting for this document

PR #869, opened 2023-08-28
feat: support nested props in fractional evaluator

### LINKED ISSUE #843: Improvements to fractional custom evalutor

craigpastro:
## This PR

This PR adds the ability to use nested properties for the bucket key in fractional evaluation. It is a breaking change from the current behaviour. If the bucket key is not provided it will default to `{"cat": [{"var":"$flagd.flagKey"}, {"var":"targetingKey"}]}`.

@toddbaert I am not sure that my approach to the missing bucket key is the best. I would have preferred to add `{"cat": [{"var":"$flagd.flagKey"}, {"var":"targetingKey"}]}` to the targeting object and let JsonLogic handle it, but it would be a bit clumsy and involve string manipulation. I don't think there is a nice way to do it, but I'll play around a bit more.

### Related Issues

See #848.
Closes #843.

### DIFF
--- a/core/pkg/eval/fractional_evaluation.go
+++ b/core/pkg/eval/fractional_evaluation.go
@@ -9,6 +9,8 @@ import (
 	"github.com/twmb/murmur3"
 )
 
+const targetingKeyKey = "targetingKey"
+
 type FractionalEvaluator struct {
 	Logger *logger.Logger
 }
@@ -41,11 +43,6 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 		return "", nil, errors.New("fractional evaluation data has length under 2")
 	}
 
-	bucketBy, ok := valuesArray[0].(string)
-	if !ok {
-		return "", nil, errors.New("first element of fractional evaluation did not resolve to a string value")
-	}
-
 	dataMap, ok := data.(map[string]any)
 	if !ok {
 		return "", nil, errors.New("data isn't of type map[string]any")
@@ -54,6 +51,16 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 	// Ignore the error as we can't really do anything if the properties are
 	// somehow missing.
 	properties, _ := getFlagdProperties(dataMap)
+
+	bucketBy, ok := valuesArray[0].(string)
+	if ok {
+		valuesArray = valuesArray[1:]
+	} else {
+		bucketBy, ok = dataMap[targetingKeyKey].(string)
+		if !ok {
+			return "", nil, errors.New("targeting key not supplied nor in context")
+		}
+	}
 
 	feDistributions, err := parseFractionalEvaluationDistributions(valuesArray)
 	if err != nil {
@@ -66,7 +73,7 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluationDistribution, error) {
 	sumOfPercentages := 0
 	var feDistributions []fractionalEvaluationDistribution
-	for i := 1; i < len(values); i++ {
+	for i := 0; i < len(values); i++ {
 		distributionArray, ok := values[i].([]any)
 		if !ok {
 			return nil, errors.New("distribution elements aren't of type []any")
--- a/core/pkg/eval/fractional_evaluation.go
+++ b/core/pkg/eval/fractional_evaluation.go
@@ -67,7 +67,7 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 		return "", nil, err
 	}
 
-	return fmt.Sprintf("%s$%s", properties.FlagKey, bucketBy), feDistributions, nil
+	return fmt.Sprintf("%s%s", properties.FlagKey, bucketBy), feDistributions, nil
 }
 
 func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluationDistribution, error) {
--- a/core/pkg/eval/fractional_evaluation.go
+++ b/core/pkg/eval/fractional_evaluation.go
@@ -9,8 +9,6 @@ import (
 	"github.com/twmb/murmur3"
 )
 
-const targetingKeyKey = "targetingKey"
-
 type FractionalEvaluator struct {
 	Logger *logger.Logger
 }
@@ -58,7 +56,7 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 	} else {
 		bucketBy, ok = dataMap[targetingKeyKey].(string)
 		if !ok {
-			return "", nil, errors.New("targeting key not supplied nor in context")
+			return "", nil, errors.New("bucketing value not supplied and no targetingKey in context")
 		}
 	}
 
--- a/core/pkg/eval/fractional_evaluation.go
+++ b/core/pkg/eval/fractional_evaluation.go
@@ -22,7 +22,7 @@ func NewFractionalEvaluator(logger *logger.Logger) *FractionalEvaluator {
 	return &FractionalEvaluator{Logger: logger}
 }
 
-func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) interface{} {
+func (fe *FractionalEvaluator) FractionalEvaluation(values, data any) any {
 	valueToDistribute, feDistributions, err := parseFractionalEvaluationData(values, data)
 	if err != nil {
 		fe.Logger.Error(fmt.Sprintf("parse fractional evaluation data: %v", err))
@@ -32,8 +32,8 @@ func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) in
 	return distributeValue(valueToDistribute, feDistributions)
 }
 
-func parseFractionalEvaluationData(values, data interface{}) (string, []fractionalEvaluationDistribution, error) {
-	valuesArray, ok := values.([]interface{})
+func parseFractionalEvaluationData(values, data any) (string, []fractionalEvaluationDistribution, error) {
+	valuesArray, ok := values.([]any)
 	if !ok {
 		return "", nil, errors.New("fractional evaluation data is not an array")
 	}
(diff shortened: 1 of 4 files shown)


core/pkg/eval/fractional_evaluation.go
```diff
@@ -22,7 +24,7 @@ func NewFractionalEvaluator(logger *logger.Logger) *FractionalEvaluator {
 	return &FractionalEvaluator{Logger: logger}
 }
 
-func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) interface{} {
+func (fe *FractionalEvaluator) FractionalEvaluation(values, data any) any {
```
craigpastro, line 27: Since we are on Go 1.19 we can replace `interface{}` with `any`. I replaced all occurrences in this file.

toddbaert: A definite readability improvement.

core/pkg/eval/fractional_evaluation.go
```diff
@@ -32,54 +34,49 @@ func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) in
...
+		}
 	}
 
 	feDistributions, err := parseFractionalEvaluationDistributions(valuesArray)
 	if err != nil {
 		return "", nil, err
 	}
 
-	return fmt.Sprintf("%s$%s", properties.FlagKey, valueToDistribute), feDistributions, nil
+	return fmt.Sprintf("%s$%s", properties.FlagKey, bucketBy), feDistributions, nil
```
toddbaert, line 70:
```suggestion
	return fmt.Sprintf("%s%s", properties.FlagKey, bucketBy), feDistributions, nil
```
I think this token technically violates the specification. We can actually add it, but I'm dubious of the value, and in other implementations that actually inject the concatenation expression, it might just make the impl more difficult.

core/pkg/eval/fractional_evaluation.go
```diff
@@ -32,54 +34,49 @@ func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) in
...
-	if !ok {
-		return "", nil, fmt.Errorf("var: %v isn't of type string", v)
+	bucketBy, ok := valuesArray[0].(string)
+	if ok {
+		valuesArray = valuesArray[1:]
+	} else {
+		bucketBy, ok = dataMap[targetingKeyKey].(string)
+		if !ok {
+			return "", nil, errors.New("targeting key not supplied nor in context")
+		}
```
toddbaert, line 62: I think this impl is just fine, and much cleaner than manually injecting the concatination expression as you mentioned.

docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md
```diff
@@ -10,9 +10,9 @@ referenced evaluation context property, the [MurmurHash3](https://github.com/aap
 hash function should be used. This is to ensure that flag resolution requests yield the same result,
 regardless of which implementation of the in-process flagd provider is being used.
 
-array containing at least two items, with the first item being an optional [json logic variable declaration](https://jsonlogic.com/operations.html#var)
+The supplied array must contain at least two items, with the first item being an optional [json logic variable declaration](https://jsonlogic.com/operations.html#var)
 specifying the target property to base the distribution of values on. If not supplied, a concatination of the
-`flagKey` and `targetingKey` are used: `{"cat": [{"var":"$flagd.flag_key"}, {"var":"user.email"}]}`.
+`flagKey` and `targetingKey` are used: `{"cat": [{"var":"$flagd.flagKey"}, {"var":"targetingKey"}]}`.
```
toddbaert, line 15: nice catch!

docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md
```diff
@@ -59,18 +59,19 @@ The following flow chart depicts the logic of this evaluator:
 
 ```mermaid
 flowchart TD
-A[Parse targetingRule] --> B{Is an array containing at least two items?};
+A[Parse targetingRule] --> B{Is an array containing at least one item?};
```
toddbaert, line 62: Thanks!

core/pkg/eval/fractional_evaluation.go
```diff
@@ -32,54 +34,49 @@ func (fe *FractionalEvaluator) FractionalEvaluation(values, data interface{}) in
...
-	valueToDistribute, ok := v.(string)
-	if !ok {
-		return "", nil, fmt.Errorf("var: %v isn't of type string", v)
+	bucketBy, ok := valuesArray[0].(string)
+	if ok {
+		valuesArray = valuesArray[1:]
+	} else {
+		bucketBy, ok = dataMap[targetingKeyKey].(string)
+		if !ok {
+			return "", nil, errors.New("targeting key not supplied nor in context")
```
toddbaert, line 61:
```suggestion
			return "", nil, errors.New("bucking value not supplied and no targetingKey in context")
```

I think this is a bit clearer... but let me know if you agree. The `targetingKey` is technically being used as a `bucketing value`.

docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md
```diff
@@ -10,9 +10,9 @@ referenced evaluation context property, the [MurmurHash3](https://github.com/aap
 hash function should be used. This is to ensure that flag resolution requests yield the same result,
 regardless of which implementation of the in-process flagd provider is being used.
 
-array containing at least two items, with the first item being an optional [json logic variable declaration](https://jsonlogic.com/operations.html#var)
+The supplied array must contain at least two items, with the first item being an optional [json logic variable declaration](https://jsonlogic.com/operations.html#var)
 specifying the target property to base the distribution of values on. If not supplied, a concatination of the
```
toddbaert, line 14:
```suggestion
specifying the bucketing property to base the distribution of values on. If not supplied, a concatination of the
```

docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md
```diff
@@ -59,18 +59,19 @@ The following flow chart depicts the logic of this evaluator:
...
-B -- No --> D[Return nil];
-C -- Yes --> E[targetPropertyValue := targetingRule at index 0];
-C -- No --> D;
-E -- Yes --> F[Iterate through the remaining elements of the targetingRule array and parse the variants and their percentages];
-F --> G{Parsing successful?};
-G -- No --> D;
-G -- Yes --> H{Does percentage of variants add up to 100?};
+B -- No --> D[return nil]
+C -- No --> E[targetPropertyValue := default to targetingKey];
+C -- Yes --> F[targetPropertyValue := targetingRule at index 0];
```
toddbaert, line 66:
```suggestion
C -- No --> E[bucketingPropertyValue := default to targetingKey];
C -- Yes --> F[bucketingPropertyValue := targetingRule at index 0];
```

docs/other_resources/in-process-providers/evaluators/fractional_evaluation.md
```diff
@@ -59,18 +59,19 @@ The following flow chart depicts the logic of this evaluator:
...
+E --> G[Iterate through the remaining elements of the targetingRule array and parse the variants and their percentages];
+F --> G;
+G --> H{Parsing successful?};
 H -- No --> D;
-H -- Yes --> I[hash := murmur3Hash of targetPropertyValue divided by Int64.MaxValue]
-I --> L[Iterate through the variant and increment the threshold by the percentage of each variant. Return the first variant where the bucket is smaller than the threshold.]
+H -- Yes --> I{Does percentage of variants add up to 100?};
+I -- No --> D;
+I -- Yes --> J[hash := murmur3Hash of targetPropertyValue divided by Int64.MaxValue]
```
toddbaert, line 73:
```suggestion
I -- Yes --> J[hash := murmur3Hash of bucketingPropertyValue divided by Int64.MaxValue]
```

toddbaert (approved):
Looks good to me except one small thing which I think is technically a violation of the spec, here.

Otherwise I have left some nits which I would love your input on. To reduce confusion, I've proposed `bucketing property` instead of `target property` since the latter has a bit too much overlap with `targetinKey` and they aren't exactly the same thing. I hope that make sense :sweat_smile: 

Thanks again for your help @craigpastro

core/pkg/eval/fractional_evaluation.go
craigpastro: Yes, this is better. Updated. Thank you!

core/pkg/eval/fractional_evaluation.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/twmb/murmur3"
 )
 
+const targetingKeyKey = "targetingKey"
```
Kavindu-Dodan, line 12: Nit - this fits to `json_evaluator.go` const section (with a comment to explain the purpose). This should be re-usable for other evaluators if neede.

Kavindu-Dodan (approved):
Looks good (left a suggestion for improvement)

And I don't see this as a breaking change, so we don't have to mark the PR as a breaking change 🤔  This introduce a behavior change but that's from an invalid fractional evaluation to `targetingKey` as a fallback

core/pkg/eval/fractional_evaluation.go
craigpastro: Good point. Thank you! Updated and comment added.

toddbaert:
@craigpastro I'm really bummed but the changes seem to have broken the nice spread on your friends-character emails :weary: :weary: :weary:

toddbaert:
> Looks good (left a suggestion for improvement)
> 
> And I don't see this as a breaking change, so we don't have to mark the PR as a breaking change 🤔 This introduce a behavior change but that's from an invalid fractional evaluation to `targetingKey` as a fallback

Ya, using a "hard-coded" string also seems to still work. So I agree with @Kavindu-Dodan unless I'm missing something.

craigpastro:
> @craigpastro I'm really bummed but the changes seem to have broken the nice spread on your friends-character emails 😩 😩 😩

LOL. No, we got it back by replacing Pheobe with Joey (and RossG with just Ross)... perhaps some deeper meaning here.

craigpastro:
> > Looks good (left a suggestion for improvement)
> > And I don't see this as a breaking change, so we don't have to mark the PR as a breaking change 🤔 This introduce a behavior change but that's from an invalid fractional evaluation to `targetingKey` as a fallback
> 
> Ya, using a "hard-coded" string also seems to still work. So I agree with @Kavindu-Dodan unless I'm missing something.

Oh, yes, this is true. It will no longer extract the value of that hardcoded string from the context though.

craigpastro:
> Thanks again for your help @craigpastro

It's my pleasure! Will definitely look for other opportunities to contribute in the future.

toddbaert:
I'm going to merge this, but cc'ing @bacherfl . Hoping to put this out in a release soon now that  https://github.com/open-feature/flagd/issues/843 seems complete. Let me know if you have concerns.

PR #899, opened 2023-09-08
chore: disable caching on integration tests

toddbaert:
see title.


Kavindu-Dodan (approved):
+1 for disabling cache and making validation strict

PR #913, opened 2023-09-14
fix: use 32bit murmur calculation (64 is not stable)

Kavindu-Dodan:
## This PR

Fixes murmur3 hash calculation to be cross language compatible.

See PR discussion here for more details - https://github.com/open-feature/java-sdk-contrib/pull/434

Depends on flagd-testbed PR - https://github.com/open-feature/flagd-testbed/pull/65


toddbaert (approved):
Thanks @Kavindu-Dodan !

toddbaert:
FYI @craigpastro , this is a small change, but many languages don't support a 64 bit murmur3. It's not in the murmur3 spec, so we're using 32 now which seems to be consistently implemented.

cc @beeme1mr

Issue #867, opened 2023-08-28
Rename "metrics" port to "management" port


toddbaert:
Currently, the "metrics" port configuration controls the port the HTTP readyz/healthz checks, gRPC health check, and metrics.

We should change the name to "management" port to make clear it exposes more than just metrics.

Relates to: https://github.com/open-feature/flagd/pull/863

Be sure to update any related doc.

Saurabhkr952:
I'd like to work on this. Changing "metrics" to "management" port makes sense given its broader functionality. I'm ready to implement the change and update the docs. 
Please guide me on the next steps or any specific requirements for this contribution.

Thank you!

toddbaert:
Thanks @Saurabhkr952 . Once https://github.com/open-feature/flagd/pull/863 is merged, I think it should be a matter of:

- updating the viper config (use `managementPort` and fallback to `metricsPort` if `managementPort` is not defined), and add some kind of deprecation warning if possible
- update all mention of `metrics port / metricsPort / etc` in docs, and mention the deprecation above
- update code / config objects to use `managementPort` instead of `metricsPort`

I think that'll about cover it.

agardnerIT:
Please also document this as a new page on the website (in `/web-docs`) as a Definition of done - so that this doesn't go out without documentation.

Saying that, I don't think we have documentation on this piece yet at all...

craigpastro:
It seems that work on the issue has gotten a bit stale. Does anyone mind if I pick this task up? @Saurabhkr952?

beeme1mr:
Hey @craigpastro, that would be great! I've reassigned the issue to you. Thank you 🙇

craigpastro:
Thanks @beeme1mr! I'll try to get to it soon.

PR #921, opened 2023-09-16
feat: added management port configuration (#867)

Saurabhkr952:
Fixes [#867](https://github.com/open-feature/flagd/issues/867)

-  If 'managementPort' is not specified, it will use 'metricsPort' as the default.
- Removed the reference to 'MetricsPort' and added 'ManagementPort' in the 'core/pkg/runtime/from_config.go' file to align with the updated configuration structure.
- Seeking assistance to incorporate a deprecation warning.

### DIFF
--- a/core/pkg/runtime/from_config.go
+++ b/core/pkg/runtime/from_config.go
@@ -60,7 +60,7 @@ type SourceConfig struct {
 // Config is the configuration structure derived from startup arguments.
 type Config struct {
 	MetricExporter    string
-	MetricsPort       uint16
+	ManagementPort    uint16
 	OtelCollectorURI  string
 	ServiceCertPath   string
 	ServiceKeyPath    string
--- a/flagd/cmd/start.go
+++ b/flagd/cmd/start.go
@@ -19,7 +19,7 @@ const (
 	evaluatorFlagName      = "evaluator"
 	logFormatFlagName      = "log-format"
 	metricsExporter        = "metrics-exporter"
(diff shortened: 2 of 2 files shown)


core/pkg/runtime/from_config.go
```diff
@@ -60,7 +60,7 @@ type SourceConfig struct {
 // Config is the configuration structure derived from startup arguments.
 type Config struct {
 	MetricExporter    string
-	MetricsPort       uint16
+	ManagementPort    uint16
```
Kavindu-Dodan, line 63: You also need to change where this configuration is being used.

For example, this configuration is propagated to the service level. See this usage - https://github.com/open-feature/flagd/blob/main/core/pkg/runtime/from_config.go#L137 . And then you can change the name of that usage too.

toddbaert:
> Once the above is done, we need to release the core module. Then you can follow up on the flagd & flagd-proxy changes (with dependencies updated). This will look like what you already started with flagd/cmd/start.go file [3](https://github.com/open-feature/flagd/pull/921#user-content-fn-3-b3f49ee0a155d75f8fcd9f205116eadc). But this needs to be done from a dedicated PR.

It's true that the modules are published separately, but in the same workspace they use the local version - so all these changes can be made at once (unless I'm quite confused).

beeme1mr:
Hi @Saurabhkr952, is this something you're still interested in working on? If so, do you need any help for us?

beeme1mr:
Closing due to inactivity.

PR #936, opened 2023-09-25
chore: add feature-flagging and references

toddbaert:
* adds "feature-flagging" section
* rolls "OpenFeature Compliance" into this section... I think it makes sense here rather than on its own... I have another PR coming soon that will remove that page and do some related minor restructuring 
* makes distinction between custom operation configuration, and implementation


web-docs/concepts/feature-flagging.md
```diff
@@ -0,0 +1,23 @@
...
+| dynamic configuration                  | Flag definitions from any sync source are monitored for changes, with some syncs supporting near real time updates.                                                                                                      |
+| dynamic (context-sensitive) evaluation | flagd evaluations are context sensitive. Rules can use arbitrary context attributes as inputs for flag evaluation logic.                                                                                                 |
+| fractional evaluation                  | flagd's [fractional](../reference/custom-operations/fractional.md) custom operation supports pseudorandom assignment of flag values.                                                                                                                                   |
+| progressive roll-outs                  | Progressive roll-outs of new features can be accomplished by leveraging the [fractional](../reference/custom-operations/fractional.md) custom operation as well as automation in your build pipeline, SCM, or infrastructure which updates the distribution over time. |
```
toddbaert, line 23: @beeme1mr I think this shows that we really need an improvement here (`$flagd.currentTime`) as we spoke about. This solution is not great.

Issue #851, opened 2023-08-21
feature: add standard set of "ambient" props to context


toddbaert:
We already need to add the `flag_key` to the context to improve out bucketing algorithm, (see [issue](https://github.com/open-feature/flagd/issues/843) and [PR](https://github.com/open-feature/flagd/pull/847)).

We should consider what other properties could be useful, and add them under some sort of namespace to the context. A list could include:

- `flag_key`
- `current_timestamp`

I kept thinking about things like `version` (the version of flagd) or `connection_type` (HTTP vs gRPC), but these would be unreliable and inconsistent between flagd RPC and in-process implementations. We should come up with a list that's relevant to all models, it think.

AHB102:
I am new to open source, can I try to resolve this issue ?

toddbaert:
> I am new to open source, can I try to resolve this issue ?

Definitely. I would start by first monitoring the progress on the [blocking issue](https://github.com/open-feature/flagd/pull/847), trying to understand those changes, and helping with review.

AHB102:
Definitely! I will look into it.

toddbaert:
@AHB102 Are you still interested in doing this? I can assign to you if you'd like. The pre-requisite work is done.

AHB102:
> @AHB102 Are you still interested in doing this? I can assign to you if you'd like. The pre-requisite work is done.

Currently working on something else,I'm sorry I can't work on this.

toddbaert:
> > @AHB102 Are you still interested in doing this? I can assign to you if you'd like. The pre-requisite work is done.
> 
> Currently working on something else,I'm sorry I can't work on this.

I understand, no problem!

craigpastro:
Hi @toddbaert. I can finish this off if that is okay. Are you still okay with the properties listed above?

beeme1mr:
I think we should keep it simple and add more as new use cases emerge. My suggestion would be to start with `timestamp` for now. That would allow us to create time based flag configurations in a constant way.

In my opinion, we don't need to include `current` in the name. We can include a full description in the docs.

craigpastro:
Thank you @beeme1mr. Sounds good. I'll add `timestamp` for now.

craigpastro:
By the way @toddbaert, would you like to close this issue for now and create a new one if you think of other properties to add?

toddbaert:
> By the way @toddbaert, would you like to close this issue for now and create a new one if you think of other properties to add?

[...]

PR #958, opened 2023-10-06
feature: add timestamp to json evaluator

craigpastro:
## This PR

Adds "timestamp" to the json evaluation context.

### Related Issues

Related to #851. I am not sure we want to say that it closes the issue though.

### DIFF
--- a/core/pkg/eval/json_evaluator.go
+++ b/core/pkg/eval/json_evaluator.go
@@ -9,6 +9,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/diegoholiveira/jsonlogic/v3"
 	"github.com/open-feature/flagd/core/pkg/logger"
@@ -38,7 +39,8 @@ const (
 var regBrace *regexp.Regexp
 
 type flagdProperties struct {
-	FlagKey string `json:"flagKey"`
+	FlagKey   string    `json:"flagKey"`
+	Timestamp time.Time `json:"timestamp"`
 }
 
 func init() {
@@ -321,7 +323,8 @@ func (je *JSONEvaluator) evaluateVariant(reqID string, flagKey string, context m
 		}
 
 		context = je.setFlagdProperties(context, flagdProperties{
-			FlagKey: flagKey,
+			FlagKey:   flagKey,
+			Timestamp: time.Now(),
 		})
 
 		b, err := json.Marshal(context)
--- a/core/pkg/eval/json_evaluator_test.go
+++ b/core/pkg/eval/json_evaluator_test.go
@@ -1248,4 +1248,36 @@ func TestFlagdAmbientProperties(t *testing.T) {
 			t.Fatalf("expected %s, got %s", model.TargetingMatchReason, reason)
 		}
 	})
+
+	t.Run("timestampIsInTheContext", func(t *testing.T) {
+		evaluator := eval.NewJSONEvaluator(logger.NewLogger(nil, false), store.NewFlags())
+
+		_, _, err := evaluator.SetState(sync.DataSync{FlagData: `{
+			"flags": {
+				"welcome-banner": {
+					"state": "ENABLED",
+					"variants": {
+						"true": true,
+						"false": false
+					},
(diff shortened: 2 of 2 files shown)


core/pkg/eval/json_evaluator.go
```diff
@@ -321,7 +323,8 @@ func (je *JSONEvaluator) evaluateVariant(reqID string, flagKey string, context m
 		}
 
 		context = je.setFlagdProperties(context, flagdProperties{
-			FlagKey: flagKey,
+			FlagKey:   flagKey,
+			Timestamp: time.Now(),
```
toddbaert, line 327: I think `time.Now()` returns a `Time` type, which is useful in Go, but not very useful in JSONLogic rules.

Consistent with flagd's unix-ish, and bare-bones philosophy, I think we should use `time.Now().Unix()` which will return a unix timestamp. I don't think any more precision than seconds since 1970 is needed - I cant imagine anyone would write a rule that needs to be precise to the nanosecond.

This will allow users to write rules which conceptually say `if $flagd.timestamp > {my-release-unix-timestamp} return {my-variant}`.

cc @beeme1mr @thisthat @Kavindu-Dodan @bacherfl

craigpastro: Excellent point.  Updated to use a unix timestamp.

toddbaert (approved):
Makes sense to me! Thanks!

@Kavindu-Dodan @bacherfl @beeme1mr @thisthat  what do you think about this implementation? And about my points [here](https://github.com/open-feature/flagd/pull/958#discussion_r1350737670)?

I think this is a nice feature, because it can allow for a really simple "timed rollout", especially when combined with the fractional operation.

Kavindu-Dodan (approved):

core/pkg/eval/json_evaluator.go
Kavindu-Dodan: +1 for the timestamp

Kavindu-Dodan:
> Makes sense to me! Thanks!
> 
> @Kavindu-Dodan @bacherfl @beeme1mr @thisthat what do you think about this implementation? And about my points [here](https://github.com/open-feature/flagd/pull/958#discussion_r1350737670)?
> 
> I think this is a nice feature, because it can allow for a really simple "timed rollout", especially when combined with the fractional operation.

Timed rollouts are an interesting feature.  Besides, this can be used combined with other json logic operators to enable flag in time range.

beeme1mr (approved):
Love it, thanks!

As a follow up, can you update the following docs to include information on these `flagd` properties?

- https://github.com/open-feature/flagd/blob/main/docs/reference/flag-definitions.md
- https://github.com/open-feature/flagd/blob/main/docs/reference/specifications/in-process-providers.md

In the flag definitions file, adding a section right before `Shared evaluators` makes sense. It can include a short description that explains that these properties are automagically included in the context. Below that could be a table listing the properties, the version they were introduced, and a short description.

The in-process spec should also include a section that lists flagd properties as a requirement and what their expected behavior is.

bacherfl (approved):

core/pkg/eval/json_evaluator.go
thisthat: Also +1 on my side for unix timestamp 👍

thisthat (approved):

toddbaert:
> Love it, thanks!
> 
> As a follow up, can you update the following docs to include information on these `flagd` properties?
> 
> * https://github.com/open-feature/flagd/blob/main/docs/reference/flag-definitions.md
> * https://github.com/open-feature/flagd/blob/main/docs/reference/specifications/in-process-providers.md
> 
> In the flag definitions file, adding a section right before `Shared evaluators` makes sense. It can include a short description that explains that these properties are automagically included in the context. Below that could be a table listing the properties, the version they were introduced, and a short description.
> 
> The in-process spec should also include a section that lists flagd properties as a requirement and what their expected behavior is.

Agreed on all counts. I didn't want to do this until everyone agreed on this approach, but since it's clear we all do I'll do documentation next (or create an issue for it).

PR #1041, opened 2023-11-30
fix: various edge cases in targeting

toddbaert:
While discussing [this](https://github.com/open-feature/flagd-testbed/issues/84), @beeme1mr and I realized it was worth adding a bit more precision here. Please let me know what you folks think, particularly @Kavindu-Dodan and @bacherfl who helped implement some of these.

The main reason this is important, is I will be creating some "negative" gherkin tests to make sure bad inputs and mis-configurations behave consistently across all flagd engine implementations.

Update: new gherkin is [here](https://github.com/open-feature/flagd-testbed/pulls).


bacherfl (approved):
lgtm

bacherfl:
> While discussing [this](https://github.com/open-feature/flagd-testbed/issues/84), @beeme1mr and I realized it was worth adding a bit more precision here. Please let me know what you folks think, particularly @Kavindu-Dodan and @bacherfl who helped implement some of these.
> 
> The main reason this is important, is I will be creating some "negative" gherkin tests to make sure bad inputs and mis-configurations behave consistently across all flagd engine implementations.

Approved, to me it makes sense to be more specific here, and I agree that for boolean results returning `false` instead of `nil` is the better option

toddbaert:
@bacherfl please re-review when you can.

docs/reference/specifications/custom-operations/string-comparison-operation-spec.md
```diff
@@ -32,7 +32,7 @@ The following flow chart depicts the logic of this evaluator:
 flowchart TD
 A[Parse targetingRule] --> B{Is an array containing exactly two items?};
 B -- Yes --> C{Is targetingRule at index 0 a string?};
-B -- No --> D[Return nil];
+B -- No --> D[Return false];
```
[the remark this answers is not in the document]
Kavindu-Dodan: This is confirmed with the json logic rules - https://jsonlogic.com/truthy.html

Kavindu-Dodan (approved):

Kavindu-Dodan (approved):

docs/reference/specifications/custom-operations/semver-operation-spec.md
```diff
@@ -21,7 +21,7 @@ The 'sem_ver' evaluation rule contains exactly three items:
 1. Target property value: the resolved value of the target property referenced in the targeting rule
 2. Operator: One of the following: `=`, `!=`, `>`, `<`, `>=`, `<=`, `~` (match minor version), `^` (match major version)
 3. Target value: this needs to resolve to a semantic versioning string. If this condition is not met, the evaluator should
-log an appropriate error message and return `nil`
+log an appropriate error message and return `false`
```
beeme1mr, line 24: Should we specific the log level?

beeme1mr (approved):

docs/reference/flag-definitions.md
```diff
@@ -301,6 +318,54 @@ Example:
...
+          "true",
+          "false"
+        ]
+      }
+    }
+  }
+}
+```
+
+can be shorted to this:
```
toddbaert, line 349:
```suggestion
can be shortened to this:
```

docs/reference/specifications/custom-operations/semver-operation-spec.md
toddbaert: I think that's probably too far, I'm not sure they are the same for all languages. It feels like something devs can make their own call on

docs/reference/flag-definitions.md
```diff
@@ -147,10 +153,18 @@ Example of an invalid configuration:
...
 A targeting rule **must** be valid JSON.
 Flagd uses a modified version of [JsonLogic](https://jsonlogic.com/), as well as some custom pre-processing, to evaluate these rules.
-The output of the targeting rule **must** match the name of one of the variants defined above.
-If an invalid or null value is returned by the targeting rule, the `defaultVariant` value is used.
 If no targeting rules are defined, the response reason will always be `STATIC`, this allows for the flag values to be cached, this behavior is described [here](specifications/rpc-providers.md#caching).
 
+#### Variants Returned From Targeting Rules
+
+The output of the targeting rule **must** match the name of one of the defined variants.
+Once exception to the above is that rules may return `true` or `false` which will map to the variant indexed by the equivalent string (`"true"`, `"false"`).
```
beeme1mr, line 161:
```suggestion
One exception to the above is that rules may return `true` or `false` which will map to the variant indexed by the equivalent string (`"true"`, `"false"`).
```

core/pkg/eval/fractional_evaluation_test.go
```diff
@@ -55,13 +55,13 @@ func TestFractionalEvaluation(t *testing.T) {
...
-		expectedVariant string
-		expectedReason  string
-		expectedError   error
+		flags             Flags
+		flagKey           string
+		context           map[string]any
+		expectedValue     string
+		expectedVariant   string
+		expectedReason    string
+		expectedErrorCode string
```
thisthat, line 64: maybe?
```suggestion
		expectedErrorMsg string
```
with code I expect an integer

core/pkg/eval/json_evaluator.go
```diff
@@ -341,21 +341,26 @@ func (je *JSONEvaluator) evaluateVariant(reqID string, flagKey string, context m
...
+		}
+
 		// strip whitespace and quotes from the variant
-		variant = strings.ReplaceAll(strings.TrimSpace(result.String()), "\"", "")
+		variant = strings.ReplaceAll(trimmed, "\"", "")
 
 		// if this is a valid variant, return it
 		if _, ok := flag.Variants[variant]; ok {
 			return variant, flag.Variants, model.TargetingMatchReason, metadata, nil
+		} else {
```
bacherfl: the linter is right, this else can be removed and we can directly do the `return` here

bacherfl (commented):
thanks for the update @toddbaert - conceptually that makes perfect sense. just had some minor suggestions regarding the code, but when those are fixed this PR should be good to go :)

core/pkg/eval/json_evaluator.go
toddbaert: You are somehow commenting on an old diff. You are right that this was a problem, but I fixed it yesterday :thinking:

core/pkg/eval/fractional_evaluation_test.go
toddbaert: These are [openfeature error codes](https://openfeature.dev/specification/types#error-code) (which are particular strings, kinda like `ENOENT` or `EACCES` in unix).

PR #1077, opened 2023-12-12
chore: use client-go library for retrieving FeatureFlag CRs

### LINKED ISSUE #1072: Use client-go library instead of controller-runtime for kubernetes sync implementation.
Use client-go library instead of controller-runtime for kubernetes sync implementation. See background: 

> It seems like this is happening because the kubernetes sync of the flagd-proxy is using the `k8s.io/controller-runtime` client, which is intended to be used in an operator, such as the OF operator. The controller runtime has the requirement of the `log.SetLogger()` function to be called within 30 seconds after the application has been started (see [1]), otherwise the error message will be shown. I just verified this be starting `flagd-proxy`, waiting 30 seconds, and then doing a call to `SyncFlags`, resulting in the message being shown. If `SyncFlags` is called within 30 seconds, no error message is printed.

> To get rid of this error message we would need to make sure the said `log.SetLogger()` function is called during startup of `flagd-proxy`. However, I would recommend to switch to the client go [2] library instead, as this is more intended for this kind of application, whereas using the client from the controller-runtime is a bit of a misuse in this setting.
Switching to the `client-go` library will require making some changes in the usage of the client, but could be done with relatively low effort imho.

Thanks @bacherfl 

References:

[1] - https://github.com/kubernetes-sigs/controller-runtime/blob/1ea2be573f7887a9fbd766e9a921c5af344da6eb/pkg/log/log.go#L78
[2] - https://github.com/kubernetes/client-go

_Originally posted by @bacherfl in https://github.com/open-feature/flagd/issues/1063#issuecomment-1845279927_

bacherfl:
Closes #1072 

This PR removes the usage of the `controller-runtime` client, and instead uses the dynamic client that is already used as a base for the informer.


core/pkg/sync/builder/syncbuilder.go
```diff
@@ -172,27 +169,22 @@ func (sb *SyncBuilder) newGRPC(config sync.SourceConfig, logger *logger.Logger)
...
 
 type IK8sClientBuilder interface {
-	GetK8sClients() (client.Reader, dynamic.Interface, error)
+	GetK8sClients() (dynamic.Interface, error)
 }
 
 type KubernetesClientBuilder struct{}
 
-func (kcb KubernetesClientBuilder) GetK8sClients() (client.Reader, dynamic.Interface, error) {
+func (kcb KubernetesClientBuilder) GetK8sClients() (dynamic.Interface, error) {
```
thisthat, line 177:
```suggestion
func (kcb KubernetesClientBuilder) GetK8sClient() (dynamic.Interface, error) {
```

core/pkg/sync/kubernetes/kubernetes_sync_test.go
```diff
@@ -443,18 +452,13 @@ func TestSync_fetch(t *testing.T) {
...
-				},
-				readClient: &MockClient{
-					getResponse: tt.args.ClientResponse,
-					clientErr:   tt.args.ClientError,
-				},
 				logger: logger.NewLogger(nil, false),
 			}
 
+			if tt.injectionFunc != nil {
+				tt.injectionFunc(k)
```
toddbaert, line 459: I like this functional test/mock setup pattern. I ended up doing it in a few of my tests a while back and was worried it was "not very Go", but it's reassuring to see you've used something similar.

toddbaert (approved):
I tried but couldn't find anything I didn't like :smile:

beeme1mr (approved):

Issue #948, opened 2023-10-04
Improve gRPC service-names, deprecate old services


toddbaert:
We should rename the protos [here](https://github.com/open-feature/schemas/blob/main/protobuf/schema/v1/schema.proto) and [here](https://github.com/open-feature/schemas/blob/main/protobuf/sync/v1/sync_service.proto) to `flagd.evaluation.v1` and `flagd.sync.v1`, respectively.

Why?

gRPC services should be named/namespaced appropriately, particularly for the following reasons:

- collisions [like this](https://stackoverflow.com/questions/63569200/golang-protobufs-has-a-name-conflict-over) can occur when loading protos with the same name; not having a `flagd` namespace increases this likelihood significantly.
- when HTTP traffic is inspected, lack of `flagd` will make it difficult to identify, increasing likelihood of admins blocking traffic, etc
- in browser client usage, reverse proxy rules for `/schema` requests aren't obviously related to flagd; improving the name will make such rules' purpose obvious and prevent clashes 
- semantic names are generally important, and `schema.v1` doesn't say anything about what this RPC/service actually DOES.

How?

In order to reduce breakage, we should support both the old and new RPCs. I'm not sure how best to do this. There may be a way to "proxy" traffic from one to the other. Alternatively, we might have to build and run both protos (so we would pull 2 versions of the protobuff schema).

We would then slowly update providers to use the new protos, and deprecate the old proto in flagd 1.0.

bacherfl:
@toddbaert I would be happy to work on this

toddbaert:
Thanks @bacherfl ! Feel free. Let me know if you need help - it could get a bit hairy.

One think you _may_ need to do is pull 2 versions of the [schemas](https://github.com/open-feature/schemas/tree/main/protobuf) repo (you can check out a tag).

PR #1083, opened 2023-12-15
feat: support new flagd.evaluation and flagd.sync schemas

### LINKED ISSUE #1029: [FEATURE] serve both old and new protos
### Requirements

With [this PR](https://github.com/open-feature/flagd-schemas/pull/109) we added better-named protos and endpoints. We should serve both of these for a while before removing the old ones in 1.0. See all the protos here: https://buf.build/open-feature/flagd

@bacherfl did a POC here: https://github.com/bacherfl/flagd/pull/1

Note this should be done for both the flagd-proxy (sync.proto) and flagd itself (evaluation.proto).

See reasoning for new naming here: https://github.com/open-feature/flagd/issues/948

cc @Kavindu-Dodan @bacherfl

bacherfl:
Closes #1029 

This PR introduces support for the newly introduced evaluation and sync schemas.

Supporting both the old and new schemas involves some duplication, but I tried to keep it as minimal as possible. I'm of course open for suggestions for any ideas on how to make this simpler :)

See reasoning for new naming [here](https://github.com/open-feature/flagd/issues/948).


core/pkg/service/flag-evaluation/flag_evaluator_v2.go
```diff
@@ -0,0 +1,274 @@
...
+type FlagEvaluationServiceV2 struct {
+	logger                *logger.Logger
+	eval                  evaluator.IEvaluator
+	metrics               *telemetry.MetricsRecorder
+	eventingConfiguration *eventingConfiguration
+	flagEvalTracer        trace.Tracer
+}
+
+// NewFlagEvaluationServiceV2 creates a FlagEvaluationService with provided parameters
+func NewFlagEvaluationServiceV2(log *logger.Logger,
```
toddbaert, line 31: This is a nit, but I think it might be better not postfix this with `V2`. I understand this is just a namespacing, but I think we risk some confusion if we ever actually use `flagd.evaluation.v2` (right now this is associated with `flagd.evaluation.v1`.

core/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -45,6 +45,7 @@ func NewFlagEvaluationService(log *logger.Logger,
 	}
 }
 
+// nolint:dupl
```
toddbaert, line 48: Nit: I recommend moving this to the top of the file and adding another comment here saying eventually the whole thing can be removed.

toddbaert:
@bacherfl since you're going through this, would you mind sprinkling references to [this new issue](https://github.com/open-feature/flagd/issues/1088) I created where relevant? That way it'll be super easy to make sure that we remove all the right stuff by just searching for `1088`.

toddbaert:
> Closes #1029
> 
> This PR introduces support for the newly introduced evaluation and sync schemas.
> 
> Supporting both the old and new schemas involves some duplication, but I tried to keep it as minimal as possible. I'm of course open for suggestions for any ideas on how to make this simpler :)

I think the effort we should put into reducing duplication that we know will be deleted should be low, so unless the solutions are very easy, I'm fine with it as is.

core/pkg/service/flag-evaluation/flag_evaluator_v2.go
beeme1mr: Or maybe rename the old evaluation service to something else.

toddbaert: > Or maybe rename the old evaluation service to something else.

Actually that might be better.

toddbaert:
Can you update all e2e tests and documentation to use the new paths? I think a good approach might be to simply search for `schema.v1.Service` and `sync.v1` and replace them with the new paths and RPC names.

core/pkg/service/sync/sync_metrics.go
```diff
@@ -30,7 +30,7 @@ func (s *Server) captureMetrics() error {
 	}
 
 	_, err = meter.RegisterCallback(func(_ context.Context, o api.Observer) error {
-		o.ObserveInt64(syncGuage, s.handler.syncStore.GetActiveSubscriptionsInt64())
+		o.ObserveInt64(syncGuage, s.oldHandler.syncStore.GetActiveSubscriptionsInt64())
```
toddbaert, line 33: I think we'll want to add the sums for both handlers here, or otherwise handle the fact that functionally we want to aggregate metrics from both endpoints.

toddbaert (changes requested):
I have 2 small concerns: [this](https://github.com/open-feature/flagd/pull/1083/files#r1430527730) and [this](https://github.com/open-feature/flagd/pull/1083#discussion_r1430527730).

[This](https://github.com/open-feature/flagd/pull/1083#issuecomment-1861267139) is optional, but would be nice to make the subsequent removal easier later.

bacherfl:
> Manually tested and this works for me.
> 
> I have 2 small concerns: [this](https://github.com/open-feature/flagd/pull/1083/files#r1430527730) and [this](https://github.com/open-feature/flagd/pull/1083#discussion_r1430527730).
> 
> [This](https://github.com/open-feature/flagd/pull/1083#issuecomment-1861267139) is optional, but would be nice to make the subsequent removal easier later.
> 
> We also could consider adding a middleware (or something like that) that would log a one-time warning when the old endpoint is used.

thanks for the review, I will adapt the PR accordingly

core/pkg/service/sync/sync_metrics.go
toddbaert: Ah that makes sense... and I see you're back to using `s.handler` which will work going forward after we remove `oldHandler`. :+1:

toddbaert (approved):
Looks great! Thanks a lot!

beeme1mr (approved):

thisthat (approved):

core/pkg/service/flag-evaluation/flag_evaluator_v2.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	values := s.eval.ResolveAllValues(sCtx, reqID, evalCtx)
+	span.SetAttributes(attribute.Int("feature_flag.count", len(values)))
+	for _, value := range values {
+		// register the impression and reason for each flag evaluated
+		s.metrics.RecordEvaluation(sCtx, value.Error, value.Reason, value.Variant, value.FlagKey)
+		switch v := value.Value.(type) {
+		case bool:
+			res.Flags[value.FlagKey] = &evalV1.AnyFlag{
```
thisthat, line 72: evalV1 in a v2 file reads weirdly but when we will remove the support for the old proto we can rename this file :)

toddbaert:
I'll merge this EOD unless I hear objections.

Issue #1141, opened 2024-01-08
[FEATURE] flagd should use standard OTel SDK env vars


austindrenski:
## Request

Replace `--otel-collector-uri`/`FLAGD_OTEL_COLLECTOR_URI` and `--metrics-exporter`/`FLAGD_METRICS_EXPORTER` with well-known OTel SDK configuration environment variables.

- https://opentelemetry.io/docs/concepts/sdk-configuration/
  - https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/
  - https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/ 

```diff
services:
  flagd:
    command:
    - start
    - --debug
    - --uri
    - https://raw.githubusercontent.com/open-feature/flagd/main/docs/assets/demo.flagd.json
    depends_on:
      otel:
        condition: service_started
    environment:
-     FLAGD_METRICS_EXPORTER: otel
-     FLAGD_OTEL_COLLECTOR_URI: otel:4317
+     OTEL_EXPORTER_OTLP_ENDPOINT: http://otel:4318
+     OTEL_EXPORTER_OTLP_PROTOCOL: http/protobuf
+     OTEL_METRICS_EXPORTER: otlp
    image: ghcr.io/open-feature/flagd
    ports:
    - 8013:8013
    - 8014:8014

  otel:
    image: ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector
    ports:
    - 4317:4317
    - 4318:4318
```

IMO this approach would provide a better telemetry story overall for `flagd` since one of the main tenants of the OTel project is to avoid re-inventing the wheel on configuration, but more specifically, I'm opening this issue after running into a pinch-point caused by `flagd` making an opinionated decision about `OTEL_EXPORTER_OTLP_PROTOCOL=grpc`.

In my specific use case, supporting the standard OTel SDK env vars would mean that I could, for example, set `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` in order to communicate with an existing OTel Collector gateway cluster that doesn't accept gRPC traffic.

_(I could, of course, work around this by making my existing `otelcol` gateway cluster listen for gRPC traffic too, but (1) there are genuine, non-trivial DevSecOps considerations involved, and more importantly, (2) I don't think `flagd` __intended__ to be opinionated about this, so hoping there's an appetite for revisiting the current setup.)_

beeme1mr:
[...]

PR #1142, opened 2024-01-08
refactor: Use standard OTel SDK env vars

austindrenski:
- https://opentelemetry.io/docs/concepts/sdk-configuration/
  - https://opentelemetry.io/docs/concepts/sdk-configuration/general-sdk-configuration/
  - https://opentelemetry.io/docs/concepts/sdk-configuration/otlp-exporter-configuration/

Closes: #1141

___

> [!NOTE]
>
> _Fully aware that this neither sufficient nor suitable as is, but opening as a companion to #1141 to provide a starting point for discussion/consideration._


austindrenski:
_Full transparency, I'm still relatively new to golang, so please take everything I've written here with a grain of salt._

That said, I think this is in a decent place for a first round of reviews.

Areas of concern/in need of feedback:
- Should telemetry wire-up _ever_ return errors?
  - OTel SDKs tend to err on the side of `never break the app no matter what`
  - Existing behavior returns errors in some cases.
  - But on some new code paths I've opted to log debug messages instead of returning errors, based on the rationale that if we're deferring to the OTel SDK for wire-up, then we should defer to its style of graceful failure too.
  - Mostly asking this question for my edification in the event that this is something the project already has a strong opinion on.
- The tests are passing...but I also may have broken the intent of one or more of the tests.
  - Some of the tests were related to the above `should we error` question, so just looking for feedback here on what the project/org expects in terms of coverage/quality in tests (on both fronts, I expect what is currently here will probably need some work).
- Do we/anyone have a well-defined/formal integration environment? Are the unit tests that environment? How would we know if I've borked something subtle here?

Anyways, please have at it and let me know what you think 🙇 

_(and again, I'm still new to golang, so if you see garbage code, please call it out for being garbage code, rather get the feedback sooner rather than later lol)_

beeme1mr:
I think this is a good idea, but I would prefer to make this a non-breaking change. We can mark the arguments as deprecated and completely remove them at a later point. I think we can make the existing arguments and environment variables into the official OTel environment variables.

austindrenski:
FYI waylaid by some other priorities, but this is still on my radar.

beeme1mr:
I support this idea but closing this PR due to inactivity.

PR #1152, opened 2024-01-17
feat(core): support any auth scheme in HTTP-sync auth header

### LINKED ISSUE #1150: [FEATURE] Support any auth scheme in HTTP-sync auth header
### Requirements

Currently, we only support the `Bearer` scheme (see code [here](https://github.com/open-feature/flagd/blob/2adfa573a29692d293b37c52b3daf048fd5850d6/core/pkg/sync/http/http_sync.go#L124)). We should support [any scheme](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes).

We can do this by simply letting authors specify the entire contents of the Auth header.

Tasks:

- [ ] deprecate [BasicAuth](https://github.com/open-feature/flagd/blob/2adfa573a29692d293b37c52b3daf048fd5850d6/core/pkg/sync/isync.go#L68) option (but continue to support it) so this isn't a breaking change
- [ ] add new AuthHeader option, which defines entire auth header as an alternative to `BasicAuth` and send in HTTP request
- [ ] document new option, and interaction with deprecated `basicAuth` option "dealer's choice" here: 
  - throw an error if both are defined?
  - default to ignore one if the other is defined?
- [ ] tests

Olunusib:
## This PR

- adds support for any auth scheme in HTTP-sync auth header

### Related Issues

Closes #1150

### DIFF
--- a/core/pkg/sync/http/http_sync.go
+++ b/core/pkg/sync/http/http_sync.go
@@ -49,7 +49,7 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
 
 func (hs *Sync) Init(_ context.Context) error {
 	if hs.BearerToken != "" {
-		hs.Logger.Info("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
+		hs.Logger.Warn("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
 	}
 	return nil
 }
--- a/core/pkg/sync/http/http_sync.go
+++ b/core/pkg/sync/http/http_sync.go
@@ -21,6 +21,7 @@ type Sync struct {
 	LastBodySHA string
 	Logger      *logger.Logger
 	BearerToken string
+	AuthHeader  string
 	Interval    uint32
 	ready       bool
 }
@@ -47,7 +48,9 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
 }
 
 func (hs *Sync) Init(_ context.Context) error {
-	// noop
+	if hs.BearerToken != "" {
+		hs.Logger.Info("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
+	}
 	return nil
 }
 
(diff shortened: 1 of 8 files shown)


core/pkg/sync/http/http_sync.go
```diff
@@ -47,7 +48,9 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
 }
 
 func (hs *Sync) Init(_ context.Context) error {
-	// noop
+	if hs.BearerToken != "" {
+		hs.Logger.Info("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
+	}
```
Olunusib, line 53: This alert regarding `bearerToken` might be redundant since the behavior is already documented. Open to not adding it

core/pkg/sync/http/http_sync.go
```diff
@@ -47,7 +48,9 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
 }
 
 func (hs *Sync) Init(_ context.Context) error {
-	// noop
+	if hs.BearerToken != "" {
+		hs.Logger.Info("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
```
toddbaert, line 52:
```suggestion
		hs.Logger.Warn("Deprecation Alert: bearerToken option is deprecated, please use authHeader instead")
```

docs/reference/sync-configuration.md
```diff
@@ -27,20 +27,26 @@ The flagd accepts a string argument, which should be a JSON representation of an
...
+| tls         | optional `boolean` | Enable/Disable secure TLS connectivity. Currently used only by gRPC sync. Default (ex: if unset) is false, which will use an insecure connection                              |
+| providerID  | optional `string`  | Value binds to grpc connection's providerID field. gRPC server implementations may use this to identify connecting flagd instance                                             |
+| selector    | optional `string`  | Value binds to grpc connection's selector field. gRPC server implementations may use this to filter flag configurations                                                       |
+| certPath    | optional `string`  | Used for grpcs sync when TLS certificate is needed. If not provided, system certificates will be used for TLS connection                                                      |
+
+### Important Notes
+
+- **Mutual Exclusivity**: `authHeader` and `bearerToken` cannot be used simultaneously. Setting both fields in the configuration
+  results in an error, preventing the application from starting.
```
toddbaert, line 45: Nit: this is fine, but I think another option would do just mention in each they are incompatible with each other, respectively, for example:

`Used for http sync; set this to include the complete `Authorization` header value for any authentication scheme (e.g., "Bearer token_here", "Basic base64_credentials", etc.). Cannot be used with `bearerToken`.

toddbaert (approved):
Looks great! Thanks a lot!

docs/reference/sync-configuration.md
Olunusib: Added

thisthat (approved):
LGTM 👍

PR #1174, opened 2024-01-29
docs: move OFO doc references

### LINKED ISSUE #1055: [DOC] featureFlagSpec: > flagSpec
### Change in the documentation

https://flagd.dev/reference/openfeature-operator/crds/featureflag/ mentions `featureFlagSpec` but I believe is deprecated in v1beta1 and should be `flagSpec` instead

beeme1mr:
## This PR

- adds redirects to appropriate OFO docs
- adds an OFO overview page

### Related Issues

Fixes #1055

### Notes

This PR removes most of the OFO docs so that we only have to maintain content in a single location. In the future, it would be ideal if we could include relative content automatically in the flagd docs, but that was out of scope of this PR.


Kavindu-Dodan (approved):
I think this is a good move. Maintaining guides for OFO at flagd make them to quickly outdate

toddbaert (approved):

PR #1180, opened 2024-01-30
docs: update playground flagd version, add new scenarios

beeme1mr:
## This PR

- adds a shared evaluator scenario
- adds a boolean shorthand scenario
- updates flagd core the latest release

## Demo

https://deploy-preview-1180--polite-licorice-3db33c.netlify.app/playground/


playground-app/src/scenarios/share-evaluators.ts
```diff
@@ -0,0 +1,46 @@
+import type { Scenario } from "../types";
+import { contextToPrettyJson, featureDefinitionToPrettyJson } from "../utils";
+
+export const sharedEvaluator: Scenario = {
```
toddbaert, line 4: I'd recommend changing the filename to match this export ("d") is missing.

playground-app/src/scenarios/index.ts
```diff
@@ -21,6 +23,8 @@ export const scenarios = {
   "Enable based on the current time": enableByTime,
   "Multi-variant experiment": pseudoRandomSplit,
   "Progressive rollout": progressRollout,
+  "Shared evaluators": sharedEvaluator,
+  "Boolean shorthand": booleanShorthand,
```
toddbaert, line 27: Not necessary, but I wonder if this is better directly after the boolean flag (2nd in the list).

toddbaert (approved):
Optional comments.

playground-app/src/scenarios/share-evaluators.ts
beeme1mr: I can do that but I'm assuming you mean "s".

Kavindu-Dodan (approved):

playground-app/src/scenarios/share-evaluators.ts
Kavindu-Dodan: `shared-evaluators.ts` I think this is the file name suggestion

playground-app/src/scenarios/index.ts
beeme1mr: I'm going to leave the order as is because it matches how we introduce these topics in the [flag definition docs](https://flagd.dev/reference/flag-definitions/).

PR #1188, opened 2024-01-31
chore: update test dependencies, fix for otel api change and update renovate configuration

Kavindu-Dodan:
## This PR

Update test dependencies and add specific paths to renovate configurations so that renovate can take care of test module updates.

Also, update to match connect API change - https://github.com/connectrpc/otelconnect-go/releases/tag/v0.7.0 

Note - `test/**` was ignored as this comes from default preset [^1] [^2]

[^1]: https://docs.renovatebot.com/presets-default/
[^2]: https://docs.renovatebot.com/presets-default/#ignoremodulesandtests


Kavindu-Dodan:
Depends on https://github.com/open-feature/go-sdk-contrib/pull/442

toddbaert (approved):

core/pkg/runtime/from_config.go
```diff
@@ -82,6 +82,11 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
 		return nil, err
 	}
 
+	options, err := telemetry.BuildConnectOptions(telCfg)
```
thisthat, line 85: IMHO we should swallow the error, warn the user, and proceed anyhow. Telemetry is nice to have but not a blocker to operate flagd

Kavindu-Dodan: Yeag this is a fair point 🤔  I will choose to log and continue then

Kavindu-Dodan: hmm I checked other places where we have telemetry errors (specially on set-up phase), and strategy is to fail when there're errors (see - https://github.com/open-feature/flagd/blob/main/core/pkg/runtime/from_config.go#L47-L57).

If I change this now, we will be breaking this pattern. So I think this is something to discuss and change in a future iteration 🤔 

cc - @beeme1mr & @toddbaert

toddbaert: Agreed. If @thisthat is right we should make the change in a different PR and be consistent. You can create an investigation issue if you want.

Kavindu-Dodan: I created a follow-up tasl - https://github.com/open-feature/flagd/issues/1194

@thisthat If you are satisfied, please resolve this conversation and approve the PR :)

thisthat (approved):

Issue #1207, opened 2024-02-12
[FEATURE] Upgrade to go version 1.21


Kavindu-Dodan:
### Requirements

Go release 1.22 is out this month (02/2024), and this makes flagd's current go version 1.20 to unsupported version [^1]

**AC**

- Upgrade flagd's CI, deps to Go version 1.20
- Make sure to follow up on In-process provider as we rely on Core module there 

[^1]: https://endoflife.date/go

PR #1208, opened 2024-02-13
chore: bump go to 1.21

odubajDT:
Let's align with the latest Go release

Fixes https://github.com/open-feature/flagd/issues/1207


core/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/open-feature/flagd/core
 
-go 1.20
+go 1.21
```
toddbaert, line 3: My only concern here is that the flagd go provider uses this module as a library, and it's using 1.20: https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/go.mod

odubajDT: should be also fine if we merge this as well https://github.com/open-feature/go-sdk-contrib/pull/452. 1.21 should be fully backwards compatible with 1.20

Kavindu-Dodan: I think we should not upgrade all providers in the contrib repository. But we can upgrade flagd provider independently and there are providers already using Go 1.21 (hence the pipeline using the latest Go version). 

Similarly, we should upgrade Go-SDK gradually as it's a central piece of OF. I know Go have a backward compatibility promise, but once upgraded and if we use a newer API, then it can break users who are yet to upgrade their Go version. This could happen for enterprise customers where upgrades are slow and timeconsuming compared to Go's 6 month release cycle :)

odubajDT: Good point, I wasn't aware that some of the users are using parts of the flagd code as a library.

This PR might be then a candidate for 1.0 stable version. At least I would vote for aligning all versions to 1.21.

Kavindu-Dodan: @odubajDT it's actually our flagd go provider [1] that depends on the core. 

Personally, I am fine with moving ahead with Go 1.21 for flagd. And then we can bump the go provider of flagd itself. But I would avoid changing other providers for sometime :) I hope this is clearer now? 

[1] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/go.mod#L15

Kavindu-Dodan: ohh, and please check for API changes if any in the PR. If there's any, it might be good to mark the PR as a breaking change as we loose forward compatibility for users

Kavindu-Dodan:
Closing as this is outdated

PR #1214, opened 2024-02-15
feat!: new proto (flagd.sync.v1) for sync sources

RealAnna:
## This PR
- update mockgen
- update proto in core
- update to go 1.21


core/pkg/sync/grpc/grpc_sync_test.go
```diff
@@ -677,3 +661,14 @@ func (b *bufferedServer) SyncFlags(_ *v1.SyncFlagsRequest, stream syncv1grpc.Fla
 func (b *bufferedServer) FetchAllFlags(_ context.Context, _ *v1.FetchAllFlagsRequest) (*v1.FetchAllFlagsResponse, error) {
 	return b.fetchAllFlagsResponse, b.fetchAllFlagsError
 }
+
+func (b *bufferedServer) GetMetadata(_ context.Context, _ *v1.GetMetadataRequest) (*v1.GetMetadataResponse, error) {
+	return &v1.GetMetadataResponse{
```
RealAnna, line 666: not sure what would be meaningful here

core/pkg/sync/grpc/grpc_sync.go
```diff
@@ -169,55 +169,56 @@ func (g *Sync) connectWithRetry(
 // handleFlagSync wraps the stream listening and push updates through dataSync channel
 func (g *Sync) handleFlagSync(stream syncv1grpc.FlagSyncService_SyncFlagsClient, dataSync chan<- sync.DataSync) error {
```
RealAnna, line 170: what to do with this one now?

core/pkg/sync/grpc/grpc_sync_test.go
toddbaert: I think we can leave it empty for now.
Eventually we will decide how to populate it.

toddbaert: @Kavindu-Dodan @RealAnna I've pushed an update which uses a slightly new (breaking) proto. It changes the `metadata` type to a Struct, which is optional, so now the `GetMetadataResponse` is completely empty.

toddbaert:
@RealAnna @Kavindu-Dodan I've marked this as breaking since it requires sync-sources to implement this new proto.

Kavindu-Dodan (approved):
Looks good 

@odubajDT  this will conflict with your 1.21 though, please check

Once merged, we need to release the core. Then we can update flagd & flagd-proxy with the core change of protocs :)

core/pkg/sync/grpc/grpc_sync_test.go
Kavindu-Dodan: Thanks @toddbaert  Right now we do not expose any metadata, but this is a good move :)

toddbaert (approved):

Issue #1230, opened 2024-02-27
[FEATURE] Support serving sync.proto


toddbaert:
## Problem

Currently, flagd doesn't support the [flagd.sync.proto](https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1). This means that flagd cannot directly serve [in-process providers](https://flagd.dev/reference/specifications/in-process-providers/). Users can still use in process providers by implementing a server that serves this proto, which is great if they are interested in building a custom control plane, perhaps with a UI. However, it's perfectly reasonable for users to want to use flagd itself with in-process providers.

## Proposal

Support serving the [flagd.sync.proto](https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1) in flagd. When a sync client connects, it can use its `selector` field to select a source flagd is already watching (an http endpoint, file, or CR instance). If no `selector` is specified, the client will get a merged representation of all sources flagd is watching (using the same [merge logic](https://flagd.dev/concepts/syncs/#merging) as flagd). Note that unlike the flagd proxy, dynamically watching additional sources would not be supported. The sync endpoint would run on a separate port from the evaluation endpoint.

## Benefits

- support for in-process providers by flagd (https://github.com/open-feature/flagd/issues/961#issuecomment-1858994446)
  - in-process providers completely eliminate I/O for each flag evaluation, bringing significant latency improvements
  - in-process providers cache a ruleset, which can be used even if flagd is unavailable, which results in improved fault tolerance
- we can completely delete the [sync server stub](https://github.com/open-feature/flagd-testbed/tree/main/sync) used in e2e tests

## Tasks

- [x] implement [flagd.sync.proto](https://buf.build/open-feature/flagd/docs/main:flagd.sync.v1) server in flagd
  - [x] support use of `selector` as described above to select source (or merge all source if not supplied)
  - [x] run on separate port from evaluation.proto
  - [ ] use connect protocol, so support both web and server clients (optional, but suggested)
- [x] associated testing
- [x] associated documentation

Kavindu-Dodan:
I will start working on this

toddbaert:
> I will start working on this

Amazing... let me know if you have any questions or want to meet up to think through anything.

One thing I'm so far not sure on is the use of connect... if it costs nothing, I'm fine with it... but if it's adds a lot of complexity we can re-evaluate cc @beeme1mr .

Kavindu-Dodan:
Update - 01.03.2024

- POC on buf connect works fine
[...]

PR #1234, opened 2024-03-01
chore: move packaging & isolate service implementations 

Kavindu-Dodan:
## This PR

**Background**

This is an attempt to move services and wiring logic to their end usage.

For example,  having `runtime` package in the core module did not provide any advantage. This package was only used by `flagd` and contained the logic that binds configurations to service implementations. 

Similarly, `flagd-proxy` exposed `sync` service, and having that inside the core module seemed out of place. 

**What's changed**

I have moved `runtime` package to `flagd` & specific service implementations (flag evaluation and sync service) to their end usage. 

**Why ?**

I considered this to be a pre-requisite for https://github.com/open-feature/flagd/issues/1230. If I attempted to add another `sync` implementation to the current core module, it would have made the whole packaging confusing (ex:- we have two sync proto implementations, which do not serve the same purpose. One for flagd-proxy and the other for flagd).

After this refactoring, I could add the flagd's sync service, where it serve buf connect and I can locate that within flagd. 

Besides, we still have a core module with all the sharables. 

Consider the current packaging in the below diagram,

image

* nore - core sync is for flag data sync listeners such as files, K8s ...


test/integration/go.mod
```diff
@@ -12,60 +12,104 @@ require (
 )
 
 require (
-	buf.build/gen/go/open-feature/flagd/connectrpc/go v1.12.0-20231031123731-ac2ec0f39838.1 // indirect
-	buf.build/gen/go/open-feature/flagd/grpc/go v1.3.0-20231031123731-ac2ec0f39838.2 // indirect
-	buf.build/gen/go/open-feature/flagd/protocolbuffers/go v1.31.0-20231031123731-ac2ec0f39838.2 // indirect
-	connectrpc.com/connect v1.14.0 // indirect
+	buf.build/gen/go/open-feature/flagd/connectrpc/go v1.15.0-20240215170432-1e611e2999cc.1 // indirect
```
toddbaert, line 15: optional, but if we're doing big renames/refactors, can we perhaps also change this to `tests.e2e`? These are really e2e tests. You will also need to change the makefile and CI commands: https://github.com/open-feature/flagd/blob/517973fa84f0faaa88c12ce06285b30e655860d9/Makefile#L47-L48

toddbaert (approved):
Yes I think this makes sense. Left an optional recommendation for you to consider.

test/integration/go.mod
toddbaert: On second thought, maybe we should do this in a PR, so test changes are isolated from refactors, for increase confidence.

Kavindu-Dodan: Yes, i think I will defer this to another PR :) This PR only focuses on moving sources to repackage

odubajDT (approved):

thisthat (approved):
Well done 👍

Kavindu-Dodan:
Thank you for the reviews @toddbaert @odubajDT @thisthat , I will merge this now :)

PR #1237, opened 2024-03-04
feat: serve sync.proto on port 8015

Kavindu-Dodan:
## This PR

Introduce flag sync capability to flagd as discussed at [1] and fixes https://github.com/open-feature/flagd/issues/1230 .

**What's changed ?**

The change included with this PR introduces the gRPC sync contract [2]. This allows flagd to expose its store to in-process provider which consumes gRPC flag stream. 

image

**How to use ?**

flagd will start sync service on port `8015`. You can alter the default sync port by providing desired port to the startup flag `syncPort` (`--sync-port=8686`) 

**Implementation details**

- Sync service startup is delayed to allow configured flag sources to complete their initial loading (dealy is 5 seconds)
- Sync request's `selector` can be used to specify the specific flag source. If unset, all flags will be sent
- A new stream response will be created whenever there are updates from flag sources 

[1] - https://github.com/open-feature/flagd/discussions/1153 
[2] - https://github.com/open-feature/flagd-schemas/blob/main/protobuf/flagd/sync/v1/sync.proto

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -37,7 +37,7 @@ In-process deployments embed the flagd evaluation engine directly into the clien
 The in-process provider is connected via the sync protocol to an implementing [gRPC service](./concepts/syncs.md#grpc-sync) that provides the flag definitions.
 You can use flagd as a [gRPC sync service](./reference/grpc-sync-service.md).
 In this mode, the flag sync stream will expose aggregated flag configurations currently configured through [syncs](./concepts/syncs.md).
-This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead, since no inter-process communication is required.
+This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead for flag evaluations, since no inter-process communication is required.
 
 ```mermaid
 ---
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -35,7 +35,9 @@ erDiagram
 
 In-process deployments embed the flagd evaluation engine directly into the client application through the use of an [in-process provider](./installation.md#in-process).
 The in-process provider is connected via the sync protocol to an implementing [gRPC service](./concepts/syncs.md#grpc-sync) that provides the flag definitions.
-This pattern requires an in-process implementation of the flagd evaluation engine, but has the benefit of no I/O overhead, since no inter-process communication is required.
+You can use flagd itself as a [gRPC service](./reference/grpc-sync-service.md).
+In this mode, the flag sync stream will expose aggregated flag configurations currently configured through [syncs](./concepts/syncs.md).
+This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead, since no inter-process communication is required.
 
 ```mermaid
 ---
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -35,7 +35,7 @@ erDiagram
 
 In-process deployments embed the flagd evaluation engine directly into the client application through the use of an [in-process provider](./installation.md#in-process).
 The in-process provider is connected via the sync protocol to an implementing [gRPC service](./concepts/syncs.md#grpc-sync) that provides the flag definitions.
-You can use flagd itself as a [gRPC service](./reference/grpc-sync-service.md).
+You can use flagd as a [gRPC sync service](./reference/grpc-sync-service.md).
 In this mode, the flag sync stream will expose aggregated flag configurations currently configured through [syncs](./concepts/syncs.md).
 This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead, since no inter-process communication is required.
 
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -5,7 +5,6 @@ description: flagd as a gRPC sync service
 # Overview
 
 flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag definitions.
-This mode is **disabled** by default, and you can enable it by using startup flag `--sync-enabled` (or `-e` shorthand flag).
 The gRPC sync stream contains flag definitions currently configured at flagd as [sync-configurations](./sync-configuration.md).
 
 ```mermaid
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -22,7 +22,7 @@ erDiagram
 
 You may change the default port of the service using startup flag `--sync-port` (or `-g` shothand flag).
 
-By default, the gRPC stream expose all the flag configurations merged following merge strategy of flagd.
+By default, the gRPC stream exposes all the flag configurations, with conflicting flag keys merged following flag's standard merge strategy.
 You can read more about the merge strategy in our dedicated [concepts guide on syncs](../concepts/syncs.md).
 
 If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector valur to a source and stream specific flags.
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -25,7 +25,7 @@ You may change the default port of the service using startup flag `--sync-port` 
 By default, the gRPC stream exposes all the flag configurations, with conflicting flag keys merged following flag's standard merge strategy.
 You can read more about the merge strategy in our dedicated [concepts guide on syncs](../concepts/syncs.md).
 
-If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector valur to a source and stream specific flags.
+If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector value to a source, and stream just the flags identified in that source.
 For example, if `selector` is set to `myFlags.json`, service will stream flags observed from `myFlags.json` file.
 And the request will fail if there is no flag source matching the requested `selector`.
 flagd provider implementations expose the ability to define the `selector` value.
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -4,7 +4,7 @@ description: flagd as a gRPC sync service
 
 # Overview
 
-flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag configurations.
+flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag definitions.
 This mode is **disabled** by default, and you can enable it by using startup flag `--sync-enabled` (or `-e` shorthand flag).
 The gRPC sync stream contains flag configurations currently configured at flagd as [sync-configurations](./sync-configuration.md).
 
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -27,8 +27,10 @@ You can read more about the merge strategy in our dedicated [concepts guide on s
 
 If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector value to a source, and stream just the flags identified in that source.
 For example, if `selector` is set to `myFlags.json`, service will stream flags observed from `myFlags.json` file.
+Note that, to observe flags from `myFlags.json` file, you may use startup option `uri` like `--uri myFlags.json` or `source` option `--sources='[{"uri":"myFlags.json", provider":"file"}]`.
 And the request will fail if there is no flag source matching the requested `selector`.
-flagd provider implementations expose the ability to define the `selector` value.
+
+flagd provider implementations expose the ability to define the `selector` value. Please consider below example for Java,
 
 ```java
 final FlagdProvider flagdProvider =
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -6,7 +6,7 @@ description: flagd as a gRPC sync service
 
 flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag definitions.
 This mode is **disabled** by default, and you can enable it by using startup flag `--sync-enabled` (or `-e` shorthand flag).
-The gRPC sync stream contains flag configurations currently configured at flagd as [sync-configurations](./sync-configuration.md).
+The gRPC sync stream contains flag definitions currently configured at flagd as [sync-configurations](./sync-configuration.md).
 
 ```mermaid
 ---
--- a/docs/reference/grpc-sync-service.md
+++ b/docs/reference/grpc-sync-service.md
@@ -1,0 +1,41 @@
+---
+description: flagd as a gRPC sync service
+---
+
+# Overview
+
+flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag configurations.
+This mode is **disabled** by default, and you can enable it by using startup flag `--sync-enabled` (or `-e` shorthand flag).
+The gRPC sync stream contains flag configurations currently configured at flagd as [sync-configurations](./sync-configuration.md).
+
+```mermaid
+---
+title: gRPC sync
+---
+erDiagram
+    flagd ||--o{ "sync (file)" : watches
+    flagd ||--o{ "sync (http)" : polls
+    flagd ||--o{ "sync (grpc)" : "sync.proto (gRPC/stream)"
+    flagd ||--o{ "sync (kubernetes)" : watches
+    "In-Process provider" ||--|| flagd : "gRPC sync stream (default port 8015)"
+```
+
+You may change the default port of the service using startup flag `--sync-port` (or `-g` shothand flag).
+
+By default, the gRPC stream expose all the flag configurations merged following merge strategy of flagd.
+You can read more about the merge strategy in our dedicated [concepts guide on syncs](../concepts/syncs.md).
+
+If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector valur to a source and stream specific flags.
+For example, if `selector` is set to `myFlags.json`, service will stream flags observed from `myFlags.json` file.
+And the request will fail if there is no flag source matching the requested `selector`.
+flagd provider implementations expose the ability to define the `selector` value.
+
+```java
+final FlagdProvider flagdProvider =
+        new FlagdProvider(FlagdOptions.builder()
+                .resolverType(Config.Evaluator.IN_PROCESS)
+                .host("localhost")
+                .port(8015)
+                .selector("myFlags.json")
+                .build());
+```
--- a/flagd/cmd/start.go
+++ b/flagd/cmd/start.go
@@ -26,7 +26,6 @@ const (
 	serverKeyPathFlagName  = "server-key-path"
 	socketPathFlagName     = "socket-path"
 	sourcesFlagName        = "sources"
-	syncEnabledFlagName    = "sync-enabled"
 	syncPortFlagName       = "sync-port"
 	uriFlagName            = "uri"
 )
@@ -67,9 +66,6 @@ func init() {
 	flags.StringP(otelCollectorURI, "o", "", "Set the grpc URI of the OpenTelemetry collector "+
 		"for flagd runtime. If unset, the collector setup will be ignored and traces will not be exported.")
 
-	flags.BoolP(syncEnabledFlagName, "e", false, "Enables the gRPC sync service from flagd. "+
-		"This is disabled by default")
-
 	_ = viper.BindPFlag(corsFlagName, flags.Lookup(corsFlagName))
 	_ = viper.BindPFlag(logFormatFlagName, flags.Lookup(logFormatFlagName))
 	_ = viper.BindPFlag(metricsExporter, flags.Lookup(metricsExporter))
@@ -82,7 +78,6 @@ func init() {
 	_ = viper.BindPFlag(sourcesFlagName, flags.Lookup(sourcesFlagName))
 	_ = viper.BindPFlag(uriFlagName, flags.Lookup(uriFlagName))
 	_ = viper.BindPFlag(syncPortFlagName, flags.Lookup(syncPortFlagName))
-	_ = viper.BindPFlag(syncEnabledFlagName, flags.Lookup(syncEnabledFlagName))
 }
 
 // startCmd represents the start command
@@ -138,7 +133,6 @@ var startCmd = &cobra.Command{
 			ServicePort:       viper.GetUint16(portFlagName),
 			ServiceSocketPath: viper.GetString(socketPathFlagName),
 			SyncServicePort:   viper.GetUint16(syncPortFlagName),
-			WithSyncService:   viper.GetBool(syncEnabledFlagName),
 			SyncProviders:     syncProviders,
 		})
 		if err != nil {
--- a/flagd/cmd/start.go
+++ b/flagd/cmd/start.go
@@ -67,7 +67,7 @@ func init() {
 	flags.StringP(otelCollectorURI, "o", "", "Set the grpc URI of the OpenTelemetry collector "+
 		"for flagd runtime. If unset, the collector setup will be ignored and traces will not be exported.")
 
-	flags.BoolP(syncEnabledFlagName, "e", false, "Set true to enable gRPC sync service from flagd. "+
+	flags.BoolP(syncEnabledFlagName, "e", false, "Enables the gRPC sync service from flagd. "+
 		"This is disabled by default")
 
 	_ = viper.BindPFlag(corsFlagName, flags.Lookup(corsFlagName))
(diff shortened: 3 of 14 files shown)


flagd/cmd/start.go
```diff
@@ -18,16 +18,17 @@ import (
...
-	metricsExporter        = "metrics-exporter"
 	managementPortFlagName = "management-port"
+	metricsExporter        = "metrics-exporter"
 	otelCollectorURI       = "otel-collector-uri"
 	portFlagName           = "port"
 	serverCertPathFlagName = "server-cert-path"
 	serverKeyPathFlagName  = "server-key-path"
 	socketPathFlagName     = "socket-path"
 	sourcesFlagName        = "sources"
+	syncEnabledFlagName    = "syncEnabled"
```
bacherfl, line 29: should we change this to `sync-enabled` and `sync-port`, to stay in line with the other flags which are all in kebap-case?

flagd/pkg/runtime/from_config.go
```diff
@@ -12,6 +12,7 @@ import (
 	syncbuilder "github.com/open-feature/flagd/core/pkg/sync/builder"
 	"github.com/open-feature/flagd/core/pkg/telemetry"
 	flageval "github.com/open-feature/flagd/flagd/pkg/service/flag-evaluation"
+	flag_sync "github.com/open-feature/flagd/flagd/pkg/service/flag-sync"
```
bacherfl, line 15: nitpick, to use the recommended [naming scheme](https://go.dev/blog/package-names) : 
```suggestion
	flagsync "github.com/open-feature/flagd/flagd/pkg/service/flag-sync"
```

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -0,0 +1,193 @@
...
+		from = r.subs
+	} else {
+		from = r.selectorSubs[selector]
+	}
+
+	delete(from, id)
+}
+
+// GetALlFlags per specific source
+func (r *Multiplexer) GetALlFlags(source string) (string, error) {
```
bacherfl, line 138:
```suggestion
func (r *Multiplexer) GetAllFlags(source string) (string, error) {
```

flagd/cmd/start.go
Kavindu-Dodan: Good point, will change to kebab-case

Kavindu-Dodan: done this the commit https://github.com/open-feature/flagd/pull/1237/commits/d7c5798048864dcc504fffafbdebbf61fbdf1938

flagd/pkg/runtime/from_config.go
Kavindu-Dodan: agree, done with the commit https://github.com/open-feature/flagd/pull/1237/commits/d7c5798048864dcc504fffafbdebbf61fbdf1938

flagd/pkg/service/flag-sync/sync-multiplexer.go
Kavindu-Dodan: commit https://github.com/open-feature/flagd/pull/1237/commits/d7c5798048864dcc504fffafbdebbf61fbdf1938 fixes this

flagd/cmd/start.go
```diff
@@ -63,6 +67,9 @@ func init() {
 	flags.StringP(otelCollectorURI, "o", "", "Set the grpc URI of the OpenTelemetry collector "+
 		"for flagd runtime. If unset, the collector setup will be ignored and traces will not be exported.")
 
+	flags.BoolP(syncEnabledFlagName, "e", false, "Set true to enable gRPC sync service from flagd. "+
```
beeme1mr, line 70:
```suggestion
	flags.BoolP(syncEnabledFlagName, "e", false, "Enables the gRPC sync service. "+
```

beeme1mr: I could go either way on this one. I was trying to make it sound similar to the other description.

beeme1mr (approved):

flagd/cmd/start.go
Kavindu-Dodan: I like the suggestion, fixed with commit https://github.com/open-feature/flagd/pull/1237/commits/bf818c7a163d4b019e408fce9653bcc4a72aa1b4 as we need to update documentation as well

flagd/pkg/runtime/runtime.go
```diff
@@ -99,6 +101,19 @@ func (r *Runtime) Start() error {
 		}
 		return nil
 	})
+
+	g.Go(func() error {
+		// startup delay - allow all sync sources to finish the initial sync
+		<-time.After(5 * time.Second)
```
thisthat, line 107: why this timeout? I don't think we can guarantee that all sync fetch their flags in 5s 🤔

bacherfl: do we even need to wait for the initial flag retrieval to be finished here? Connecting to the flag sync endpoint will establish a stream over which any future updates will be communicated anyway, so if the initial sync is not finished at the time a client calls the `Sync` endpoint, the fetched configuration will be passed to any connected client anyway. However, please correct me if there is something I am missing / not considering here

docs/architecture.md
```diff
@@ -35,7 +35,9 @@ erDiagram
 
 In-process deployments embed the flagd evaluation engine directly into the client application through the use of an [in-process provider](./installation.md#in-process).
 The in-process provider is connected via the sync protocol to an implementing [gRPC service](./concepts/syncs.md#grpc-sync) that provides the flag definitions.
-This pattern requires an in-process implementation of the flagd evaluation engine, but has the benefit of no I/O overhead, since no inter-process communication is required.
+You can use flagd as a [gRPC sync service](./reference/grpc-sync-service.md).
+In this mode, the flag sync stream will expose aggregated flag configurations currently configured through [syncs](./concepts/syncs.md).
+This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead, since no inter-process communication is required.
```
toddbaert, line 40:
```suggestion
This pattern requires an in-process implementation of the flagd evaluation engine but has the benefit of no I/O overhead for flag evaluations, since no inter-process communication is required.
```

docs/reference/grpc-sync-service.md
```diff
@@ -0,0 +1,41 @@
+---
+description: flagd as a gRPC sync service
+---
+
+# Overview
+
+flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag configurations.
```
toddbaert, line 7:
```suggestion
flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag definitions.
```

docs/reference/grpc-sync-service.md
```diff
@@ -0,0 +1,41 @@
...
+    flagd ||--o{ "sync (file)" : watches
+    flagd ||--o{ "sync (http)" : polls
+    flagd ||--o{ "sync (grpc)" : "sync.proto (gRPC/stream)"
+    flagd ||--o{ "sync (kubernetes)" : watches
+    "In-Process provider" ||--|| flagd : "gRPC sync stream (default port 8015)"
+```
+
+You may change the default port of the service using startup flag `--sync-port` (or `-g` shothand flag).
+
+By default, the gRPC stream expose all the flag configurations merged following merge strategy of flagd.
```
toddbaert, line 25:
```suggestion
By default, the gRPC stream exposes all the flag configurations merged consistent with the flag's standard merge strategy.
```

docs/reference/grpc-sync-service.md
```diff
@@ -0,0 +1,41 @@
...
+You can read more about the merge strategy in our dedicated [concepts guide on syncs](../concepts/syncs.md).
+
+If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector valur to a source and stream specific flags.
```
toddbaert, line 28:
```suggestion
If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector value to a source, and stream just the flags identified in that source.
```

docs/reference/grpc-sync-service.md
```diff
@@ -0,0 +1,41 @@
...
+For example, if `selector` is set to `myFlags.json`, service will stream flags observed from `myFlags.json` file.
```
toddbaert, line 29: Does the selector have to match the CLI arg exactly? If so, we should probably make it `file://myFlags.json` for clarity.

flagd/cmd/start.go
```diff
@@ -18,16 +18,17 @@ import (
...
-	metricsExporter        = "metrics-exporter"
 	managementPortFlagName = "management-port"
+	metricsExporter        = "metrics-exporter"
 	otelCollectorURI       = "otel-collector-uri"
 	portFlagName           = "port"
 	serverCertPathFlagName = "server-cert-path"
 	serverKeyPathFlagName  = "server-key-path"
 	socketPathFlagName     = "socket-path"
 	sourcesFlagName        = "sources"
+	syncEnabledFlagName    = "sync-enabled"
```
toddbaert, line 29: Not sure, but we could _probably_ just remove this flag and always serve this. If you don't want to use it, you can simply not expose it on the network. Interested in other opinions on this one.

flagd/cmd/start.go
```diff
@@ -63,6 +67,9 @@ func init() {
 	flags.StringP(otelCollectorURI, "o", "", "Set the grpc URI of the OpenTelemetry collector "+
 		"for flagd runtime. If unset, the collector setup will be ignored and traces will not be exported.")
 
+	flags.BoolP(syncEnabledFlagName, "e", false, "Enables the gRPC sync service from flagd. "+
```
toddbaert, line 70: See [this](https://github.com/open-feature/flagd/pull/1237/files#r1517882344).

flagd/pkg/runtime/runtime.go
toddbaert: > if the initial sync is not finished at the time a client calls the Sync endpoint, the fetched configuration will be passed to any connected client anyway.

This is true, but there's a potential downside here, since the initial sync response is what providers use to signal their READY state. If they get this message, they will assume flags are ready to be evaluated, though in reality they are not since flagd hasn't reconciled it's internal state yet from all sources and only a partial (or possibly empty) set is returned.

I think the best solution would be to delay any sync responses until all sources have resolved, OR some timeout is reached (in which case an error is logged).

Kavindu-Dodan: This is how I looked at this, 

When there are multiple syncs (ex:- file sync + http sync + ...), each initial sync may complete with some delay. Consider the following mermaid diagram, 

```mermaid
gantt
    title Data Syncs & Service start
    dateFormat  s
    axisFormat %S

    section Data Syncs
    Sync 1          :a1, 1, 2s
    Sync 2          :a2, 1, 3s
    Sync 3          :a3, 2, 4s
    
    section Sync Service

    Service Start & running    :b2, 6, 10s
```

If we start the sync service without some delay, then sync subscriptions will receive multiple syncs till all data syncs stabilize with the initial data fetch. But if we add a start delay, at least we can attempt to stream stable data stream. 


 > I don't think we can guarantee that all sync fetch their flags in 5s  @thisthat 
 
 💯 , this is true and a 5s delay doesn't guarantee that all syncs are complete. We could optimize this with some tracking functionality but I did not invest time on this optimization. The same is true for re-syncs (especially after a delete). I think current implementation is good enough till we see the need to improve the logic
 
 > Connecting to the flag sync endpoint will establish a stream over which any future updates will be communicated anyway @bacherfl 
 
 Correct, this delay is only there for initial data sync stabilization. Like I explained above, not the optimal solution but I thought it's good to add this.
 
 And once service start, it's business as usual. Any data sync that doesn't trigger a resync will trigger a new sync stream update. 
 
 > Why not at resync ? 
 
 A resync means data store is not stable (a delete has occured). So there will be new data fetchings. Here too we can optmize to emit streams only when all resyncs are done. But I did not do this optimization either :)

toddbaert: > I think current implementation is good enough till we see the need to improve the logic

I also agree with this, but I think we should add an issue mentioning it. Something that basically adds a configurable timeout, but otherwise waits for every sync to initialize before responding the first time. Happy to consider this convo resolved if we define such an issue.

Kavindu-Dodan: >  since the initial sync response is what providers use to signal their READY state @toddbaert 

Yes, this is a really good point. And I will spend some time to see if we can optimize this 🤔 Or at least create a follow-up to improve this further

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (s *Service) Start() error {
+	err := s.server.Serve(s.listener)
+	if err != nil {
+		return fmt.Errorf("error from server start: %w", err)
+	}
+
+	return nil
+}
```
toddbaert, line 70: I dunno if it's something we should add here, but I think there's some missing context/cancellation handling somewhere.

If I send a `SIGINT` to flagd WITHOUT the `-e` flag, it shuts down fine within a couple seconds. If I send a `SIGINT` WITH the `-e` flag, flagd never shuts down.

docs/reference/grpc-sync-service.md
Kavindu-Dodan: Good question, I had to double check on this. 

No, this doesn't have to match CLI arg of URI format. It's only the source. For example if `--uri=file://myFlags.json` is used, then selector should be `myFlags.json`.

toddbaert (changes requested):
Looks really good, looks very clean for how much new functionality is here, but I guess part of that is because you did that nice refactor in advance, good job.

I tested this locally, and besides [this](https://github.com/open-feature/flagd/pull/1237#discussion_r1517923018) and [this](https://github.com/open-feature/flagd/pull/1237#issuecomment-1986008374), things work as expected. I tested both the `Sync` and the `FetchAllFlags` methods. I think I can approve once these are addressed.

flagd/cmd/start.go
Kavindu-Dodan: I added this service as "disabled by default" so that users won't be affected by a new service when we do the next release. This can happen if a K8s deployment uses the latest tag or by mistake if someone forgets to check release notes.

flagd/pkg/service/flag-sync/sync_service.go
Kavindu-Dodan: Good observation. I checked the shutdown logic and improved it to respect the SIGTERM. Updated the PR with commit https://github.com/open-feature/flagd/pull/1237/commits/711a21b916f8f4e65aa8c30fcc58b8df8dcc3293

Kavindu-Dodan:
> This seems a bit brittle and confusing, but I think it's out of scope of this PR. I think we should just document it for now, and perhaps add a new PR that allows users to define an friendly label for each source in the JSON config mode. @toddbaert 

Agree. The issue is how we handle `--uri` or `-f` , we convert the parameter to a `SourceConfig` [1] at startup. And then when updates are emitted from data syncs (ex:- file, k8s), the payload only contains the URI, which is merely the source without any prefix. 

I Will update the documentation highlighting this detail. 


[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/sync/builder/utils.go#L36

flagd/cmd/start.go
toddbaert: > I added this service as "disabled by default" so that users won't be affected by a new service when we do the next release. This can happen if a K8s deployment uses the latest tag or by mistake if someone forgets to check release notes.

That makes sense, however, since this is on a different port, I would be surprised if there was any severe networking or security implications since no forwarding/exposure would be occurring on this port from the wider infra. I think just mentioning this explicitly on the docs and title should be enough to assuage that concern.

I don't feel strongly about this point, but I think it's a bit odd having to opt into this considering we don't require users to opt into the evaluation.proto, and they are both equally valid to use.

My suggestion would be to remove this and change the title to `feat: serve sync.proto on port 8015` which will make things very clear.

flagd/pkg/service/flag-sync/sync_service.go
toddbaert: Tested and works great now.

toddbaert (approved):
SIGINT and docs look good now. Thanks a lot @Kavindu-Dodan

flagd/pkg/runtime/runtime.go
Kavindu-Dodan: After considering several options, I implemented a simple solution. Please see commit https://github.com/open-feature/flagd/pull/1237/commits/4e4e4df988345c8206ca837112dce14daf50f224

- I have transferred startup delay as well as emit condition to sync_service 

This means runtime no longer track resync condition nor it delay service startup. This responsiblity is at sync_service

- A simple count down latch `syncTracker` tracks flag sync updates 
- Server startup is delayed till either `syncTracker` emit a rediness OR a timeout
- `syncTracker` emit a ready when all known sync sources emit at least one sync payload 

I think this is good enough solution till we see more improvements :)

Kavindu-Dodan:
> SIGINT and docs look good now. Thanks a lot @Kavindu-Dodan

Thank you for detailed reviews 🤝

flagd/cmd/start.go
Kavindu-Dodan: I will wait for @beeme1mr , @thisthat and @bacherfl feedback and do this change if all of is agree :)

bacherfl: I agree, I think if this new endpoint is highlighted explicitly in the release notes, the flag should not be required

flagd/pkg/runtime/runtime.go
bacherfl: thanks for adding this @Kavindu-Dodan, I think that's a good solution :)

flagd/cmd/start.go
thisthat: I like the idea of hiding the new functionalities behind a "release feature flag" :p But I also agree with the points of @toddbaert and I don't think we really need this

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -0,0 +1,164 @@
...
+type syncTracker struct {
+	sources  []string
+	doneChan chan interface{}
+}
+
+func (t *syncTracker) done() <-chan interface{} {
+	return t.doneChan
+}
+
+func (t *syncTracker) trackAndUpdate(source string) {
```
thisthat, line 140: maybe `trackAndRemove`

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// syncTracker is a helper to track sync payloads at the startup
+// It simply starts with known set of sync sources and remove
+type syncTracker struct {
+	sources  []string
+	doneChan chan interface{}
+}
+
+func (t *syncTracker) done() <-chan interface{} {
```
thisthat, line 136: since we don't expose this function, why we need this getter? 🤔

flagd/pkg/runtime/runtime.go
Kavindu-Dodan: Done, removed the enabling flag with commit https://github.com/open-feature/flagd/pull/1237/commits/1c330ff9dcfcf9fdab4a3a6031b29c50ff3e53f3

flagd/cmd/start.go
Kavindu-Dodan: Done, I have removed the flag with commit https://github.com/open-feature/flagd/pull/1237/commits/1c330ff9dcfcf9fdab4a3a6031b29c50ff3e53f3

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (s *Service) Start(ctx context.Context) error {
+	// derive errgroup so we track ctx for exit as well as startup errors
+	g, lCtx := errgroup.WithContext(ctx)
+
+	g.Go(func() error {
+		// delay server start until we see all syncs from known sync sources OR timeout
+		select {
+		case <-time.After(5 * time.Second):
+			s.logger.Warn("timeout waiting for all sync sources to complete their initial sync. " +
```
thisthat, line 80: I am not a native speaker but this sounds more readable 🤔 
```suggestion
			s.logger.Warn("finish waiting for all sync sources to complete their initial sync. " +
```

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -0,0 +1,193 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if source != "" && !slices.Contains(r.sources, source) {
+		return fmt.Errorf("no flag watcher setup for source %s", source)
+	}
+
+	var initSync string
+
+	if source == "" {
```
thisthat, line 62: do we document this behavior? I was under the impression that if we don't provide any source then we don't return any flags

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -0,0 +1,193 @@
...
+		selectorSubs:  map[string]map[interface{}]subscription{},
+		selectorFlags: map[string]string{},
+	}
+
+	return m, m.reFill()
+}
+
+// Register a subscription
+func (r *Multiplexer) Register(id interface{}, source string, con chan payload) error {
+	r.mu.Lock()
```
thisthat, line 53: do we really need to lock here?

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -0,0 +1,193 @@
...
+	if !slices.Contains(r.sources, source) {
+		return "", fmt.Errorf("no flag watcher setup for source %s", source)
+	}
+
+	return r.selectorFlags[source], nil
+}
+
+// SourcesAsMetadata returns all known sources, comma separated to be used as service metadata
+func (r *Multiplexer) SourcesAsMetadata() string {
+	r.mu.Lock()
```
thisthat, line 152:
```suggestion
	r.mu.RLock()
```

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -0,0 +1,193 @@
...
+	} else {
+		from = r.selectorSubs[selector]
+	}
+
+	delete(from, id)
+}
+
+// GetAllFlags per specific source
+func (r *Multiplexer) GetAllFlags(source string) (string, error) {
+	if source == "" {
```
thisthat, line 139: I think we should read-lock here as well

flagd/pkg/service/flag-sync/sync-multiplexer.go
Kavindu-Dodan: This behaviour is agreed and defined through the issue itself :)

flagd/pkg/service/flag-sync/sync_service.go
Kavindu-Dodan: Given the struct is in the same module, yes it's possible to access the internal channel. However, as a best practice, I like to avoid directly using internals and rely on a getter instead.

flagd/pkg/service/flag-sync/sync_service.go
Kavindu-Dodan: done, updated :)

flagd/pkg/service/flag-sync/sync_service.go
Kavindu-Dodan: I tried to improve this log. Here the meaning should include a timeout as this is a timeout waiting for all sync sources to complete.

flagd/pkg/service/flag-sync/sync-multiplexer.go
Kavindu-Dodan: done :)

flagd/pkg/service/flag-sync/sync-multiplexer.go
Kavindu-Dodan: This is done :)

flagd/pkg/service/flag-sync/sync_service.go
thisthat: "done" feels like closing the `syncTracker`, maybe we could call it `getDone()` 🤔

thisthat (approved):

flagd/pkg/service/flag-sync/sync_service.go
toddbaert: It's a nit but I kinda agree.

Kavindu-Dodan:
Merging with all the approvals. Thank you @thisthat @toddbaert @bacherfl @beeme1mr for reviews 🤝

flagd/pkg/service/flag-sync/sync_service.go
Kavindu-Dodan: This is done :)

Issue #1242, opened 2024-03-08
[BUG] panic when yaml file watchers are updated


toddbaert:
### Observed behavior

flagd sometimes (~50% of the time) panics and crashes whena `.yaml` file it's watching is updated (note this DOES NOT seem to impact JSON payloads/files, and also doesn't impact payloads delivered over the wire).

```
2024-03-08T18:52:11.374Z        warn    evaluator/json.go:432   flag definition does not conform to the schema; validation errors:  1:(root): Invalid type. Expected: object, given: null         {"component": "evaluator", "evaluator": "json"}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x93641d]

goroutine 32 [running]:
github.com/open-feature/flagd/core/pkg/evaluator.validateDefaultVariants(0xc000638668?)
        /src/core/pkg/evaluator/json.go:452 +0x1d
github.com/open-feature/flagd/core/pkg/evaluator.(*JSON).configToFlags(0xc0002225e0, {0xc000639b78, 0x4}, 0xc00009a0b8)
        /src/core/pkg/evaluator/json.go:447 +0x251
github.com/open-feature/flagd/core/pkg/evaluator.(*JSON).SetState(0xc0002225e0, {{0xc000639b78, 0x4}, {0xc0001309c0, 0x16}, 0x0})
        /src/core/pkg/evaluator/json.go:110 +0x5b6
github.com/open-feature/flagd/core/pkg/runtime.(*Runtime).updateWithNotify(0xc00038b930, {{0xc000639b78, 0x4}, {0xc0001309c0, 0x16}, 0x0})
        /src/core/pkg/runtime/runtime.go:124 +0xed
github.com/open-feature/flagd/core/pkg/runtime.(*Runtime).Start.func1()
        /src/core/pkg/runtime/runtime.go:52 +0x12d
golang.org/x/sync/errgroup.(*Group).Go.func1()
        /go/pkg/mod/golang.org/x/sync@v0.5.0/errgroup/errgroup.go:75 +0x56
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
        /go/pkg/mod/golang.org/x/sync@v0.5.0/errgroup/errgroup.go:72 +0x96
```

### Expected Behavior

flagd never crashes when an watched-file is updated.

### Steps to reproduce

[...]

PR #1246, opened 2024-03-11
fix: occasional panic when watched YAML files change

dinhlockt02:
## This PR
At the moment, the yamlToJSON will convert `[]byte("")` to `"null"`. While "null" is a normal string, it will bypass all validator and lead to panic.
### Related Issues

Fixes [#1242](https://github.com/open-feature/flagd/issues/1242)

### Notes
 I found that sometime, the os.Readfile will return empty byte and no error, this will cause the yamlToJSON convert the rawData ("") to jsonData ("null"). 
Moreover, when I check the log, the watcher is reading file content a lot at a times for checking file changes, so it could be the reason why os.Readfile return empty byte?
image
### Follow-up Tasks
I think we have to check the behavior of os.Readfile?

### DIFF
--- a/core/pkg/sync/file/filepath_sync.go
+++ b/core/pkg/sync/file/filepath_sync.go
@@ -171,6 +171,10 @@ func (fs *Sync) fetch(_ context.Context) (string, error) {
 // yamlToJSON is a generic helper function to convert
 // yaml to json
 func yamlToJSON(rawFile []byte) (string, error) {
+	if len(rawFile) == 0 {
+		return "", nil
+	}
+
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):
Thank you for identifying the root cause. I think this is a good enough fix for the bug :)

toddbaert (approved):
Thanks!

toddbaert:
@dinhlockt02 Thanks so much for your contribution. The only change I will make is the PR title, so it might be a bit easier to understand in our release notes.

PR #1247, opened 2024-03-12
feat:  OFREP support for flagd 

### LINKED ISSUE #1245: [FEATURE] implement the initial version of OFREP in flagd
### Requirements

**Background**

With PR https://github.com/open-feature/protocol/pull/2 already merged, OpenFeature now has the first version of the OFREP ready. Given that flagd is a popular subproject in OpenFeature, flagd should implement OFREP as a reference implementation. 

**Consider**

When implementing the first version, consider the following tasks, 

- [x] Expose OFREP endpoint from a dedicated port
- [x] Expose OFREP single flag implementation
- [x] Expose OFREP bulk evaluation
- [x] For the first version, 
  - [x] Avoid introducing a caching layer for bulk evaluation and keep the implementation simple 
  - [x] Avoid any filtering for bulk evaluation requests to keep the implementation simple  
  - [x] Use minimalistic libraries for request validation against models [1] & request routing [2]
  - [x] Basic documentation with notification the implementation is experimental

**AC**

flagd expose OFREP and allow both individual and bulk flag evaluations 

[1] - https://github.com/go-playground/validator
[2] - https://github.com/gorilla/mux

Kavindu-Dodan:
## This PR

Fixes #1245 and introduce OFREP support for flagd

OFREP service runs on 8016 by default and can simply use curl for flag evaluations, 

```shell
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/myBoolFlag'
```

NOTE - This PR touch several files as I had to migrate from `github.com/golang/mock/gomock` to `go.uber.org/mock/gomock`. This is because https://github.com/golang/mock is no longer maintained.


core/pkg/evaluator/json.go
```diff
@@ -270,6 +270,20 @@ func (je *JSON) ResolveObjectValue(
...
+func (je *JSON) ResolveAsAnyValue(
+	ctx context.Context,
+	reqID string,
+	flagKey string,
+	context map[string]any,
+) AnyValue {
+	_, span := je.jsonEvalTracer.Start(ctx, "resolveAnyValue")
+	defer span.End()
+
+	je.Logger.DebugWithID(reqID, fmt.Sprintf("evaluating flag ` %s` as a generic flag", flagKey))
```
thisthat, line 282:
```suggestion
	je.Logger.DebugWithID(reqID, fmt.Sprintf("evaluating flag `%s` as a generic flag", flagKey))
```

core/pkg/service/ofrep/models.go
```diff
@@ -0,0 +1,106 @@
...
+	status := 400
+
+	switch result.Error.Error() {
+	case model.FlagNotFoundErrorCode:
+		status = 404
+		payload.ErrorCode = model.FlagNotFoundErrorCode
+		payload.ErrorDetails = fmt.Sprintf("flag `%s` does not exisit", result.FlagKey)
+	case model.ParseErrorCode:
+		payload.ErrorCode = model.ParseErrorCode
+		payload.ErrorDetails = "error parsing the flag"
```
thisthat, line 97:
```suggestion
		payload.ErrorDetails = fmt.Sprintf("error parsing the flag `%s`", result.FlagKey)
```

core/pkg/service/ofrep/models.go
```diff
@@ -0,0 +1,106 @@
...
+		Key: result.FlagKey,
+	}
+
+	status := 400
+
+	switch result.Error.Error() {
+	case model.FlagNotFoundErrorCode:
+		status = 404
+		payload.ErrorCode = model.FlagNotFoundErrorCode
+		payload.ErrorDetails = fmt.Sprintf("flag `%s` does not exisit", result.FlagKey)
```
thisthat, line 94:
```suggestion
		payload.ErrorDetails = fmt.Sprintf("flag `%s` does not exist", result.FlagKey)
```

core/pkg/service/ofrep/models.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func ContextErrorResponseForBulkAndStatus() BulkEvaluationError {
+	return BulkEvaluationError{
+		ErrorCode:    model.InvalidContextCode,
+		ErrorDetails: "provided context is invalid",
+	}
+}
+
+func ValueToStatusAndError(result evaluator.AnyValue) (int, EvaluationError) {
```
thisthat, line 83: I would maybe change the name to `ErrorResponseFrom` (similar to the prev function) or something more descriptive that we should call this function only when an error occurs

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (h *handler) HandleFlagEvaluation(w http.ResponseWriter, r *http.Request) {
+	requestID := xid.New().String()
+
+	// obtain flag key
+	vars := mux.Vars(r)
+	flagKey := vars[key]
+	if vars == nil || strings.TrimSpace(flagKey) == "" {
+		h.writeJSONToResponse(
+			http.StatusInternalServerError,
```
thisthat, line 47: I think this error is a client error and should be HTTP 400 - Bad Request

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -0,0 +1,115 @@
...
+		h.writeJSONToResponse(http.StatusBadRequest, ofrep.ContextErrorResponseForBulkAndStatus(), w)
+		return
+	}
+
+	context := map[string]any{}
+	if res, ok := request.Context.(map[string]any); ok {
+		context = res
+	} else {
+		h.Logger.WarnWithID(requestID, "provided context does not comply with flagd, continuing ignoring the context")
+	}
```
thisthat, line 92: this is repeated, maybe we could export it into its own private function 🤔

core/pkg/service/ofrep/models.go
Kavindu-Dodan: Done, renamed to `EvaluationErrorResponseFrom`

flagd/pkg/service/flag-evaluation/ofrep/handler.go
Kavindu-Dodan: I thought about using 4xx error code, however, this can only happen if our mux configuration or path configurations are incorrect. Otherwise, requests to `ofrep/v1/evaluate/flags/` (missing flag in path) is considered as a bulk evaluation request. This is why I use 500 status code.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
Kavindu-Dodan: Done, common logic was extracted to helper functions :)

flagd/pkg/service/flag-evaluation/ofrep/handler.go
thisthat: good point, I was thinking of having a key value equal to `%20`, which is trimmed away 🤔 
Maybe we need to split:
- `vars = nil` -> 500
- empty key -> 400

Kavindu-Dodan: @thisthat this is good point. I update to allow `empty key` as there's no rule or parser that prevents flagd from having a flag key with empty string (ex: `  ` is valid ). And updated vars= nil to be a 500

thisthat (approved):

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -0,0 +1,119 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Add("Content-Type", "application/json")
+	w.WriteHeader(status)
+	_, err = w.Write(marshal)
+	if err != nil {
+		h.Logger.Warn(fmt.Sprintf("error while writing response: %v", err))
+		return
```
bacherfl, line 96:
```suggestion
```

flagd/pkg/service/flag-evaluation/ofrep/handler_test.go
```diff
@@ -0,0 +1,263 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			recorder := httptest.NewRecorder()
+
+			h.writeJSONToResponse(test.status, test.payload, recorder)
+
+			if test.expectedStatus != recorder.Code {
+				t.Errorf("expected status code %d, but got %d", test.expectedStatus, recorder.Code)
+			}
+
```
bacherfl, line 257: here we could also add a check if the payload was set by accessing it via

```
payload, err := io.ReadAll(recorder.Body)
```

bacherfl (approved):
great work! approved with some minor suggestions :)

flagd/pkg/service/flag-evaluation/ofrep/handler_test.go
Kavindu-Dodan: Good suggestion, updated with the commit - https://github.com/open-feature/flagd/pull/1247/commits/83857f55635b1045100706e602e09b8eb4241dd1

PR #1253, opened 2024-03-21
docs: playground share button

DBlanchard88:
Added share button to flagd playground


toddbaert:
Hey @DBlanchard88 . Thanks so much!

This basically works! The one thing you missed (and I don't blame you) was to run the command `make playground-publish` - this takes the little react app and moves it into actual doc-site so it's available at the playground link. I ran this command and committed the newly modified output, so now you can see it in the preview Netflify made above (this is the only change I made). You have to do this little step to make sure your changes in the playground exist on the real docsite.

There's a couple other small things I request:

Can you add a tiny bit of padding/margin above the buttons (evaluate/reset/share)? They are stuck right under the context (it was like this before your changes, but it looks kinda weird).
 
image

Also, in addition to the share button updating the URL, can you make it [copy the URL to the clipboard](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard)? And maybe change it to "Copy URL" since it doesn't actually prompt the share dialog (which isn't available on many desktop browsers anyway).

playground-app/src/App.tsx
```diff
@@ -198,6 +233,26 @@ function App() {
...
+  
+    if (Object.keys(scenarios).includes(selectedTemplate) &&
+      scenarios[selectedTemplate].flagDefinition === featureDefinition) {
+      newUrl.searchParams.set('scenario_name', selectedTemplate);
+    } else {
+      newUrl.searchParams.delete('scenario_name');
+      newUrl.searchParams.set('flags', encodedConfig);
+      newUrl.searchParams.set('flagKey', flagKey);
+      newUrl.searchParams.set('returnType', returnType);
+      newUrl.searchParams.set('evalContext', encodedEvalContext);
```
toddbaert, line 250: use _kebab-case_ for all query params

toddbaert:
@DBlanchard88 note I've changed your title to "docs: ... " since this IS a feature only for the docs, not really for flagd.

DBlanchard88:
thanks for the feedback! ill make those changes

playground-app/src/App.tsx
```diff
@@ -198,6 +233,32 @@ function App() {
...
+
+    if (Object.keys(scenarios).includes(selectedTemplate) &&
+      scenarios[selectedTemplate].flagDefinition === featureDefinition) {
+      newUrl.searchParams.set('scenario-name', selectedTemplate);
+    } else {
+      newUrl.searchParams.delete('scenario-name');
+      newUrl.searchParams.set('flags', encodedConfig);
+      newUrl.searchParams.set('flag-key', flagKey);
+      newUrl.searchParams.set('return-type', returnType);
+      newUrl.searchParams.set('eval-context', encodedEvalContext);
```
toddbaert, line 250: nit: use constants for these strings since we use them in 2 places.

toddbaert (approved):

Kavindu-Dodan (approved):

beeme1mr:
Thanks for working on this @DBlanchard88. I tried it out but ran into a few issues.

## Issues

### 414 URI Too Long

1. Select the `Progressive rollout` scenario and copy the URL
2. Paste the URL in a new tab
3. The browser will show a 414 error

### Evaluation context not copied

1. Modify evaluation context and copy the URL
2. Paste the URL in a new tab
4. The modification the evaluation context is lost

## Feedback

I recommend changing the copy on the button from "Copy URL" to "Share" and displaying a toast message that the URL was copied.

toddbaert:
Netlify deploy is failing due to this: https://github.com/open-feature/flagd/issues/1260

beeme1mr (approved):
Hey @DBlanchard88, this looks good! Would you be able to have to "URL copied" toast message fade out after a few seconds?

DBlanchard88:
@beeme1mr absolutely

toddbaert:
> Hey @DBlanchard88, this looks good! Would you be able to have to "URL copied" toast message fade out after a few seconds?

Maybe it's just me, but the message seems a bit oddly positioned too. It doesn't have the highest z-index and it's not really close to the button. You might want to just position it close to the bottom middle of the screen, absolutely, with a high z index

DBlanchard88:
@toddbaert how do you feel about these two possibilities?  
brave_r281Z2ZPJF
brave_B8eUIAmdl4

Kavindu-Dodan (approved):
Cheked the preview [1] and it looks really good :) 

[1] - https://deploy-preview-1253--polite-licorice-3db33c.netlify.app/playground/

PR #1259, opened 2024-03-28
chore: refactor evaluation core

Kavindu-Dodan:
## This PR

Refactor the flagd evaluator such that storage and evaluation logic are isolated. This allows us to plug any storage layer to the same evaluation logic, which could be valuable in the long run

image

### DIFF
--- a/core/pkg/evaluator/json.go
+++ b/core/pkg/evaluator/json.go
@@ -28,43 +28,32 @@ import (
 
 const (
 	SelectorMetadataKey = "scope"
-
-	flagdPropertiesKey = "$flagd"
-
+	flagdPropertiesKey  = "$flagd"
 	// targetingKeyKey is used to extract the targetingKey to bucket on in fractional
 	// evaluation if the user did not supply the optional bucketing property.
 	targetingKeyKey = "targetingKey"
+	Disabled        = "DISABLED"
 )
 
 var regBrace *regexp.Regexp
 
-type flagdProperties struct {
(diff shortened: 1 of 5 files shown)


core/pkg/evaluator/json.go
```diff
@@ -301,22 +314,21 @@ func resolve[T constraints](reqID string, key string, context map[string]any, va
...
-// runs the rules (if defined) to determine the variant, otherwise falling through to the default
 // nolint: funlen
-func (je *JSON) evaluateVariant(reqID string, flagKey string, context map[string]any) (
+func (je *Resolver) evaluateVariant(reqID string, flagKey string, c map[string]any) (
 	variant string, variants map[string]interface{}, reason string, metadata map[string]interface{}, err error,
 ) {
 	metadata = map[string]interface{}{}
 
-	flag, ok := je.store.Get(flagKey)
+	flag, ok := je.store.Get(context.Background(), flagKey)
```
thisthat, line 323: maybe we can create a `ctx := context.Background()` and reuse it in other calls

Kavindu-Dodan: @thisthat good suggestion, updated with commit https://github.com/open-feature/flagd/pull/1259/commits/ec1c907e3016b51886ea72938007f51c2b749685

toddbaert (approved):
Makes sense to me.

thisthat (approved):

PR #1294, opened 2024-04-19
fix: fraction operation - add validations for nil values (#1285)

### LINKED ISSUE #1285: [BUG] Fraction operation - add validations for nil values
### Observed behavior

fractional custom json logic validation fails when a context-bound variable is set to nil.

Consider following rule,

```json
{
  "fractional": [
    {
      "var": "email"
    },
    [
      "red",
      50
    ],
    [
      "blue",
      50
    ]
  ]
}
```

if context-bound variable `email` is nil. then parsing will faile at [1]

[1] - https://github.com/open-feature/flagd/blob/core/v0.9.0/core/pkg/evaluator/fractional.go#L79-L82

### Expected Behavior

We should perform a best-effort parsing and remove `nil` variables. 

### Steps to reproduce

_No response_

Baalekshan:
## This PR

adds validation for nil values by using []interfaces{} instead of []any

### Related Issues

Fixes #1285

### DIFF
--- a/core/pkg/evaluator/fractional.go
+++ b/core/pkg/evaluator/fractional.go
@@ -72,13 +72,13 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 	return bucketBy, feDistributions, nil
 }
 
-func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluationDistribution, error) {
+func parseFractionalEvaluationDistributions(values []interface{}) ([]fractionalEvaluationDistribution, error) {
 	sumOfPercentages := 0
 	var feDistributions []fractionalEvaluationDistribution
 	for i := 0; i < len(values); i++ {
-		distributionArray, ok := values[i].([]any)
+		distributionArray, ok := values[i].([]interface{})
 		if !ok {
-			return nil, errors.New("distribution elements aren't of type []any")
+			return nil, errors.New("distribution elements aren't of type []interface{}")
 		}
 
 		if len(distributionArray) != 2 {


core/pkg/evaluator/fractional.go
```diff
@@ -72,13 +72,13 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 	return bucketBy, feDistributions, nil
 }
 
-func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluationDistribution, error) {
+func parseFractionalEvaluationDistributions(values []interface{}) ([]fractionalEvaluationDistribution, error) {
```
Kavindu-Dodan, line 75: Unfortunately, this does not solve the underlying issue . `any` & `interface{}` are type aliasses.

You have to validate the content of the array and ignore nil values to continue.

Kavindu-Dodan (changes requested):

core/pkg/evaluator/fractional.go
Baalekshan: if values[i] == nil {
			continue
		}


Will this be good if I included it inside the loop.

Kavindu-Dodan: That's an option, but I would avoid adding the check into the `parseFractionalEvaluationDistributions`. The parameters here should already be filtered and validated.

The fix should focus on parsing of the distributions [1]. If the first parameter is not a string, then we should check it's non-nil 

[1] - https://github.com/open-feature/flagd/blob/core/v0.9.0/core/pkg/evaluator/fractional.go#L58-L64

Kavindu-Dodan:
@Baalekshan thank you for your interest in this issue.

I had to fix this as we were preparing the next release of flagd [1]. You can check the fix I did here [2]. I hope you find it helpful and continue to contribute in future :) 

[1] - https://github.com/open-feature/flagd/pull/1292
[2] - https://github.com/open-feature/flagd/pull/1295

Baalekshan:
@Kavindu-Dodan Yeah I look for contributing in future.

Issue #1312, opened 2024-05-17
Feature-versioning via flagd schema


toddbaert:
Now that we have multiple components implementing flagd's JSON evaluator, we need a way to track and communicate the compliance of various implementations. Doing this by versioning the artifacts themselves isn't practical, since implementations will have their own fixes and breaking changes. **I propose we rigorously version the flagd-schema, and then each implementation can note what version they support, and log warnings or throw (we should decide on these specifics).** We already version the schema release itself, so some of this work is already done.

Work required for this:

- [ ] add a full version path to https://flagd.dev/schema/v0/flags.json (ie: https://flagd.dev/schema/v1.2.3/flags.json)
  - enhance the Makefile https://github.com/open-feature/flagd/blob/34756fe02d16e62e12b9baf65e983cd4157fa89e/Makefile#L128 to pull the latest version and create the corresponding folder to serve it from
- [ ] add a version to the flagd-schema itself, in its `$id` property, currently it's `"$id": "https://flagd.dev/schema/v0/flags.json"`, this should be made something like `"$id": "https://flagd.dev/schema/v1.2.3/flags.json""`
  - this should be automated with release please
- [ ] update flagd specification to mention this versioning
- update all the providers to leverage this and denote their support
  - [ ] Java
  - [ ] .NET
  - [ ] JS
  - [ ] Python

PR #1313, opened 2024-05-21
feat: support relative weighting for fractional evaluation

### LINKED ISSUE #1282: [FEATURE] Change `fractional` custom op from percentage-based to relative weighting.
### Requirements

Currently, the fractional operator is percentage-based; meaning you must specify the relative distribution of the variants in terms of a percentage. I believe we should change this instead to be a relative weight, meaning that the integers do not need to sum to 100. For example, this fractional:

```json
"fractional": [
          [
            "red",
            25
          ],
          [
            "blue",
            25
          ],
          [
            "green",
            25
          ],
          [
            "grey",
            25
          ]
        ]
```

Would remain valid, but could also be expressed as:

```json
"fractional": [
          [
            "red",
            1
          ],
          [
            "blue",
            1
          ],
          [
            "green",
            1
          ],
          [
            "grey",
            1
          ]
        ]
```

We could even make the number in the bucket optional, so that an even split would be assumed if none was provided (we could simply default to 1).

Benefits:

- adding new buckets becomes easy, there's no need to alter previous buckets
- supports much finer targeting
  - clear and consistent support across implementations for targeting finer than a single percentage (I don't believe all impls currently support non-integer distributions)
  - easy a clear support for fractions that aren't factors of 100 (`1/1/1` vs `33/33/34`)
- **non breaking** (current percentage based rules will still work)

Challenges:

- we should specify max values for the integers, and perhaps a max array length for the bucket array

**_If there's a general consensus to this, I will create some child issues in the relevant repositories and use this issue as the parent "tracking" issue._**

bacherfl:
Closes #1282 

This PR adds support for using relative weights instead of percentages that need to add up to 100.
The behavior for existing flag configs does not change with this PR, so those will continue to work as they did previously

### DIFF
--- a/core/pkg/evaluator/fractional.go
+++ b/core/pkg/evaluator/fractional.go
@@ -16,8 +16,21 @@ type Fractional struct {
 }
 
 type fractionalEvaluationDistribution struct {
-	variant    string
-	percentage int
+	totalWeight      int
+	weightedVariants []fractionalEvaluationVariant
+}
+
+type fractionalEvaluationVariant struct {
+	variant string
+	weight  int
+}
+
+func (v fractionalEvaluationVariant) getPercentage(totalWeight int) float64 {
+	if totalWeight == 0 {
+		return 0
+	}
+
+	return 100 * float64(v.weight) / float64(totalWeight)
 }
 
 func NewFractional(logger *logger.Logger) *Fractional {
@@ -34,7 +47,7 @@ func (fe *Fractional) Evaluate(values, data any) any {
 	return distributeValue(valueToDistribute, feDistributions)
 }
 
-func parseFractionalEvaluationData(values, data any) (string, []fractionalEvaluationDistribution, error) {
+func parseFractionalEvaluationData(values, data any) (string, *fractionalEvaluationDistribution, error) {
 	valuesArray, ok := values.([]any)
 	if !ok {
 		return "", nil, errors.New("fractional evaluation data is not an array")
@@ -77,9 +90,11 @@ func parseFractionalEvaluationData(values, data any) (string, []fractionalEvalua
 	return bucketBy, feDistributions, nil
 }
 
-func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluationDistribution, error) {
-	sumOfPercentages := 0
-	var feDistributions []fractionalEvaluationDistribution
+func parseFractionalEvaluationDistributions(values []any) (*fractionalEvaluationDistribution, error) {
+	feDistributions := &fractionalEvaluationDistribution{
+		totalWeight:      0,
+		weightedVariants: make([]fractionalEvaluationVariant, len(values)),
+	}
 	for i := 0; i < len(values); i++ {
 		distributionArray, ok := values[i].([]any)
 		if !ok {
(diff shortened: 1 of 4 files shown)


core/pkg/evaluator/fractional.go
```diff
@@ -96,37 +111,32 @@ func parseFractionalEvaluationDistributions(values []any) ([]fractionalEvaluatio
 			return nil, errors.New("first element of distribution element isn't string")
 		}
 
-		percentage, ok := distributionArray[1].(float64)
+		weight, ok := distributionArray[1].(float64)
```
thisthat, line 114: @toddbaert you proposed to make the second element optional w/ `1` as default, is it still something that we want to do? 🤔

thisthat (approved):
LGTM 👍  I left a question for @toddbaert to clarify an open requirement :)

core/pkg/evaluator/fractional.go
bacherfl: I will adapt the PR to also support that use case - since this does not break anything it makes sense to also have this :)

toddbaert:
Nice @bacherfl this is great! I will review thoroughly next week. I think we can merge this, but I don't want to add schema support for it until we have something like https://github.com/open-feature/flagd/issues/1312 so that we can track the implementation of this across different providers and avoid confusion as to why some flagd and some providers support it and others do not.

I don't think it will be that much work to support https://github.com/open-feature/flagd/issues/1312. I will follow up next week.

Kavindu-Dodan (approved):

docs/reference/specifications/custom-operations/fractional-operation-spec.md
```diff
@@ -1,10 +1,10 @@
 # Fractional Operation Specification
 
 This evaluator allows to split the returned variants of a feature flag into different buckets,
-where each bucket can be assigned a percentage, representing how many requests will resolve to the corresponding
+where each bucket can be assigned a weight, representing how many requests will resolve to the corresponding
 variant.
 
-The sum of all weights must be `100`, and the distribution must be performed by using the value of a referenced
+The distribution must be performed by using the value of a referenced
```
toddbaert, line 7:
```suggestion
The distribution must be performed by using the value of a referenced property 
```

docs/reference/specifications/custom-operations/fractional-operation-spec.md
```diff
@@ -1,10 +1,10 @@
 # Fractional Operation Specification
 
 This evaluator allows to split the returned variants of a feature flag into different buckets,
-where each bucket can be assigned a percentage, representing how many requests will resolve to the corresponding
+where each bucket can be assigned a weight, representing how many requests will resolve to the corresponding
 variant.
 
-The sum of all weights must be `100`, and the distribution must be performed by using the value of a referenced
+The distribution must be performed by using the value of a referenced property
```
toddbaert, line 7:
```suggestion
The distribution must be performed by using the value of a referenced property
```

docs/reference/custom-operations/fractional-operation.md
```diff
@@ -56,8 +80,8 @@ The value is an array and the first element is a nested JsonLogic rule which res
 This rule should typically consist of a seed concatenated with a session variable to use from the evaluation context.
 This value should typically be something that remains consistent for the duration of a users session (e.g. email or session ID).
 The seed is typically the flagKey so that experiments running across different flags are statistically independent, however, you can also specify another seed to either align or further decouple your allocations across different feature flags or use-cases.
-The other elements in the array are nested arrays with the first element representing a variant and the second being the percentage that this option is selected.
-There is no limit to the number of elements but the configured percentages must add up to 100.
+The other elements in the array are nested arrays with the first element representing a variant and the second being the relative weight for this option.
+There is no limit to the number of elements.
```
toddbaert, line 84:
```suggestion
There is no limit to the number of elements.

> [!NOTE]
> Older versions of the `fractional` operation were percentage based, and required all variants weights to sum to 100.
```

toddbaert (changes requested):
This looks great to me, but I'm going to block it for just a while to see about the feasibility of [this](https://github.com/open-feature/flagd/issues/1312), which would help us to release this in a consistent way across multiple implementations.

Consider this approved but temporarily blocked for merge.

toddbaert:
@bacherfl @Kavindu-Dodan @thisthat see: 

https://github.com/open-feature/flagd-schemas/pull/163
https://github.com/open-feature/flagd/pull/1321

With these, we will serve fully versioned schemas for flagd, which we can use to denote features and fixes such as the one in this PR.

core/pkg/evaluator/fractional.go
toddbaert: Yep!

toddbaert:
> @bacherfl @Kavindu-Dodan @thisthat see:
> 
> [open-feature/flagd-schemas#163](https://github.com/open-feature/flagd-schemas/pull/163) #1321
> 
> With these, we will serve fully versioned schemas for flagd, which we can use to denote features and fixes such as the one in this PR.

I'm not doing this afterall. I think it's more maintenance than it's worth, and @aepfli took the time to implement this in most other impls, so I think we can just move forward with this now. I will do the release(s) over the next couple days.

toddbaert (approved):
Thanks so much @bacherfl

PR #1321, opened 2024-05-27
chore: Makefile full schema versioning

toddbaert:
Adds make target that pulls the specified version of the flagd-json-schema and serves it at the appropriate location

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,7 +127,7 @@ generate-proto-docs: pull-schemas-submodule
 	&& rm -f ${PWD}/$(DOCS_DIR)/reference/specifications/protos-with-toc.md
 
 # update or add the publicly served schema at https://flagd.dev/schema/${JSON_SCHEMA_VERSION}/flags.json
-# update JSON_SCHEMA_VERSION var at the top of this file with a the version to be updated or addeds
+# update JSON_SCHEMA_VERSION var at the top of this file with the version to be updated or added and served publicly
 .PHONY: update-public-schema
 update-public-schema: pull-schemas-submodule
 	rm -rf ./$(PUBLIC_JSON_SCHEMA_DIR) || true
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -125,11 +126,12 @@ generate-proto-docs: pull-schemas-submodule
 	&& sed '/^## Table of Contents/,/#top/d' ${PWD}/$(DOCS_DIR)/reference/specifications/protos-with-toc.md >> ${PWD}/$(DOCS_DIR)/reference/specifications/protos.md \
 	&& rm -f ${PWD}/$(DOCS_DIR)/reference/specifications/protos-with-toc.md
 
-# Update the schema at flagd.dev
-# PUBLIC_JSON_SCHEMA_DIR above controls the dir (and therefore major version)
+# update or add the publicly served schema at https://flagd.dev/schema/${JSON_SCHEMA_VERSION}/flags.json
+# update JSON_SCHEMA_VERSION var at the top of this file with a the version to be updated or addeds
```
toddbaert, line 130:
```suggestion
# update JSON_SCHEMA_VERSION var at the top of this file with the version to be updated or added and served publicly
```

Makefile
```diff
@@ -1,6 +1,7 @@
 PHONY: .docker-build .build .run .mockgen
 PREFIX=/usr/local
-PUBLIC_JSON_SCHEMA_DIR=docs/schema/v0/
+JSON_SCHEMA_VERSION=v0.2.3
```
Kavindu-Dodan, line 3: can this be something for renovate based auto updates? [1] 🤔 

[1] - https://github.com/open-feature/flagd/blob/main/renovate.json

toddbaert: I was thinking about that but I decided not to. My reasoning is that in many cases we will only want to serve a new schema with some thought and consideration, and probably at least an implementation in flagd (like for example @bacherfl change for weighing)

toddbaert: Yes. I could also take it from an env var and just default it here. You only need to run it once and it creates the dirs (which correspond to the served URLs)

Kavindu-Dodan: This is fine, we just need to remember the change :)

PR #1326, opened 2024-06-15
feat: support `FLAGD_DEBUG` / `--debug` / `-x`

### LINKED ISSUE #1323: [FEATURE] Allow flagd debug flag to be set through environment variable
**Background**

flagd `debug` flag is defined at the root level and available as a persistent flag. However persistent flags are not modifiable through environment variables.

**Requirement**

Given flagd support environment variable-based configurations, we should add support to define debug flag through environment variable (ex:- `FLAG_DEBUG` or `DEBUG`)

Also consider the following,

- Given we haven't visited cobra and viper configurations for a long time, it could be a good opportunity to refactor flagd startup setup
- Document `FLAGD` prefix based environment variable resolving capability

snosratiershad:
## This PR
- Observes debug env variable [with or prefix will be `FLAGD_DEBUG`] and binds `--debug` and `-x` flags to it.

### Related Issues
Fixes #1323

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -9,7 +9,7 @@ description: troubleshooting flagd
 If a flag or targeting rule isn't proceeding the way you'd expect, you may want to enable more verbose logging.
 
 flagd and flagd providers typically have debug or verbose logging modes that you can use for this sort of troubleshooting.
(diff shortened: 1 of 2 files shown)


Kavindu-Dodan (approved):

toddbaert:
Thanks @snosratiershad

PR #1338, opened 2024-06-20
feat!: support emitting errors from the bulk evaluator

### LINKED ISSUE #1328: [Improvement] Improve bulk evaluator with ability to return errors
### Requirements

**Background**

flagd's `IResolver` defines `ResolveAllValues` contract and currently it only supports returning an array of values [1]. This contract is currently used by flagd's ofrep implementation. 

**Improvement**

Given flagd's core is reusable, it should support returning bulk evaluation errors. For example, this could happen as a result of an invalid Store value or Store retrieval error [2]. 

**AC**

Consider adding error return based on potential errors that could happen with bulk evaluation (for example Store errors). Further use this in OFREP evaluation and utilize error response for bulk evaluation [3]

[1] - https://github.com/open-feature/flagd/blob/flagd/v0.10.3/core/pkg/evaluator/ievaluator.go#L77
[2] - https://github.com/open-feature/flagd/blob/flagd/v0.10.3/core/pkg/store/flags.go#L15
[3] - https://github.com/open-feature/protocol/blob/main/service/openapi.yaml#L137-L142

Kavindu-Dodan:
Fixes #1328

### Improvement

flagd's core components are intended to be reused. This PR change the`IStore` interface by allowing an error to be returned from `GetAll`.  This error is then propagated through `ResolveAllValues`. This change enables custom `IStore` implementations to return errors and propagate them through the resolver layer.

With this change, I have upgrade OFREP bulk evaluator and flagd RPC  `ResolveAll` with error propagation.

OFREP - Log warning with resolver error and return HTTP 500 with a tracking reference
RPC - Log warning with resolver error and return an error with a tracking reference

### DIFF
--- a/core/pkg/evaluator/fractional.go
+++ b/core/pkg/evaluator/fractional.go
@@ -27,7 +27,7 @@ func NewFractional(logger *logger.Logger) *Fractional {
 func (fe *Fractional) Evaluate(values, data any) any {
 	valueToDistribute, feDistributions, err := parseFractionalEvaluationData(values, data)
 	if err != nil {
-		fe.Logger.Error(fmt.Sprintf("parse fractional evaluation data: %v", err))
+		fe.Logger.Warn(fmt.Sprintf("parse fractional evaluation data: %v", err))
 		return nil
 	}
 
--- a/core/pkg/store/flags.go
+++ b/core/pkg/store/flags.go
@@ -12,7 +12,7 @@ import (
 )
 
 type IStore interface {
-	GetAll(ctx context.Context) map[string]model.Flag
+	GetAll(ctx context.Context) (map[string]model.Flag, error)
 	Get(ctx context.Context, key string) (model.Flag, bool)
 	SelectorForFlag(ctx context.Context, flag model.Flag) string
 }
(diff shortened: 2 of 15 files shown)


core/pkg/evaluator/fractional.go
```diff
@@ -27,7 +27,7 @@ func NewFractional(logger *logger.Logger) *Fractional {
 func (fe *Fractional) Evaluate(values, data any) any {
 	valueToDistribute, feDistributions, err := parseFractionalEvaluationData(values, data)
 	if err != nil {
-		fe.Logger.Error(fmt.Sprintf("parse fractional evaluation data: %v", err))
+		fe.Logger.Warn(fmt.Sprintf("parse fractional evaluation data: %v", err))
```
Kavindu-Dodan, line 30: Note - noted this while local validation. Error logs add stack traces, which is an overkill. Hence reduced to warn level.

aepfli (approved):

core/pkg/store/flags.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 type IStore interface {
-	GetAll(ctx context.Context) map[string]model.Flag
+	GetAll(ctx context.Context) (map[string]model.Flag, error)
```
thisthat, line 15: why this change? 🤔 it looks like only `nil` is returned

Kavindu-Dodan: @thisthat good question. The implementation of flagd currently does not use error return. But this change allows us to add an error if we see the need. But the main usage is for custom IStore implementations which use flagd's JSON evaluator [1]. These users now can use the evaluator and and return store-bound errors (ex:- Store access error if store implements a database access or another API)

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/evaluator/json.go#L73

toddbaert (approved):
Makes perfect sense to me!

PR #1366, opened 2024-07-22
feat: Support blob type sources and GCS as an example of such source.

alan-kut:
## This PR
- adds support for blob type sources
- adds support for GCS using blob source

### Notes
In a similar fashion other cloud storages supported by go-cloud blob can be added.

### How to test
- create a Google Cloud Storage bucket and a file with flags definition inside
- get application default credentials: https://cloud.google.com/docs/authentication/provide-credentials-adc
- start the flagd with source pointing to the file (gs://...) and provider set to "gcs".


beeme1mr:
Hey @alan-kut, thanks for the PR. I'll test this out E2E tomorrow.

alan-kut:
Thanks. I fixed lint error and data race.

Few assumptions I made in this code:
- It will return error from Sync and Resync if the blob cannot be accessed (authorization error, doesn't exist or other problems)
- It will log warning when sync on cron cannot access the blob
- It won't fetch blob and send updated data if the updated time of the blob hasn't changed (except for ReSync)

core/go.mod
```diff
@@ -41,6 +41,10 @@ require (
 )
 
 require (
+	cloud.google.com/go v0.112.1 // indirect
+	cloud.google.com/go/compute/metadata v0.3.0 // indirect
+	cloud.google.com/go/iam v1.1.6 // indirect
+	cloud.google.com/go/storage v1.39.1 // indirect
```
toddbaert, line 47: I appreciate that we've tried to keep this neutral, and that we think it will support other cloud storage providers as well, but I'm a bit dubious about these google-cloud specific dependencies.

Can they be avoided?

Will users be happy to see these?

Will they significantly increase the binary size?

cc @beeme1mr @thisthat @Kavindu-Dodan

toddbaert: I think I understand a bit better - these are just the drivers that power the gocloud stuff, right? I guess as long as our API service is still neutral, I'm happy... as long as the bin isn't suddenly massive.

beeme1mr (approved):
I've tested this locally and it works great. My only concern is the size of the change detection log message.

Could you please [sign off your comments](https://github.com/open-feature/flagd/pull/1366/checks?check_run_id=27790626199) and update the [sync config docs](https://github.com/open-feature/flagd/blob/main/docs/reference/sync-configuration.md) when you get a moment?

core/pkg/sync/blob/blob_sync.go
```diff
@@ -0,0 +1,138 @@
...
+		if hs.lastUpdated == updated {
+			hs.Logger.Debug("configuration hasn't changed, skipping fetching full object")
+			return nil
+		}
+	}
+	msg, err := hs.fetchObject(ctx, bucket)
+	if err != nil {
+		return fmt.Errorf("couldn't get object: %v", err)
+	}
+	hs.Logger.Info(fmt.Sprintf("configuration updated: %s", msg))
```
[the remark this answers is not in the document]
alan-kut: Changed to debug.

alan-kut:
I changed the log to debug, signed off commit, updated the docs (in a new commit).

Should I squash the commits somehow?
Rebase them on main?

core/pkg/sync/blob/blob_sync.go
```diff
@@ -0,0 +1,138 @@
...
+
+	return nil
+}
+
+func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error {
+	return hs.sync(ctx, dataSync, true)
+}
+
+func (hs *Sync) sync(ctx context.Context, dataSync chan<- sync.DataSync, skipCheckingModTime bool) error {
+	bucket, err := hs.getBucket(ctx)
```
tangenti, line 71: There are a lot of empty checks on `Bucket` and `Object`, but they are not changing once the syncer is built. Would it be better to move the validation into `Init` and open the bucket in `Init` as well?

core/pkg/sync/blob/blob_sync.go
```diff
@@ -0,0 +1,138 @@
...
+		return time.Time{}, errors.New("no object string set")
+	}
+	attrs, err := bucket.Attributes(ctx, hs.Object)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("error fetching attributes for object %s/%s: %w", hs.Bucket, hs.Object, err)
+	}
+	return attrs.ModTime, nil
+}
+
+func (hs *Sync) fetchObject(ctx context.Context, bucket *blob.Bucket) (string, error) {
```
tangenti, line 121: How is `fetchObject` different from `Bucket.Download` ?

core/pkg/sync/builder/syncbuilder.go
```diff
@@ -27,6 +29,7 @@ const (
 	syncProviderGrpc       = "grpc"
 	syncProviderKubernetes = "kubernetes"
 	syncProviderHTTP       = "http"
+	syncProviderGcs        = "gcs"
```
tangenti, line 32: The implementation seems to work for other object storages as well (would require registering their drivers). Shall we rename it to something like `blob` or `objectStorage`?

core/pkg/sync/blob/blob_sync.go
alan-kut: Moved checks to init. I did it in the first place in during sync mimicking the http sync.
I think (but I'm not sure) it's safe to open bucket once for GCS but I don't know if it's the case for other drivers.

core/pkg/sync/blob/blob_sync.go
alan-kut: bucket.Download needs writer. Used it inside fetchObject to make it simpler.

core/pkg/sync/builder/syncbuilder.go
alan-kut: GCS sync provider uses blob as implementation but has it's own uri format.
The flagd users don't need to know what is the internal implementation.

core/go.mod
Kavindu-Dodan: These deps cannot be avoided as we have a direct dependency on gcs.

beeme1mr:
I've reviewed all the licensing issues detected by FOSSA. The library is licensed BSD-3, but comments in the code referred to other licenses, which triggered a false positive.

core/pkg/sync/blob/blob_sync.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil
+}
+
+func (hs *Sync) IsReady() bool {
+	return hs.ready
+}
+
+func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
+	hs.Logger.Info(fmt.Sprintf("starting sync from %s/%s with interval %d", hs.Bucket, hs.Object, hs.Interval))
```
kamil-nowosad, line 49: Nit: please log the unit (sec)

core/pkg/sync/blob/blob_sync.go
```diff
@@ -0,0 +1,131 @@
...
+		return fmt.Errorf("couldn't get bucket: %v", err)
+	}
+	defer bucket.Close()
+	var updated time.Time
+	if !skipCheckingModTime {
+		updated, err = hs.fetchObjectModificationTime(ctx, bucket)
+		if err != nil {
+			return fmt.Errorf("couldn't get object attributes: %v", err)
+		}
+		if hs.lastUpdated == updated {
```
kamil-nowosad, line 87: Optional: Perhaps it's worth logging a warning if hs.lastUpdated < updated and also not update the config in such case.

alan-kut: I'm not sure. I wanted to avoid downloading the whole blob every time.
Regarding updated time - GCS updates it automatically on any change so it should always grow.
The other system - I don't know.

However even if there is a change that decreased the time - IMO we should still switch configuration to that file. E.g. if some system would keep the time from the local file and there is a roll-back to old config and the updated time decreases - the flagd should switch to it.

kamil-nowosad: Switching config to that file SGTM. But still logging a warning in such case may be useful.

alan-kut: Added

toddbaert:
> I changed the log to debug, signed off commit, updated the docs (in a new commit).
> 
> Should I squash the commits somehow? Rebase them on main?

Commits are squashed automatically, everything will be clean in main.

toddbaert (approved):
Very cool addition.

PR #1389, opened 2024-08-29
feat: add mTLS support to otel exporter

kevinschoonover:
## This PR

The OpenTelemetry collectors in my production environment are configured to use TLS for uploading metrics / traces so this PR aims to

- add the ability to use mTLS + self-signed certificates when exporting to the opentelemetry collector

This is the 'quick and dirty' approach so wanted to make an initial PR to make sure the high level implementation is the approach you're looking for. 

### Follow-up Tasks
- [ ] update the documentation when this approach is approved

### How to test
I am struggling to figure out how to test this with self signed certificates to give a specific set of commands you can run because the TLS connection is never successful (assuming this is because of my commands)
```bash
openssl req -x509 -newkey rsa:4096 -keyout ca.key.pem -out ca.cert.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=localhost"

openssl req -x509 -newkey rsa:4096 -keyout client.key.pem -out client.cert.pem -CA ca.cert.pem -CAkey ca.key.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=localhost" -addext "subjectAltName = IP:127.0.0.1"
openssl req -x509 -newkey rsa:4096 -keyout server.key.pem -out server.cert.pem -CA ca.cert.pem -CAkey ca.key.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=localhost" -addext "subjectAltName = IP:127.0.0.1"
```

; however, when I pull certificates from my production environment to test this works


core/pkg/telemetry/builder.go
```diff
@@ -107,6 +118,38 @@ func BuildConnectOptions(cfg Config) ([]connect.HandlerOption, error) {
...
+				return nil, fmt.Errorf("can't read ca file from %s", cfg.CAPath)
+			}
+			if !capool.AppendCertsFromPEM(ca) {
+				return nil, fmt.Errorf("can't add CA '%s' to pool", cfg.CAPath)
+			}
+		}
+
+		tlsConfig := &tls.Config{
+			RootCAs: capool,
+			GetCertificate: func(chi *tls.ClientHelloInfo) (*tls.Certificate, error) {
```
kevinschoonover, line 137: for now this just re-reads the files everytime it makes a connection (which isn't very efficient). If this general approach is okay, we can take the same approach as the collector with `reload_interval` (https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/README.md#tls--mtls-configuration)

toddbaert: I'm not sure I understand this comment. Wouldn't this code only ever run once on startup?

toddbaert:
Hey @kevinschoonover this is a great addition, and I'm fine with the approach. There's some unit test failures and warnings you'll need to address (integration tests look good though).

core/pkg/telemetry/builder.go
kevinschoonover: The `GetCertificate` option to `tls.Config` is a func() (in this case a lambda) that will be called every time the TLS connection gets a ClientHello if we don't give it any certificates. Loading the capool above this section will only be done once at startup

https://pkg.go.dev/crypto/tls
```go
// GetCertificate returns a Certificate based on the given
// ClientHelloInfo. It will only be called if the client supplies SNI
// information or if Certificates is empty.
//
// If GetCertificate is nil or returns nil, then the certificate is
// retrieved from NameToCertificate. If NameToCertificate is nil, the
// best element of Certificates will be used.
//
// Once a Certificate is returned it should not be modified.
GetCertificate func(*[ClientHelloInfo](https://pkg.go.dev/crypto/tls#ClientHelloInfo)) (*[Certificate](https://pkg.go.dev/crypto/tls#Certificate), [error](https://pkg.go.dev/builtin#error))
```

toddbaert: Ahh, thanks, I missed that was a lambda being passed and not just a one time execution.

kevinschoonover:
@toddbaert - sorry for the delay. I think I have addressed the errors I saw, but don't know if i'm able to trigger the PR to fully test.

toddbaert:
There were some minor lint errors I fixed with [this commit](https://github.com/open-feature/flagd/pull/1389/commits/71943e7f5aa75ef0460add798fc2bb29c13ed6eb), please let me know if you see anything wrong.

kevinschoonover:
> There were some minor lint errors I fixed with [this commit](https://github.com/open-feature/flagd/pull/1389/commits/71943e7f5aa75ef0460add798fc2bb29c13ed6eb), please let me know if you see anything wrong.

Looks good to me! Thanks for fixing those!

beeme1mr (approved):

PR #1410, opened 2024-09-23
fix: remove dep cycle with certreloader

toddbaert:
https://github.com/open-feature/flagd/pull/1389 works well but caused a dependency cycle: core -> flagd -> core.

I've moved the `certreloader` pkg to core with the existing telemetry stuff.


kevinschoonover:
@toddbaert, thanks for finding this. Do you know why the existing tests/pipelines didn't catch this? I would love to update them so this gets caught automatically this time.

Did I just not do a build with the most recent changes and can update the PR to do a test build?

PR #1426, opened 2024-10-18
ci: fix security scan input path

beeme1mr:
A similar change was main in OFO to address the same issue.

https://github.com/open-feature/open-feature-operator/pull/713

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,9 +15,6 @@ on:
       - "README.md"
       - "docs/**"
 
-env:
-  GO_VERSION: '~1.21'
-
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -100,7 +97,7 @@ jobs:
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
         with:
-          input: /github/workspace/flagd-local.tar
+          input: ${{ github.workspace }}/flagd-local.tar
```
beeme1mr, line 100: This is the fix.

PR #1428, opened 2024-10-21
feat: support azure blob sync

wilson-matthew:
## This PR

- adds support for Azure Blob Storage sync

### How to test

- create an Azure Storage Account and a container with a flags definition file inside
- set AZURE_STORAGE_ACCOUNT env var to the name of the storage account
- set AZURE_STORAGE_KEY env var to the access key of the storage account, or use another authentication method: https://pkg.go.dev/gocloud.dev/blob/azureblob#hdr-URLs
- start flagd with the --uri flag set to azblob://my-container/my-flags.json


beeme1mr:
Hey @wilson-matthew, I'm new to Azure, so please bear with me. I tried to test using a SAS token but ran into a 403. The environment variable I set was `AZURE_STORAGE_SAS_TOKEN`, which appears to be supported by the [blob library](https://pkg.go.dev/gocloud.dev/blob/azureblob#hdr-URLs). Do you have any idea what's happening, or should I try a different auth mechanism?

wilson-matthew:
Hey @beeme1mr, I just tested using a SAS token with the AZURE_STORAGE_SAS_TOKEN environment variable and it worked okay for me. What settings did you use when generating the SAS token - I wonder if it didn’t have the correct permissions? You could try the AZURE_STORAGE_KEY auth mechanism as that is unlikely to have a permissions issue.

beeme1mr (approved):
I got it working, thanks!

toddbaert (approved):
Looks great! Thanks a lot!

PR #1448, opened 2024-11-10
feat: add context value flag

### LINKED ISSUE #1435: [FEATURE] Support addition of arbitrary properties in evaluation context
### Requirements

Currently, flagd injects a couple standard properties into the evaluation context (`$flagd.flagKey`, `$flagd.timestamp`).

In addition, we want to support the addition of ANY arbitrary string properties specified on the command line.
The command line should look something like this:

`flagd start -X name=jane -X age=31 ... ` or `flagd start --context-value name=jane --context-value age=31 ... `

The above example will add the key/value pairs `"name": "jane"` and `"age": "31"` to the evaluation context for every evaluation.

Note that values are all strings - if desired, JSONLogic can be used in rules to convert them to other types.
Also note that there's no prefix associated with these additional fields.

In addition to adding these values to the evaluation context for all evaluations done in the flagd daemon itself, flagd should return these values in the `sync-metadata` call, so that in-process providers can add these values to their evaluations as well.

- [ ] add cmd line flag
- [ ] enrich all evaluations with data from cmd line
- [ ] return data from cmd line in sync-metadata
- [ ] associated tests and README doc

alemrtv:
- add the `--context-value` command line flag to pass arbitrary key value pairs to the evaluation context

Fixes #1435 

### Follow-up Tasks
- adjust the integration test to include this flag

### How to test
issue the `flagd start` command with the -X flag and provide key value pairs like:
`flagd start -X name=jane -X age=31` or `flagd start --context-value name=jane --context-value age=31`


flagd/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -172,12 +181,20 @@ func (s *OldFlagEvaluationService) ResolveBoolean(
 	sCtx, span := s.flagEvalTracer.Start(ctx, "resolveBoolean", trace.WithSpanKind(trace.SpanKindServer))
 	defer span.End()
 	res := connect.NewResponse(&schemaV1.ResolveBooleanResponse{})
+	evalCtx := map[string]any{}
+	if e := req.Msg.GetContext(); e != nil {
+		evalCtx = e.AsMap()
+	}
+	for k, v := range s.contextValues {
+		evalCtx[k] = v
+	}
```
aepfli, line 190: This code block exists multiple times; changing how we populate the context means we must touch several parts of the code. We should extract this into our method for easier maintenance and effortless changes.

Furthermore, I think it is worth moving this code block into the resolve method and handling it once before resolving the flag.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -117,13 +119,19 @@ func extractOfrepRequest(req *http.Request) (ofrep.Request, error) {
...
+	log *logger.Logger, requestID string, request ofrep.Request, contextValues map[string]any,
+) map[string]any {
 	context := map[string]any{}
 	if res, ok := request.Context.(map[string]any); ok {
 		context = res
 	} else {
 		log.WarnWithID(requestID, "provided context does not comply with flagd, continuing ignoring the context")
 	}
 
+	for k, v := range contextValues {
```
aepfli, line 132: same topic regarding prioritization

aepfli (commented):
Thank you. Those changes look great from my perspective. I added a few remarks for ease of maintenance, which would also reduce the size of this PR, and the lines changed.

I don't have reviewer power here ;( but those are really good changes and improvements; thank you.

flagd/cmd/start.go
```diff
@@ -78,6 +78,8 @@ func init() {
 	flags.StringP(otelCAPathFlagName, "A", "", "tls certificate authority path to use with OpenTelemetry collector")
 	flags.DurationP(otelReloadIntervalFlagName, "I", time.Hour, "how long between reloading the otel tls certificate "+
 		"from disk")
+	flags.StringToStringP(contextValueFlagName, "X", map[string]string{}, "add arbitrary key value pairs "+
+		"to the flag value evaluation context")
```
beeme1mr, line 82:
```suggestion
		"to the flag evaluation context")
```

flagd/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -293,21 +301,36 @@ func (s *OldFlagEvaluationService) ResolveObject(
...
+	for k, v := range reqCtx {
+		merged[k] = v
+	}
+	return merged
+}
+
 // resolve is a generic flag resolver
 func resolve[T constraints](ctx context.Context, logger *logger.Logger, resolver resolverSignature[T], flagKey string,
 	evaluationContext *structpb.Struct, resp response[T], metrics telemetry.IMetricsRecorder,
+	configContextValues map[string]any,
```
aepfli, line 320: i do not think that we need here another value, we are merging the values with the context already on line https://github.com/open-feature/flagd/pull/1448/files#diff-65826bb36f80a7e210128a2d3493c29be6f6c6f9487feae242eccbbc61c74c4fR75 - can we simply provide the result of the other line as evaluationcontext. like 

```
evalCtx := map[string]any{}
	if e := req.Msg.GetContext(); e != nil {
		evalCtx = e.AsMap()
	}

       // add values to evalCtx here (deleted code from link) and pass only the evalCtx with all the information

	values, err := s.eval.ResolveAllValues(sCtx, reqID, evalCtx)
```
Or access `s.contextValues`  within this method and merge in here. But i do not see the need for an additional `mergeContext`, seems like we are doing this twice now.

toddbaert:
Hey @alemrtv ! Sorry for the delayed review... Many of us were away with Kubecon and still playing catch up. I'll review this next week, latest. 🙏

flagd/pkg/service/flag-evaluation/flag_evaluator.go
alemrtv: Sorry, but I'm confused. To me it looks like there're 2 different use cases: [ResolveAll](https://github.com/open-feature/flagd/blob/8c843df7714b1f2d120c5cac8e40c7220cc0c05b/flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go#L52) and the [generic resolve](https://github.com/open-feature/flagd/blob/8c843df7714b1f2d120c5cac8e40c7220cc0c05b/flagd/pkg/service/flag-evaluation/flag_evaluator.go#L297). And these don't seem to be called at the same time. So there're two places where we need to handle it.

aepfli: My bad, you are totally right sorry

aepfli: I am sorry, you are totally right. - I mixed those methods up.

flagd/pkg/service/flag-sync/handler.go
```diff
@@ -60,7 +61,8 @@ func (s syncHandler) GetMetadata(_ context.Context, _ *syncv1.GetMetadataRequest
 	*syncv1.GetMetadataResponse, error,
 ) {
 	metadata, err := structpb.NewStruct(map[string]interface{}{
-		"sources": s.mux.SourcesAsMetadata(),
+		"sources":        s.mux.SourcesAsMetadata(),
+		"context_values": s.contextValues,
```
toddbaert, line 65: Instead of this property, can you expand the s.contextValues into the existing map, with the existing "sources" taking priority in case of conflicts (if there's another key called "sources" I mean)?

This field `context_values` is not specified anywhere, and currently, a few of our providers are adding all the properties here to their context, not specifically looking for this field.

toddbaert: Thanks!

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -117,10 +119,18 @@ func extractOfrepRequest(req *http.Request) (ofrep.Request, error) {
...
+	context := make(map[string]any)
+	for k, v := range contextValues {
+		context[k] = v
+	}
+
 	if res, ok := request.Context.(map[string]any); ok {
-		context = res
+		for k, v := range res {
+			context[k] = v
+		}
```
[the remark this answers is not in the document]
toddbaert: OK don't consider this a blocker. We can discuss a bit more before release.

toddbaert:
I've pushed a [small change](https://github.com/open-feature/flagd/pull/1448/commits/95d0048ad3b80d230b66c16e55844a07bb8e348d) to fix the lint/doc CI failures as well as adding a new make target.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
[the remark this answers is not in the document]
aepfli: Well, this is an interesting discussion point, and after reading through the message and thinking a little bit, both scenarios make sense. On the one hand, it is never a good idea to change the users' expected behavior and overwrite their provided inputs without a notification (we don't have a way to give a notification). On the other hand, it makes sense to enforce it in specific scenarios from an operational perspective. As you want to ensure, the properties do have this value. 

As I think both variations are valid options, should we maybe give the users the possibility to change the behavior to the other one with another flag?

aepfli (approved):

beeme1mr (approved):

flagd/pkg/service/flag-evaluation/ofrep/handler.go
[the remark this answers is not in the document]
aepfli: I think that is hard to grasp and compare, as we do have a merge order, and the inProcess evaluation, has the possibility to do this withint he merge order (it is provider for the metadata if I remember correctly) and here we are either first or last. Hence I think we can not compare it

toddbaert:
Hey @alemrtv , thanks again. I pushed up one more [small change](https://github.com/open-feature/flagd/pull/1448/commits/c6d3f5c7d573d21c3af4d5efdf195c8f4037c0ac) which again changes the priority of the "configuration level" context attributes (we'll call them static attributes) vs incoming request based context attributes. I will explain why here for your understanding and for posterity:

Basically, we have 2 modes of operation for flagd providers (clients): `in-process` and `rpc`. In RPC mode these static attributes are injected into flagd's evaluation logic (this is what you've done), while in `in-process` mode, these values are delivered via the `GetSyncMetadata` call, and the providers on the other end of the wire [inject these](https://github.com/open-feature/java-sdk-contrib/blob/c8435e7182c3f351e8ae31492226aa0d99d126a3/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/FlagdOptions.java#L148) properties into their local evaluation. The way they do that is via a provider hook, which has the highest priority in terms of context merging. We can't change that pattern, so for consistency, we've opted to also do the same thing for evaluations in flagd (RPC mode). I hope that makes sense! I explained the same to @beeme1mr and @aepfli and they are on the same page.

Again, thanks for your help. I will merge this PR this week.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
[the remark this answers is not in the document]
toddbaert: see: https://github.com/open-feature/flagd/pull/1448#issuecomment-2518548007

PR #1494, opened 2024-12-27
build(renovate): Utilize default OpenFeature Renovate configuration

aepfli:
We do have a default OpenFeature Renovate configuration within our community-tooling repository. (https://github.com/open-feature/community-tooling/blob/main/renovate.json)

To reduce maintenance efforts, we should stick to the general one as a basis.


aepfli:
@beeme1mr fyi

PR #1522, opened 2025-01-17
feat: support yaml in blob, file, and http syncs

### LINKED ISSUE #1512: [BUG] inconsistent yaml support from `file://` vs `s3://`
### Observed behavior

Hi folks,

I've run into some trouble trying to load YAML flag definitions from S3. Looking through the docs and repo, it _seems_ like this should be supported, and indeed it is for local files, but when trying to load the YAML file from S3, it appears that the JSON evaluator is being invoked without a call to something like `yamlToJSON(...)`.

```console
docker run --rm -it --mount type=bind,source=$(pwd)/flags.flagd.yaml,target=/flags.flagd.yaml ghcr.io/open-feature/flagd:v0.11.5 start --log-format json --uri file:/flags.flagd.yaml
```

```json
{"level":"info","ts":"2025-01-14T23:57:48.739Z","caller":"file/filepath_sync.go:62","msg":"Starting filepath sync notifier","component":"sync","sync":"fileinfo"}
{"level":"info","ts":"2025-01-14T23:57:48.741Z","caller":"file/filepath_sync.go:101","msg":"watching filepath: /flags.flagd.yaml","component":"sync","sync":"fileinfo"}
```

```console
docker run --rm -it ghcr.io/open-feature/flagd:v0.11.5 start --log-format json --uri s3://${BUCKET_NAME}/flags.flagd.yaml
```

```json
{"level":"info","ts":"2025-01-14T22:54:49.922Z","caller":"blob/blob_sync.go:51","msg":"starting sync from s3://***//flags.flagd.yaml with interval 5s","component":"sync","sync":"s3"}
{"level":"warn","ts":"2025-01-14T22:54:54.923Z","caller":"flag-sync/sync_service.go:82","msg":"timeout while waiting for all sync sources to complete their initial sync. continuing sync service","component":"FlagSyncService"}
{"level":"warn","ts":"2025-01-14T22:54:55.011Z","caller":"evaluator/json.go:112","msg":"failed to execute JSON schema validation: invalid character '$' looking for beginning of value"}
{"level":"error","ts":"2025-01-14T22:54:55.013Z","caller":"runtime/runtime.go:146","msg":"transposing evaluators: unmarshal: invalid character '$' looking for beginning of value","component":"runtime","stacktrace":"github.com/open-feature/flagd/flagd/pkg/runtime.(*Runtime).updateAndEmit\n\t/src/flagd/pkg/runtime/runtime.go:146\ngithub.com/open-feature/flagd/flagd/pkg/runtime.(*Runtime).Start.func1\n\t/src/flagd/pkg/runtime/runtime.go:56\ngolang.org/x/sync/errgroup.(*Group).Go.func1\n\t/go/pkg/mod/golang.org/x/sync@v0.10.0/errgroup/errgroup.go:78"}
```

Here's a stripped down version of `flags.flagd.yaml` as it appeared at the time of these^ logs. The specific error messages relating to `$` are obviously complaining about the `$schema`, but removing `$schema` still fails, just with different this-isnt-json errors. Also, just to reiterate, this file works when used via `flagd start --uri file:flags.flagd.yaml`

```yaml
$schema: https://flagd.dev/schema/v0/flags.json

flags:
  some_flag:
    defaultVariant: "true"
    state: ENABLED
    variants:
      "true": "true"
      "false": "false"
```

Related: #180, #206

### Expected Behavior

_No response_

### Steps to reproduce

_No response_

beeme1mr:
## This PR

- adds support for yaml in blog syncs 
- rejects non-200 responses in the http sync

### Related Issues

Fixes #1512

### How to test

`make test` or `make test-core`


core/pkg/sync/blob/blob_sync.go
```diff
@@ -131,6 +135,20 @@ func (hs *Sync) fetchObject(ctx context.Context, bucket *blob.Bucket) (string, e
...
-	return buf.String(), nil
+	switch hs.fileType {
+	case "yaml", "yml":
+		return utils.YAMLToJSON(buf.Bytes())
+	case "json":
+		return buf.String(), nil
+	default:
+		// TODO make consistent with the file sync
+		return buf.String(), nil
+		// return "", fmt.Errorf("filepath extension for URI: '%s' is not supported", hs.sync())
```
beeme1mr, line 151: This retains the current behavior but is inconsistent with the file sync implementation. I'd prefer to return an error if the file extension is invalid or undefined but it would be a breaking change.

Do not merge until a decision has been made on the expected behavior.

aepfli: I agree that throwing here would be a good idea. But we could also do a slow transition, and for now, only log an error and a deprecation warning that the current file is not supported, and in future versions, this will cause an error. wdyt?

aepfli (approved):
looks good to me, i share your thoughts about the breaking change, but i added a suggestion in the comment

core/pkg/sync/blob/blob_sync.go
toddbaert: I think we should just add the break. We are still sub zero and I don't think it will effect many people, and the resolution is very clear.

core/pkg/sync/blob/blob_sync.go
```diff
@@ -131,6 +135,23 @@ func (hs *Sync) fetchObject(ctx context.Context, bucket *blob.Bucket) (string, e
 	if err != nil {
 		return "", fmt.Errorf("error downloading object %s/%s: %w", hs.Bucket, hs.Object, err)
 	}
+	if hs.fileType == "" {
+		uriSplit := strings.Split(hs.Object, ".")
+		hs.fileType = uriSplit[len(uriSplit)-1]
+	}
```
austindrenski, line 141: It may be out of scope for this PR, but bringing it up for further/future consideration that we could also take into consideration the blob's content-type header. You would then have to decide which takes precedence when the header and extension contradict, so it would complicate things, but I'm imagining a case where I've uploaded a blob object with proper headers but with a generated key that doesn't have a file extension.

Not sure if any of the other syncs take that into account, so probably best not to introduce here unless/until it can be made consistent with other syncs (at least the ones where a content-type header can be passed).

So, anyways, just some food for thought, but, after all, it's not an unreasonable requirement for files to have extensions, just highlighting that there could be cases where you're consuming a blob that might, by default, not have a file extension generated.

_(For example, I've encountered terraform modules in the wild that will default to storing blobs without extensions. Now, in my opinion, that is chaotic nonsense, but idk, some folks just want to watch the world burn, so what can you do 🤷)_

austindrenski (approved):
Awesome, thanks for implementing this!

core/pkg/sync/blob/blob_sync.go
toddbaert: I agree on all counts, but I'm a bit confused (possibly just because we are using different terminology for the same thing). Are you referring to file MIME-types? As in, the result of this command on most linux systems:

```shell
file  --mime-type flagd/config/samples/example_flags.json  #/home/todd/git/flagd/config/samples/example_flags.json: application/json
```

Or do you mean something do do with HTTP headers from blobs provided by cloud providers? If you mean the latter, I'm not sure how this would impact anything other than blob syncs. Perhaps the former could be generally useful though... it's probably better than file extensions, TBH.

toddbaert (approved):
LGTM @beeme1mr . I merged main to fix some conflicts.

I'll let you choose if you consider [this](https://github.com/open-feature/flagd/pull/1522#discussion_r1920441179) "decided".

core/pkg/sync/blob/blob_sync.go
beeme1mr: Yeah, i think that's a good idea. I'll see how easy it is to determine the content type from the blog storage. Unfortunately, it looks like YAML was [only recently](https://httptoolkit.com/blog/yaml-media-type-rfc/) defined as a content type in IANA so it may be inconsistently implemented by cloud vendors.

austindrenski: @toddbaert [wrote](https://github.com/open-feature/flagd/pull/1522#discussion_r1922816832):
> I agree on all counts, but I'm a bit confused (possibly just because we are using different terminology for the same thing). Are you referring to file MIME-types? [...]

Oh that's interesting, but definitely not thinking we'd sniff the files with `file --mime-type`.
 
> Or do you mean something do do with HTTP headers from blobs provided by cloud providers? If you mean the latter, I'm not sure how this would impact anything other than blob syncs. Perhaps the former could be generally useful though... it's probably better than file extensions, TBH.

Rather, I was thinking about the existing [Sync Configuration docs](https://flagd.dev/reference/sync-configuration/), which imply that I could serve any old file from any old web app by way of `--uri https://....` _(still not sure if grpc/k8s, i.e., `--uri grpcs://...`/`--uri core.openfeature.dev://` would return anything like an HTTP Content-Type header, but if they do, then the more the merrier)_.

So, for example, I could imagine a scenario where it could be useful to serve flag definitions from a web app _(which itself might be accessing blob storage, or dynamically constructing flag definitions from a database, etc.)_, and then start flagd with `--uri https://my-web-app/some_flags`, and have flagd merrily request and react based on the web app's response headers: `Content-Type: application/yaml` (accepted), `Content-Type: application/json` (accepted), or `Content-Type: application/some-unknown-stuff` (flagd fails to load this and logs an error, etc.).

If this all works, then flagd could even send out an accept header when making the sync requests, e.g., `Accept: application/json, application/yaml, application/x-yaml, text/yaml`, which would help with @beeme1mr's point about how recently YAML's official MIME type was defined.

beeme1mr: I've added media type support. Tomorrow I'll manually test the various sources to ensure I didn't break anything that aren't covered in our test suite.

docs/reference/sync-configuration.md
austindrenski, line 1: Can add a suggestion directly, but there's one more `http://my-flag-source.json` on L97 you might want to grab while you're in there

austindrenski (approved):
Only had time to skim through tonight, but everything looks to be in order. The convert to json tests are a nice addition and case coverage looks thorough!

aepfli (approved):

core/pkg/sync/http/http_sync.go
```diff
@@ -141,12 +145,31 @@ func (hs *Sync) fetchBodyFromURL(ctx context.Context, url string) ([]byte, error
 		}
 	}()
 
+	statusOK := resp.StatusCode >= 200 && resp.StatusCode < 300
+	if !statusOK {
+		return "", fmt.Errorf("error fetching from url %s: %s", url, resp.Status)
+	}
```
beeme1mr, line 151: We weren't erroring on non-200 status codes before.

core/pkg/sync/blob/blob_sync.go
beeme1mr: Manual tests look good.

PR #1540, opened 2025-01-29
feat!: flagSetMetadata in OFREP/ResolveAll, core refactors

toddbaert:
:warning: This PR brings a breaking change to the [flagd-core](https://pkg.go.dev/github.com/open-feature/flagd/core) library: the `IStore` interface now returns an additional value representing the flag set metadata. There are no breaking changes in flagd's behavior.

Changes in flagd:

- returns flag set metadata as metadata for error flags (best effort)
- returns flag set metadata in OFREP and RPC calls
- moves metadata merging logic to evaluator (all flags inherent flag set metadata, but can override, as as before but now reusable in flagd core)
- removes duplicated flag set metadata keys when the same flag set metadata key exists in multiple sources

### To Test

- requires `curl`, `grpcurl`, and `jq`

#### RPC

```shell
grpcurl -import-path  /...../schemas/protobuf/flagd/evaluation/v1  -proto evaluation.proto -plaintext  localhost:8013 flagd.evaluation.v1.Service/ResolveAll | jq
```

### OFREP

```shell
curl --location 'http://localhost:8016/ofrep/v1/evaluate/flags'  --header 'Content-Type: application/json' --data '{"context": {"color": "yellow"}}'  | jq
```

### Sync

```shell
grpcurl -import-path /...../schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext   localhost:8015 flagd.sync.v1.FlagSyncService/FetchAllFlags    | jq -r .flagConfiguration  | jq
```


core/pkg/evaluator/json_test.go
```diff
@@ -1184,222 +1264,6 @@ func TestState_Evaluator(t *testing.T) {
 			expectedError:  true,
 			expectedResync: false,
 		},
-		"flag metadata": {
```
toddbaert, line 1187: The idea of these tests is now moved above (metadata merging).

core/pkg/evaluator/json.go
```diff
@@ -119,15 +119,16 @@ func (je *JSON) SetState(payload sync.DataSync) (map[string]interface{}, bool, e
 	var events map[string]interface{}
 	var reSync bool
 
+	// TODO: handle other types
```
beeme1mr, line 122: Is this TODO still relevant?

flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go
```diff
@@ -107,15 +108,29 @@ func (s *FlagEvaluationService) ResolveAll(
...
-				Variant: value.Variant,
+			res.Flags[resolved.FlagKey] = &evalV1.AnyFlag{
+				Reason:  resolved.Reason,
+				Variant: resolved.Variant,
 				Value: &evalV1.AnyFlag_ObjectValue{
 					ObjectValue: val,
 				},
 			}
 		}
+		// TODO: error handling
```
beeme1mr, line 119: Is there additional error handling required?

beeme1mr (approved):
There are a couple of TODO that should either be removed or actioned. The change looks great and the functionality aligns with that JS implementation. Thanks!

flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go
toddbaert: No this is left over.

core/pkg/evaluator/json.go
toddbaert: Yes... Currently, we don't handle metadata in ADD or UPDATE syncs.

However, there's some other issues with these already, and they are not used in any implementation I know of (they are only really applicable to GRPC). I honestly think we should deprecate them. I will improve the comment and see what others think.

flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go
toddbaert, line 119:
```suggestion
```

core/pkg/evaluator/json.go
toddbaert: I updated this comment; I will open a new discussion about it.

core/pkg/evaluator/json.go
```diff
@@ -119,15 +119,16 @@ func (je *JSON) SetState(payload sync.DataSync) (map[string]interface{}, bool, e
...
+		events, reSync = je.store.Merge(je.Logger, payload.Source, payload.Selector, definition.Flags, definition.Metadata)
 	case sync.ADD:
-		events = je.store.Add(je.Logger, payload.Source, payload.Selector, newFlags.Flags)
+		events = je.store.Add(je.Logger, payload.Source, payload.Selector, definition.Flags)
 	case sync.UPDATE:
-		events = je.store.Update(je.Logger, payload.Source, payload.Selector, newFlags.Flags)
+		events = je.store.Update(je.Logger, payload.Source, payload.Selector, definition.Flags)
 	case sync.DELETE:
-		events = je.store.DeleteFlags(je.Logger, payload.Source, newFlags.Flags)
+		events = je.store.DeleteFlags(je.Logger, payload.Source, definition.Flags)
```
toddbaert, line 131: ADD/UPDATE are only relevant to hypothetical gRPC server implementations, and were a premature optimization to allow gRPC servers to tell flagd "just add these flags, or just change this flag" instead of sending a whole payload. I have never seen this implemented, and it's not relevant to any of the other sync types.

I think we should cease to enhance them and deprecate them in the protobuf.

core/pkg/evaluator/json.go
```diff
@@ -156,14 +157,16 @@ func NewResolver(store store.IStore, logger *logger.Logger, jsonEvalTracer trace
 	return Resolver{store: store, Logger: logger, tracer: jsonEvalTracer}
 }
 
-func (je *Resolver) ResolveAllValues(ctx context.Context, reqID string, context map[string]any) ([]AnyValue, error) {
+func (je *Resolver) ResolveAllValues(ctx context.Context, reqID string, context map[string]any) ([]AnyValue,
```
alexandraoberaigner, line 160: [Q] is this the function that relates to this proto command (from evaluation.proto):
`  rpc ResolveAll(ResolveAllRequest) returns (ResolveAllResponse) {}
`

core/pkg/evaluator/json_test.go
```diff
@@ -372,6 +395,62 @@ func TestResolveAllValues(t *testing.T) {
...
+	}
+	const reqID = "default"
+	evaluator := evaluator.NewJSON(logger.NewLogger(nil, false), store.NewFlags())
+	_, _, err := evaluator.SetState(sync.DataSync{FlagData: Flags})
+	if err != nil {
+		t.Fatalf("expected no error")
+	}
+
+	for _, test := range tests {
+		_, _, _, metadata, _ := evaluator.ResolveBooleanValue(context.TODO(), reqID, test.flagKey, nil)
```
alexandraoberaigner, line 414: [Q] What does context.TODO() do?

core/pkg/store/flags.go
```diff
@@ -303,3 +318,40 @@ func (f *Flags) Merge(
 	}
 	return notifications, resyncRequired
 }
+
+func (f *State) getMetadataForSource(source string) model.Metadata {
```
alexandraoberaigner, line 322: [Q] What does `source` represent?

alexandraoberaigner (commented):
Better late then never...I just finished reviewing & added a few questions :)

schemas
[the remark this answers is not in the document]
toddbaert: I see it when I try - I suspect that the proto files you are pointing to in the command you are running are out of date, and don't have the latest additions of metadata. `cd` into the schemas repo and pull it. We only [very recently added](https://github.com/open-feature/flagd-schemas/pull/180) this field to the proto.

core/pkg/evaluator/json.go
```diff
@@ -119,15 +119,16 @@ func (je *JSON) SetState(payload sync.DataSync) (map[string]interface{}, bool, e
 	var events map[string]interface{}
 	var reSync bool
 
+	// TODO: We do not handle metadata in ADD/UPDATE operations. These are only relevant for grpc sync implementations.
```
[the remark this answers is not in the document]
toddbaert: You are understanding correctly. I left it as a TODO because I wanted to open an issue to discuss the deprecation of these, which I still intend to do.

core/pkg/evaluator/json.go
toddbaert: Yes, this is the inner "domain layer" function satisfying that rpc. More directly, [this function](https://github.com/open-feature/flagd/blob/986a436e10e9766b897319085cf7dbbe2f10cb24/flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go#L55) is the actual rpc endpoint.

core/pkg/evaluator/json_test.go
toddbaert: `context` in Go is sort of like a very explicit `ThreadLocal` in Java, or a continuation in Javascript; it's passed down the entire call chain (in most cases) and can be used to propagate signals and store data (for example, request data).

`context.TODO()` is sort of a "bare" or "root" context often used in testing or at the very start of some process.

core/pkg/store/flags.go
toddbaert: It's a "source" of flags. For example, when you start flagd like this:

`flagd start -f file:../config/samples/example_flags.flagd.json -f http://some-domain.com/my-flags.json`

you have set 2 sources (one file and one http).

See: https://flagd.dev/concepts/syncs/

schemas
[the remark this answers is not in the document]
alexandraoberaigner: Ah, I see, makes sense! Thanks for explaining :)

core/pkg/evaluator/json_test.go
alexandraoberaigner: I should finish my Go course 😁 I thought it was an OpenFeature context attribute

toddbaert: That's understandable since there's a name collision here. It's not ideal. I try to refer to the OpenFeature one as "EvaluationContext" to disambiguate them.

PR #1567, opened 2025-02-18
docs: add providerID as a configuration option

beeme1mr:
## This PR

- adds `providerID` as a valid configuration option

### Related Issues

Relates to https://github.com/open-feature/go-sdk-contrib/pull/626

### Notes

`provider_id` is supported in [sync request proto](https://github.com/open-feature/flagd-schemas/blob/main/protobuf/flagd/sync/v1/sync.proto#L27). This change specifics how the properties could be set in the provider.

### Follow-up Tasks

Update test bed [config.feature](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/config.feature).

### DIFF
--- a/docs/reference/specifications/providers.md
+++ b/docs/reference/specifications/providers.md
@@ -263,7 +263,7 @@ precedence.
 Below are the supported configuration parameters (note that not all apply to both resolver modes):
 
 | Option name           | Environment variable name      | Explanation                                                            | Type & Values                | Default                       | Compatible resolver     |
-|-----------------------|--------------------------------|------------------------------------------------------------------------|------------------------------|-------------------------------|-------------------------|
+| --------------------- | ------------------------------ | ---------------------------------------------------------------------- | ---------------------------- | ----------------------------- | ----------------------- |
 | resolver              | FLAGD_RESOLVER                 | mode of operation                                                      | String - `rpc`, `in-process` | rpc                           | rpc & in-process        |
 | host                  | FLAGD_HOST                     | remote host                                                            | String                       | localhost                     | rpc & in-process        |
 | port                  | FLAGD_PORT                     | remote port                                                            | int                          | 8013 (rpc), 8015 (in-process) | rpc & in-process        |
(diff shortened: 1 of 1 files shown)


docs/reference/specifications/providers.md
```diff
@@ -280,6 +280,7 @@ Below are the supported configuration parameters (note that not all apply to bot
 | cache                 | FLAGD_CACHE                    | enable cache of static flags                                           | String - `lru`, `disabled`   | lru                           | rpc                     |
 | maxCacheSize          | FLAGD_MAX_CACHE_SIZE           | max size of static flag cache                                          | int                          | 1000                          | rpc                     |
 | selector              | FLAGD_SOURCE_SELECTOR          | selects a single sync source to retrieve flags from only that source   | string                       | null                          | in-process              |
+| providerID            | FLAGD_PROVIDER_ID              | A unique identifier for flagd(grpc client) initiating the request.     | string                       | null                          | in-process              |
```
aepfli, line 283: Generally I like the idea, out of curiosity, could this be the domain of the provider?

aepfli: Nvm I see the purpose now to make it better visible on flagd to see from whom the request was done.

chrfwow (approved):

docs/reference/specifications/providers.md
beeme1mr, line 283: `providerId`, `providerID`, or whatever fit the language best? We may not want to prescribe the names because the language may dictate (e.g. Python would be `provider_id`).

aepfli: For the gherkin tests in flagd_testbed we do have an implementation to transform camelcase to Kebapcase, which makes the testing easier, especially if we stick to one specific implementation. I would suggest to use `procividerId` as this will get automagically gets transformed to `provider_id` where as `providerID` might end up as `provider_i_d`.

I know we would need to change some implementations but I think for the ease of transformation, we should stick to a generic approach, which maps to all property name transformations for all kind of languages.

aepfli: See https://github.com/open-feature/python-sdk-contrib/blob/7048589b501f57dcd7ecd5a2cf6f2fcec55cde88/providers/openfeature-provider-flagd/tests/e2e/step/config_steps.py#L11

beeme1mr: I'm fine with that. FYI @toddbaert @cupofcat

cupofcat: I am fine with using Id (camelcase) as long as we are consistent. Perhaps, for languages where it's more natural to have ID, we could provide an alias?

aepfli (approved):

cupofcat (approved):

PR #1589, opened 2025-03-12
feat: accept version numbers which are not strings

chrfwow:
## This PR
Fixes errors when the provided version is not a string but a number (e.g. versions like 1.0, 2, ...) and improves the test suite to use our `parseSemverEvaluationData` function. Furthermore, logging now includes erroneous inputs.


flagd/pkg/service/flag-sync/test-cert/gen.sh
```diff
@@ -13,5 +13,5 @@ openssl req -x509 -newkey rsa:4096 -days 9999 -nodes -keyout ca-key.pem -out ca-
 openssl req -newkey rsa:4096 -nodes -keyout server-key.pem -out server-req.pem -subj "/CN=flagD test server PR and CSR"
 
 # 3. Use CA's private key to sign web server's CSR and get back the signed certificate
-openssl x509 -req -in server-req.pem -days 60 -CA ca-cert.pem -CAkey ca-key.pem -CAcreateserial -out server-cert.pem -extfile server-ext.cnf
+openssl x509 -req -in server-req.pem -days 9999 -CA ca-cert.pem -CAkey ca-key.pem -CAcreateserial -out server-cert.pem -extfile server-ext.cnf
```
aepfli, line 16: we needed to update the certs, as they have been running out - as we only generated valid certs for 60 days :see_no_evil:

core/pkg/evaluator/semver.go
```diff
@@ -115,17 +115,26 @@ func parseSemverEvaluationData(values interface{}) (string, string, SemVerOperat
...
+	if str, ok := v.(string); ok {
+		// It's already a string
+		return str
+	}
+	// Convert to string if not already
+	return fmt.Sprintf("%v", v)
+}
+
 func parseSemanticVersion(v interface{}) (string, error) {
 	version, ok := v.(string)
```
toddbaert, line 135: This should never be possible anymore, and this function is private, so it probably won't happen, but I suppose it's still a good safeguard if it's made public. Up to you if you want to just combine `ensureString` and `parseSemanticVersion` or not.

toddbaert (approved):
I left a very minor nit, which is completely optional and I leave to your judgement.

Excellent test suite. I'm not sure all of this is captured in our e2e suite, but it should be. I will create an issue to support numeric context values in our test suite, as well as to make sure some of the other semver edge-cases are covered. Thanks a lot!

toddbaert:
I've marked this as a `feat` not a `fix` because I don't think we ever explicitly supported this before, though I fully agree we should.

PR #1593, opened 2025-03-14
feat: add traces to ofrep endpoint

aepfli:
## This PR
our OFREP endpoint was not offering any kind of otel information, now we have traces
image

### Related Issues

Relates: https://github.com/open-telemetry/opentelemetry-demo/pull/2119


flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -57,9 +65,10 @@ func (h *handler) HandleFlagEvaluation(w http.ResponseWriter, r *http.Request) {
 		h.writeJSONToResponse(http.StatusBadRequest, ofrep.ContextErrorResponseFrom(flagKey), w)
 		return
 	}
-
 	context := flagdContext(h.Logger, requestID, request, h.contextValues)
-	evaluation := h.evaluator.ResolveAsAnyValue(r.Context(), requestID, flagKey, context)
+	evaluation := h.evaluator.ResolveAsAnyValue(rCtx, requestID, flagKey, context)
+	span.SetAttributes(attribute.String("feature_flag.key", evaluation.FlagKey))
```
aepfli, line 70: i am not sure if we want to add this information here ;)

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -23,25 +27,29 @@ type handler struct {
...
 	router.HandleFunc(bulkEvaluation, h.HandleBulkEvaluation).Methods("POST")
-	return router
+	return otelhttp.NewHandler(router, "flagd.ofrep")
 }
 
 func (h *handler) HandleFlagEvaluation(w http.ResponseWriter, r *http.Request) {
 	requestID := xid.New().String()
 	defer h.Logger.ClearFields(requestID)
 
+	rCtx, span := h.tracer.Start(r.Context(), "flagEvaluation", trace.WithSpanKind(trace.SpanKindServer))
```
aepfli, line 51: i am also not sure if we want to wrap this in another span ;)

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -72,21 +81,24 @@ func (h *handler) HandleBulkEvaluation(w http.ResponseWriter, r *http.Request) {
...
-	evaluations, metadata, err := h.evaluator.ResolveAllValues(r.Context(), requestID, context)
+	evaluations, metadata, err := h.evaluator.ResolveAllValues(rCtx, requestID, context)
 	if err != nil {
 		h.Logger.WarnWithID(requestID, fmt.Sprintf("error from resolver: %v", err))
 
 		res := ofrep.BulkEvaluationContextErrorFrom(model.GeneralErrorCode,
 			fmt.Sprintf("Bulk evaluation failed. Tracking ID: %s", requestID))
 		h.writeJSONToResponse(http.StatusInternalServerError, res, w)
 	} else {
+		span.SetAttributes(attribute.Int("feature_flag.count", len(evaluations)))
```
aepfli, line 101: this is actually already set by the evaluator, and we can remove it from here, hence the whole span for this action might not be needed at all

flagd/pkg/service/flag-evaluation/ofrep/handler.go
pichlermarc: Hmm, I'm not fully up to speed on feature flag semantic conventions, but they [seem to apply to logs rather than spans](https://github.com/open-telemetry/semantic-conventions/blob/6533b8a39e03e6925e080d5ca39234035cf87e70/docs/feature-flags/feature-flags-logs.md). 

Could it be that a log that can be correlated to the `traceId` is what you're trying to write to instead?

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -72,21 +81,24 @@ func (h *handler) HandleBulkEvaluation(w http.ResponseWriter, r *http.Request) {
 	requestID := xid.New().String()
 	defer h.Logger.ClearFields(requestID)
 
+	rCtx, span := h.tracer.Start(r.Context(), "bulkEvaluation", trace.WithSpanKind(trace.SpanKindServer))
```
pichlermarc, line 84: This is also very likely an internal span.

pichlermarc (commented):
I have some limited knowledge about both OpenFeature and feature flagging semconv, please ignore my comments wherever it does not make any sense :)

Overall, I'm wondering this PR should actually add logs for the internal operations instead, only keeping the server span for handling the HTTP request.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
aepfli: i will remove those spans for now, and the attributes, i think it is better to have some kind of information, where the span ended up, but nothing more for the begining

aepfli:
> I have some limited knowledge about both OpenFeature and feature flagging semconv, please ignore my comments wherever it does not make any sense :)
> 
> Overall, I'm wondering this PR should actually add logs for the internal operations instead, only keeping the server span for handling the HTTP request.

thank you for the great insights, i will limit this pr now, to only offer spans, but not adding any kind of additional information. Thank you

toddbaert:
> > I have some limited knowledge about both OpenFeature and feature flagging semconv, please ignore my comments wherever it does not make any sense :)
> > Overall, I'm wondering this PR should actually add logs for the internal operations instead, only keeping the server span for handling the HTTP request.
> 
> thank you for the great insights, i will limit this pr now, to only offer spans, but not adding any kind of additional information. Thank you

I'm fine with limiting this pr, but I think I agree with @pichlermarc that we should limit span use and add more events (logs) before we release these changes. I think changing the nature of our telemetry to much can cause thrashing for monitoring setups.

What do you think @aepfli cc @beeme1mr

toddbaert:
Thanks for working on this @aepfli .

What do you think about this? https://github.com/open-feature/flagd/pull/1593#issuecomment-2726804479

aepfli:
> Thanks for working on this @aepfli .
> 
> What do you think about this? [#1593 (comment)](https://github.com/open-feature/flagd/pull/1593#issuecomment-2726804479)

with this pr, we are now generating traces, beforehand there was nothing - so i think it is a good starting point, especially for the ofrep endpoint, which is often related with a span from another system anyways

beeme1mr (approved):

beeme1mr:
I think this is the right approach. We can easily add spans and custom attributes later once the use case becomes clear. Thanks?

pichlermarc (approved):

aepfli:
@open-feature/cloud-native-maintainers wdyt is this worth merging?

lukas-reining (approved):
Looks good!

PR #1595, opened 2025-03-17
feat: add traces to sync streams

aepfli:
adding traces to sync stream


aepfli:
@beeme1mr this would be the same thing as for ofrep, but excluding the stream endpoint, as this is a long span - we could manually create a span per each configuration update, not sure if that is worth it

aepfli:
do we want to keep this, or should we close this pr @beeme1mr

core/pkg/evaluator/json.go
```diff
@@ -100,19 +98,10 @@ func (je *JSON) GetState() (string, error) {
 }
 
 func (je *JSON) SetState(payload sync.DataSync) (map[string]interface{}, bool, error) {
-	_, span := je.jsonEvalTracer.Start(
```
aepfli, line 103: this traces actually do not offer a lot of value - it is a short span which offers no insights at all. the new approach spans over new data received till sending, and even enriches on the sync server messages with the information, and this should also help for rpc evaluations

flagd/pkg/runtime/runtime.go
```diff
@@ -141,20 +145,34 @@ func (r *Runtime) updateAndEmit(payload sync.DataSync) bool {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
+	tp := otel.GetTracerProvider()
```
aepfli, line 148: on each updateAndEmit we start a span now, works for all sources

toddbaert:
I'll check this out this weekend!

flagd/pkg/runtime/runtime.go
```diff
@@ -141,20 +145,34 @@ func (r *Runtime) updateAndEmit(payload sync.DataSync) bool {
...
+	tracer := tp.Tracer("flagd-inprocess-sync")
+	ctx, span := tracer.Start(context.Background(), "flagd flagset update",
+		trace.WithSpanKind(trace.SpanKindProducer),
+		trace.WithAttributes(attribute.String("feature_flag.source", payload.Source)),
+		trace.WithAttributes(attribute.String("feature_flag.selector", payload.Selector)),
+		trace.WithAttributes(attribute.String("feature_flag.sync_type", payload.Type.String())),
```
toddbaert, line 154: I think this attributes are very flagd specific, I'm not sure if we should perhaps use our own custom namespace for them?

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -201,6 +210,9 @@ func (r *Multiplexer) reFill() error {
 	for source, flags := range collector {
 		// store the corresponding metadata
 		metadata := r.store.GetMetadataForSource(source)
+		metadata["propagationTraceId"] = spanContext.TraceID()
+		metadata["propagationSpanId"] = spanContext.SpanID()
```
toddbaert, line 214: I love this idea. I want to think very carefully about our naming though. I think `propagationTraceId` and `propagationSpanId` might be a bit generic. Maybe something more like `configurationChangedTraceId` and `configurationChangedSpanId` since our event name is `PROVIDER_CONFIGURATION_CHANGED`.

flagd/pkg/runtime/runtime.go
```diff
@@ -141,20 +145,34 @@ func (r *Runtime) updateAndEmit(payload sync.DataSync) bool {
...
 	notifications, resyncRequired, err := r.Evaluator.SetState(payload)
 	if err != nil {
 		r.Logger.Error(err.Error())
+		span.SetStatus(codes.Error, "flagSync error")
+		span.RecordError(err)
 		return false
 	}
 
+	// Number of notifications correlates to the number of flags changed through this sync, record it
+	span.SetAttributes(attribute.Int("feature_flag.change_count", len(notifications)))
```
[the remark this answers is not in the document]
toddbaert: Should we put other attributes here as well, like the flag set version and id?

flagd/pkg/service/flag-sync/sync-multiplexer.go
aepfli: I am not sure I agree, because the `PROVIDER_CONFIGURATION_CHANGED` is also a provider implementation detail. we also add this information on the first connection, which is a provider ready event.

flagd/pkg/runtime/runtime.go
aepfli: i like the idea, what is our preference here. Should it be `feature_flag.flagd.<attribute>` or simply `flagd.<attribute>`

flagd/pkg/runtime/runtime.go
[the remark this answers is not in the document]
aepfli: This code was only copy pasted fro the `json.go` file, so I did not question the naming :).

But I agree with combining them within the `feature_flag.set` attribute. Maybe we should also rethink the naming, because notifications are actually the whole flags, and not just the changed ones. So it might be worth to just say `flag_count`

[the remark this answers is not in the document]
aepfli: I tried to add `version` but it is not part of the `DataSync payload`, parsing the json string would be too much effort, and changing all the syncs to add this information is - IMHO - out of scope. This seems to be a good follow-up, wdyt?

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -164,8 +164,13 @@ func (r *Multiplexer) SourcesAsMetadata() string {
 	return strings.Join(r.sources, ",")
 }
 
+type flagConfig struct {
```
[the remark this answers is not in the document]
aepfli: removed thank you

flagd/pkg/runtime/runtime.go
aepfli: adapted based on my perception, please check if you agree with this format

PR #1608, opened 2025-03-26
fix: incorrect comparison used for time

wilson-matthew:
## This PR

- Fixes a bug in the blob sync type that causes the configuration to be updated every sync.

### Notes

- The [time package recommends](https://pkg.go.dev/time#Time:~:text=In%20general%2C%20prefer%20t.Equal(u)%20to%20t%20%3D%3D%20u%2C%20since%20t.Equal%20uses%20the%20most%20accurate%20comparison%20available%20and%20correctly%20handles%20the%20case%20when%20only%20one%20of%20its%20arguments%20has%20a%20monotonic%20clock%20reading.) using .Equal(u) over t == u for comparing two time.Time types
- This bug is not present on all platforms - running flagd locally on a Macbook works as intended, but this bug is present when running in an arm64 container

### How to test

- Build a Docker image from this PR, run flagd in debug mode with a blob sync - it will no longer produce a "configuration updated" log every time it syncs


aepfli (approved):
thank you

beeme1mr (approved):
Thanks @wilson-matthew for adding a link to the docs. I learned something new today!

toddbaert:
Nice find!

wilson-matthew:
Thanks for the quick review and merge 🙂

PR #1626, opened 2025-04-27
doc: add documentation on running flagd within docker-compose

kylejuliandev:
## This PR

- Adds a section into Getting Started for using Flagd with docker-compose

### Notes

On my desktop I prefer to use docker-compose files so I don't need to remember docker run commands. Maybe this would be useful for others?

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -62,7 +62,7 @@ services:
       - '8013:8013'
 ```
 
-Create a folder called `flags` where the JSON flag and `docker-compose` files can reside. [Download the flag definition](#download-the-flag-definition) and move this JSON file to the flags folder.
+Create a folder called `flags` where the JSON flag files can reside. [Download the flag definition](#download-the-flag-definition) and move this JSON file to the flags folder.
 
 ```
 ├── flags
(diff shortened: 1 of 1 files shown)


docs/quick-start.md
```diff
@@ -43,6 +43,39 @@ docker run \
...
+#### Running flagd in Docker Compose
+
+Create a docker-compose.yaml file with the following contents:
+
+```yaml
+services:
+  flagd:
+    image: ghcr.io/open-feature/flagd:latest
+    volumes:
+      - ./flags:/etc/flagd
```
aepfli, line 55: [nit] we should evaluate if we want to use `/etc/flagd` as the base path for storing feature flags. currently the path is also the selector, so if I want to select a subset of flags - when providing multiple ones - I also need to prefix my selectors in code with `/etc/flagd` maybe just `/flags` would be sufficient (not saying that this is ideal, but we do have a ticket to improve the selector handling)?

aepfli (approved):
thank you. I think adding this information, on how to use docker-compose is for sure helpful. I wrote a little nit-picking comment. But this is definitely not a show stopper for me. just something we should keep in mind.

docs/quick-start.md
kylejuliandev: I've adjusted the mount now to be /flags instead

beeme1mr (approved):
Love it, thanks!

PR #1637, opened 2025-06-05
docs: add flag config adr

toddbaert:
Another "retrospective" ADR, like the cucumber one.

This explains our JSONLogic choice, shared `$evaluators`, as well as the the semantics of flagd's use of OpenFeature's [resolution reasons](https://openfeature.dev/specification/types#resolution-reason).

### DIFF
--- a/docs/architecture-decisions/flag-configuration.md
+++ b/docs/architecture-decisions/flag-configuration.md
@@ -29,6 +29,7 @@ flagd chose **JSON Logic** as its core evaluation engine, implementing a modifie
 This provides a secure, portable foundation where rules are expressed as JSON objects with operators as keys and parameters as values.
 
 #### Benefits realized
+
 - Rules can be stored in databases, transmitted over networks, shared between frontend/backend, and embedded in Kubernetes custom resources
 - No eval() or code injection risks - computations are deterministic and sand-boxed
 - Implementations exist in most languages
@@ -38,6 +39,7 @@ This provides a secure, portable foundation where rules are expressed as JSON ob
 The system provides two tiers of operators:
 
 ##### Primitive JSON Logic Operators (inherited from the JSONLogic)
+
 - Logical: `and`, `or`, `!`, `!!`
 - Comparison: `==`, `!=`, `>`, `<`, etc
 - Arithmetic: `+`, `-`, `*`, `/`, `%`
@@ -47,6 +49,7 @@ The system provides two tiers of operators:
 - Assignment and extraction: `var`
 
 ##### Custom flagd Extensions
+
 - `fractional`: Deterministic percentage-based distribution using murmur3 hashing
 - `starts_with`/`ends_with`: String prefix/suffix matching for common patterns
 - `sem_ver`: Semantic version comparisons with standard (npm-style) operators
@@ -57,6 +60,7 @@ The system provides two tiers of operators:
 flagd automatically injects critical context values:
 
 ##### System-provided context
+
 - `$flagd.flagKey`: The flag being evaluated (available v0.6.4+)
 - `$flagd.timestamp`: Unix timestamp of evaluation (available v0.6.7+)
 
@@ -73,6 +77,7 @@ flagd returns specific reason codes with every evaluation to indicate how the de
 5. **ERROR**: Evaluation failed due to invalid configuration
 
 This transparency enables:
+
 - Appropriate caching strategies (only STATIC flags are cached)
 - Improved debugging, telemetry, and monitoring of flag behavior
 
@@ -106,6 +111,7 @@ The `$evaluators` top-level property enables shared targeting logic:
 ##### Intelligent Caching Strategy
 
 Only flags with reason **STATIC** are cached, as they have deterministic outputs. This ensures:
+
 - Maximum cache efficiency for simple toggles
 - Fresh evaluation for complex targeting rules
 - Cache invalidation on configuration changes
@@ -113,10 +119,12 @@ Only flags with reason **STATIC** are cached, as they have deterministic outputs
 ##### Schema-Driven Configuration
 
 Two schemas validate flag configurations:
+
 - `https://flagd.dev/schema/v0/flags.json`: Overall flag structure
 - `https://flagd.dev/schema/v0/targeting.json`: Targeting rule validation
 
 These enable:
+
 - IDE support with autocomplete
 - Run-time and build-time validation
 - Separate validation of rules and overall configuration if desired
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/flag-configuration.md
```diff
@@ -0,0 +1,155 @@
...
+
+Feature flag systems require a flexible, safe, and portable way to express targeting rules that can evaluate contextual data to determine which variant of a feature to serve.
+
+flagd's targeting system was designed with several key requirements:
+
+## Requirements
+
+- **Language agnostic**: Rules must be portable across different programming languages, ideally relying on existing expression language(s)
+- **Safe evaluation**: No arbitrary code execution or system access
+- **Deterministic**: Same inputs must always produce same outputs
```
tangenti, line 20: $flagd.timestamp seems to break this requirement?

docs/architecture-decisions/flag-configuration.md
```diff
@@ -0,0 +1,155 @@
...
+
+### JSON Logic as the Foundation
+
+flagd chose **JSON Logic** as its core evaluation engine, implementing a modified version with custom extensions.
+This provides a secure, portable foundation where rules are expressed as JSON objects with operators as keys and parameters as values.
+
+#### Benefits realized
+
+- Rules can be stored in databases, transmitted over networks, shared between frontend/backend, and embedded in Kubernetes custom resources
+- No eval() or code injection risks - computations are deterministic and sand-boxed
```
tangenti, line 34: I think this depends on the implementation of the eval engine, and less about the language itself.
For example, a JavaScript JSON logic engine had a CVE that allows command injection: https://nvd.nist.gov/vuln/detail/CVE-2021-4329

docs/architecture-decisions/flag-configuration.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Flag and Targeting Configuration
+
+## Background
+
+Feature flag systems require a flexible, safe, and portable way to express targeting rules that can evaluate contextual data to determine which variant of a feature to serve.
+
+flagd's targeting system was designed with several key requirements:
+
+## Requirements
```
tangenti, line 16: What was the requirement for the typing system back then?  Is weakly typed a requirement or a result?

docs/architecture-decisions/flag-configuration.md
```diff
@@ -0,0 +1,155 @@
...
+These enable:
+
+- IDE support with autocomplete
+- Run-time and build-time validation
+- Separate validation of rules and overall configuration if desired
+
+## Considered Options
+
+- **Custom DSL**: Would require parsers in every language
+- **JavaScript/Lua evaluation**: Security risks and language lock-in
```
tangenti, line 135: What was the biggest concern for Lua? 
Lua engine normally comes with a sandboxed VM, which is more lightweight and secure IMO.

docs/architecture-decisions/flag-configuration.md
```diff
@@ -0,0 +1,155 @@
...
+- **CEL**: limited number of implementations at time of decision, can't be directly parsed/validated when embedded in Kubernetes resources
+
+## Consequences
+
+### Positive
+
+- Good, because implementations exist across languages
```
tangenti, line 142: True, but also there's not much consistency guarantee across languages.

docs/architecture-decisions/flag-configuration.md
tangenti, line 16: Mostly due to the JS typing system, JSONLogic's behavior on `undefined` values (context not provided, for example) can be very confusing. Was this considered and evaluated?

docs/architecture-decisions/flag-configuration.md
toddbaert: I guess it depends on whether or not you consider the timestamp an sort of implicit input... We could change it to "predictable" though.

tangenti: To me, what users provide through the evaluation API are the input, and timestamp is not there.

docs/architecture-decisions/flag-configuration.md
toddbaert: Ya, that's fair. At least at the time, we saw JSONLogic as _less likely_ to have these sort of issues.

docs/architecture-decisions/flag-configuration.md
toddbaert: Definitely a result.

docs/architecture-decisions/flag-configuration.md
toddbaert: Lua was something I briefly advocated for, I've had some experience with it doing nginx customization. I remember a community member being opposed to it on the basis of security concerns, and at the time it seemed like a higher bar.

docs/architecture-decisions/flag-configuration.md
toddbaert: Well, there's enough at least that our test suite works in Go, Python, Java, JS, and .NET. It covers every operation and a fair few edge cases: https://github.com/open-feature/flagd-testbed/tree/main/gherkin

I acknowledge there's inconsistencies, particularly around `undefined`s. It's one of the reasons the [JSONLogic](https://github.com/json-logic/) org exists (though admittedly it's been hard to make progress with the few volunteers there). I'm sure we could use your help if you're interested in contributing to that.

docs/architecture-decisions/flag-configuration.md
toddbaert: I can't recall if that particular behavior was discussed.

toddbaert:
@tangenti Thanks for the feedback.

I'd like to make it clear that personally, though I think our current JSONLogic implementation(s) are "good enough", I think there's ways to improve, and alternatives. To name a few:

- Leverage WASM and use a single JSON logic implementation, distributed as a module, for maximum implementation coherence
- Work on improving JSONLogic via https://github.com/json-logic/
- Support additional targeting DSLs (Lua, CEL, others), either via WASM or natively.

I would be open to any of the above, and lend my support, if there's engineering effort to spare for these.

docs/architecture-decisions/flag-configuration.md
toddbaert: OK I'll change that to "predictable".

docs/architecture-decisions/flag-configuration.md
toddbaert: My first alternative in [this comment](https://github.com/open-feature/flagd/pull/1637#issuecomment-2949513068) might be a reasonable solution to some implementation inconsistencies. Interested in your take on that; we could do it progressively and it would be non-breaking.

docs/architecture-decisions/flag-configuration.md
toddbaert: Actually I think I'll keep `Deterministic` (because I believe it still is), but mention the timestamp exception situation.

toddbaert:
Since this is historical, I'm merging this.

That doesn't mean it can't be edited or superseded, it's simply an explanation of a previous decision and reflects the current implementation.

Issue #1582, opened 2025-03-06
[FEATURE] Serverside Deadline


toddbaert:
### Requirements

We should implement a server-side deadline for flagd streams ([sync](https://github.com/open-feature/flagd/blob/5c2c64f878b8603dd37cbfd79b0e1588e4b5a3c6/flagd/pkg/service/flag-sync/handler.go#L20) and [event](https://github.com/open-feature/flagd/blob/5c2c64f878b8603dd37cbfd79b0e1588e4b5a3c6/flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go#L137)) which after a timeout, sends a gRPC error code 4, ([DEADLINE_EXCEEDED](https://grpc.io/docs/guides/status-codes/)). Server-side deadlines aren't very common, but they will be useful in our case for forcing clients (providers) to restart their streams to keep them alive in some environments where PING frames don't work, as well as for testing purposes.

Requirements:

- Stream deadlines can be configured from command line (one for both streams is enough)
- Attached to existing [FLAGD_DEADLINE_MS](https://github.com/open-feature/flagd/blob/5c2c64f878b8603dd37cbfd79b0e1588e4b5a3c6/docs/reference/specifications/providers.md?plain=1#L274) env var
- Associated tests

alexandraoberaigner:
Hi @toddbaert, I'll give this one a shot. I'd only have one question: what is meant by 'attaching to existing env var'? **EDIT**: it means that the deadline should also be settable by this env var

aepfli:
as https://github.com/open-feature/flagd/pull/1638 is merged, i think we can close this. If I have been wrong, please just reopen

PR #1638, opened 2025-06-06
feat: add server-side deadline to sync service

alexandraoberaigner:
## This PR

- adds server side deadline for sync and event streams configurable via cmd argument `--stream-deadline`

### Related Issues

#1582

### How to test

1. Run flagd with `--stream-deadline 3s` // 3s can be replaced with any duration the deadline should have
2. Test Event Stream deadline: run `grpcurl -v --proto schemas/protobuf/flagd/evaluation/v1/evaluation.proto -plaintext localhost:8013 flagd.evaluation.v1.Service/EventStream` or similar depending on your flagd settings to check if the deadline exceeded is returned after the specified duration
3. Test Sync Service Stream deadline: run `grpcurl -v --proto schemas/protobuf/flagd/sync/v1/sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags` or similar depending on your flagd settings to check if the deadline exceeded is returned after the specified duration

### DIFF
--- a/docs/reference/flagd-cli/flagd_start.md
+++ b/docs/reference/flagd-cli/flagd_start.md
@@ -29,6 +29,7 @@ flagd start [flags]
   -k, --server-key-path string               Server side tls key path
   -d, --socket-path string                   Flagd unix socket path. With grpc the evaluations service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
   -s, --sources string                       JSON representation of an array of SourceConfig objects. This object contains 2 required fields, uri (string) and provider (string). Documentation for this object: https://flagd.dev/reference/sync-configuration/#source-configuration
+      --stream-deadline duration             Set a server-side deadline for flagd sync and event streams (default 0, means no deadline).
   -g, --sync-port int32                      gRPC Sync port (default 8015)
   -e, --sync-socket-path string              Flagd sync service socket path. With grpc the sync service will be available on this address.
   -f, --uri .yaml/.yml/.json                 Set a sync provider uri to read data from, this can be a filepath, URL (HTTP and gRPC), FeatureFlag custom resource, or GCS or Azure Blob. When flag keys are duplicated across multiple providers the merge priority follows the index of the flag arguments, as such flags from the uri at index 0 take the lowest precedence, with duplicated keys being overwritten by those from the uri at index 1. Please note that if you are using filepath, flagd only supports files with .yaml/.yml/.json extension.
--- a/flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
+++ b/flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
@@ -104,7 +104,7 @@ func TestConnectServiceV2_ResolveAll(t *testing.T) {
 			).AnyTimes()
 
 			metrics, exp := getMetricReader()
-			s := NewFlagEvaluationService(logger.NewLogger(nil, false), eval, &eventingConfiguration{}, metrics, nil, nil)
+			s := NewFlagEvaluationService(logger.NewLogger(nil, false), eval, &eventingConfiguration{}, metrics, nil, nil, 0)
 
 			// when
 			got, err := s.ResolveAll(context.Background(), connect.NewRequest(tt.req))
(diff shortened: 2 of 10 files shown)


toddbaert:
@alexandraoberaigner `make generate-docs` and committing the diff should fix your CI issue; it will just add docs about your option (you may have already added them, but either have a slight text mismatch from the auto-generated or you're missing something)

alexandraoberaigner:
> @alexandraoberaigner `make generate-docs` and committing the diff should fix your CI issue; it will just add docs about your option (you may have already added them, but either have a slight text mismatch from the auto-generated or you're missing something)

Thanks @toddbaert for the hint! Docs should be fixed ✅ 

I've implemented both deadlines now, but wrote e2e tests only for the sync service; I'm not sure how to properly test the EventStream() function in the EvaluatorService. Can you point me to the proper location where to write the test? I couldn't find an existing e2e test & hesitated to mock the serverstream (just unit testing the function). See also: https://github.com/open-feature/flagd/pull/1638#discussion_r2133675842

flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
```diff
@@ -1008,3 +1018,61 @@ func Test_mergeContexts(t *testing.T) {
 		})
 	}
 }
+
+/*
+func TestFlagEvaluationService_EventStream(t *testing.T) {
```
alexandraoberaigner: - [x] remove or fix this test: server stream would need to be mocked - is there a better way to test the eventstream function?

docs/reference/flagd-cli/flagd_start.md
```diff
@@ -28,6 +28,7 @@ flagd start [flags]
   -k, --server-key-path string          Server side tls key path
   -d, --socket-path string              Flagd unix socket path. With grpc the evaluations service will become available on this address. With http(s) the grpc-gateway proxy will use this address internally.
   -s, --sources string                  JSON representation of an array of SourceConfig objects. This object contains 2 required fields, uri (string) and provider (string). Documentation for this object: https://flagd.dev/reference/sync-configuration/#source-configuration
+      --stream-deadline duration        Set a server-side deadline for flagd sync and event streams (default 0, means no deadline).
```
aepfli: [issue] The cloud native approach is to allow setting those settings via an environment variable. Hence, this env var will also apply to the providers, as this variable will be set for the whole deployment (most likely). For the providers, this is already specified [here](https://flagd.dev/reference/specifications/providers/#configuration-options). Hence, we should maybe stick to an integer without a time unit, else we have a breaking behavior between providers and flagd

toddbaert: I thought this too, but then I was looking and I don't think we actually read any ENV vars for configuring flagd itself - we just use them to configure the providers it seems (or I can't find the code that reads the env vars).

Given this, I think using a duration is better because is more flexible and consistent with our other flagd options with use durations

flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
toddbaert: I don't see a simple way to do this because we don't have the interfaces we need (the 3rd party abstractions are concrete unfortunately). I'm fine with removing this test unless somebody has a suggestion for clearly mocking this.

toddbaert (approved):
Tested and looks good, works well. I don't think it's worth some crazy mocking with the connect code to add the last test you mentioned... if someone has a quick solution to that one it's fine, but not a blocker IMO.

docs/reference/flagd-cli/flagd_start.md
aepfli: thinking about his for a time now, and your input opened some eyes, for me - where the env var is still usable, if needed and this is also a non issue for me now ;)

flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
alexandraoberaigner: Okay, I'll leave this open for a bit then to see if someone else has an idea. Otherwise removing it as you suggested 👍

docs/reference/flagd-cli/flagd_start.md
alexandraoberaigner: Forgot to mention this as comment on the ticket. I wanted to implement also an env var setting and found no code for reading env vars - which is why its just the start up flag now :) I'll close this conv, feel free to reopen if necessary

aepfli (approved):

toddbaert:
I'll fix the conflicts here and merge.

PR #1639, opened 2025-06-06
docs: add code default adr

beeme1mr:
## This PR

- ADR to add support for explicitly deferring to the code default.

### Notes

This has been a challenge for users because enabling a feature for targeting requires a default variant to be used, which may not match the default used in code.

### DIFF
--- a/docs/architecture-decisions/support-code-default.md
+++ b/docs/architecture-decisions/support-code-default.md
@@ -1,0 +1,333 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: @beeme1mr
+created: 2025-05-06
+updated: 2025-05-06
+---
+
+# Support Explicit Code Default Values in flagd Configuration
+
+This ADR proposes adding support for explicitly configuring flagd to use code-defined default values by allowing `null` as a valid default variant. This change addresses the current limitation where users cannot differentiate between "use the code's default" and "use this configured default" without resorting to workarounds like misconfigured rulesets.
+
+## Background
+
+Currently, flagd requires a default variant to be specified in flag configurations. This creates a fundamental mismatch with the OpenFeature specification and common feature flag usage patterns where code-defined defaults serve as the ultimate fallback.
+
+The current behavior leads to confusion and operational challenges:
+
+1. **Two Sources of Truth**: Applications have default values defined in code (as per OpenFeature best practices), while flagd configurations require their own default variants. This dual-default pattern violates the principle of single source of truth.
+
+2. **State Transition Issues**: When transitioning a flag from DISABLED to ENABLED state, the behavior changes unexpectedly:
+   - DISABLED state: Flag evaluation falls through to code defaults
+   - ENABLED state: Flag evaluation uses the configured default variant
+
+3. **Workarounds**: Users resort to misconfiguring rulesets (e.g., returning invalid variants) to force fallback to code defaults, which generates confusing error states and complicates debugging.
+
+4. **OpenFeature Alignment**: The OpenFeature specification emphasizes that code defaults should be the ultimate fallback, but flagd's current design doesn't provide a clean way to express this intent.
+
+Related discussions and context can be found in the [OpenFeature specification](https://openfeature.dev/specification/types) and [flagd flag definitions reference](https://flagd.dev/reference/flag-definitions/).
+
+## Requirements
+
+- **Explicit Code Default Support**: Users must be able to explicitly configure a flag to use the code-defined default value as its resolution
+- **Backward Compatibility**: Existing flag configurations must continue to work without modification
+- **Clear Semantics**: The configuration must clearly indicate when code defaults are being used versus configured defaults
+- **Appropriate Reason Codes**: Resolution details must include appropriate reason codes when code defaults are used (e.g., `DEFAULT` or a new specific reason)
+- **Schema Validation**: JSON schema must support and validate the new configuration options
+- **Provider Compatibility**: All OpenFeature providers must handle the new behavior correctly
+- **Testbed Coverage**: flagd testbed must include test cases for the new functionality
+
+## Considered Options
+
+- **Option 1: Allow `null` as Default Variant** - Modify the schema to accept `null` as a valid value for defaultVariant, signaling "use code default"
+- **Option 2: Make Default Variant Optional** - Remove the requirement for defaultVariant entirely, with absence meaning "use code default"
+- **Option 3: Special Variant Value** - Define a reserved variant name (e.g., `"__CODE_DEFAULT__"`) that signals code default usage
+- **Option 4: New Configuration Property** - Add a new property like `useCodeDefault: true` alongside or instead of defaultVariant
+- **Option 5: Status Quo with Documentation** - Keep current behavior but improve documentation about workarounds
+
+## Proposal
+
+We propose implementing **Option 1: Allow `null` as Default Variant**, potentially combined with **Option 2: Make Default Variant Optional** for maximum flexibility.
+
+The implementation leverages field presence in evaluation responses across all protocols (in-process, RPC, and OFREP). When a flag configuration has `defaultVariant: null`, the evaluation response omits the value field entirely, which serves as a programmatic signal to the client to use its code-defined default value.
+
+This approach offers several key advantages:
+
+1. **No Protocol Changes**: RPC and OFREP protocols remain unchanged
+2. **Clear Semantics**: Omitted value field = "use your code default"
+3. **Backward Compatible**: Existing clients and servers continue to work
+4. **Universal Pattern**: Works consistently across all evaluation modes
+
+The absence of a value field provides an unambiguous signal that distinguishes between "the server evaluated to null/false/empty" (value field present) and "the server delegates to your code default" (value field absent).
+
+### Implementation Details
+
+1. **Schema Changes**:
+
+   ```json
+   {
+     "defaultVariant": {
+       "oneOf": [
+         { "type": "string" },
+         { "type": "null" }
+       ],
+       "description": "Default variant to use. Set to null to use code-defined default."
+     }
+   }
+   ```
+
+2. **Protobuf Considerations**:
+
+   **No proto changes required** - The existing proto3 definitions already support this behavior:
+
+   ```protobuf
+   message ResolveBooleanResponse {
+     bool value = 1;
+     string reason = 2;
+     string variant = 3;
+     google.protobuf.Struct metadata = 4;
+   }
+   ```
+
+   In proto3, all fields are optional by default. To implement code default delegation:
+   - Do NOT set the `value` field (not even to false/0/"")
+   - Do NOT set the `variant` field
+   - Set `reason` to "DEFAULT"
+   - Optionally set `metadata`
+
+   **Critical implementation notes**:
+   - Ensure flagd omits fields rather than setting zero values
+   - Different languages detect field presence differently:
+     - Go: Check for zero values or use pointers
+     - Java: Use `hasValue()` / `hasVariant()` methods
+     - Python: Use `HasField()` or check field presence
+     - JavaScript: Check for `undefined` (not null)
+   - Consider adding proto comments documenting this behavior
+   - Test wire format to confirm fields are actually omitted
+
+   **Example - Correct vs Incorrect Implementation**:
+
+   ```go
+   // INCORRECT - sets zero value
+   response := &ResolveBooleanResponse{
+       Value:   false,  // This sends 'false' on the wire
+       Reason:  "DEFAULT",
+       Variant: "",     // This sends empty string on the wire
+   }
+   
+   // CORRECT - omits fields
+   response := &ResolveBooleanResponse{
+       Reason: "DEFAULT",
+       // Value and Variant fields are not set at all
+   }
+   ```
+
+3. **Evaluation Behavior**:
+   - When flag has `defaultVariant: null` and targeting returns no match
+   - Server responds with both value and variant fields omitted, reason set to "DEFAULT"
+   - Client detects the missing value field and uses its code-defined default
+   - This same pattern works across all evaluation modes
+
+   **In-Process Mode Special Considerations**:
+   - The in-process evaluator must return the same "shape" of response
+   - Cannot return null/nil as that's different from "no value"
+   - May need a special response type or wrapper to indicate delegation
+   - Example approach: Return evaluation result with a "useDefault" flag
+
+4. **Remote Evaluation Protocol Responses**:
+
+   **OFREP (OpenFeature Remote Evaluation Protocol)**:
+   - Return HTTP 200 with response body that omits both value and variant fields
+   - Reason field set to "DEFAULT" to indicate delegation
+   - Clear distinction from error cases (which return 4xx/5xx)
+
+   **flagd RPC**:
+   - Omit both the type-specific value field and variant field in response messages
+   - Use protobuf field presence to signal "no opinion from server"
+   - No changes needed to RPC method signatures
+
+5. **Provider Implementation**:
+   - Check for presence/absence of value field in responses
+   - When value is absent and reason is "DEFAULT", use code-defined default
+   - When value is present (even if null/false/empty), use that value
+   - Variant field will also be absent in delegation responses, resulting in undefined variant in resolution details
+   - Responses with value fields work as before, maintaining backward compatibility
+
+### Design Rationale
+
+**Using "DEFAULT" reason**: We intentionally reuse the existing "DEFAULT" reason code rather than introducing a new one (like "CODE_DEFAULT"). The distinction between a configuration default and code default is clear from the response structure:
+
+- Configuration default: Has both value and variant fields
+- Code default: Omits both value and variant fields
+
+**Field Omission Pattern**: Using field presence/absence is a well-established pattern in protocol design:
+
+- Unambiguous: Cannot confuse "null value" with "no opinion"
+- Language agnostic: Works across type systems
+- Protocol friendly: Natural in both JSON and Protobuf
+- Backward compatible: Existing responses always include values
+- Spec compliant: OpenFeature allows undefined variants
+
+The omission of both value and variant fields creates a clear, consistent signal that the server is fully delegating the decision to the client's code default.
+
+This approach maintains compatibility with the established OpenFeature terminology while providing clear semantics through response structure.
+
+### API changes
+
+**Flag Configuration**:
+
+```yaml
+flags:
+  my-feature:
+    state: ENABLED
+    defaultVariant: null  # Explicitly use code default
+    variants:
+      on: true
+      off: false
+    targeting:
+      if:
+        - "===":
+            - var: user-type
+            - "beta"
+        - on
+```
+
+**OFREP Response** when code default is indicated:
+
+#### Single flag evaluation response
+
+```json
+{
+  "key": "my-feature",
+  "reason": "DEFAULT",
+  "metadata": {}
+}
+```
+
+#### Bulk flag evaluation response
+
+```json
+{
+  "flags": [
+    {
+      "key": "my-feature",
+      "reason": "DEFAULT",
+      "metadata": {}
+    }
+  ]
+}
+```
+
+Note: Both `value` and `variant` fields are intentionally omitted to signal "use code default"
+
+**flagd RPC Response** (ResolveBooleanResponse):
+
+```protobuf
+{
+  "reason": "DEFAULT",
+  "metadata": {}
+}
+```
+
+Both the type-specific value field and variant field are omitted
+
+**Provider behavior**:
+
+1. Check response for presence of value field
+2. If value field is absent and reason is "DEFAULT", use code-defined default
+3. If value field is present (even if null/false/empty), use that value
+4. The variant field will also be absent when delegating to code defaults
+5. This logic is consistent across in-process, RPC, and OFREP modes
+
+This approach clearly differentiates between:
+
+- Server returning an actual value with a variant (both fields present)
+- Server delegating to code default (both fields absent)
+
+**Example evaluation flow**:
+
+```javascript
+// Client code
+const value = client.getBooleanValue('my-feature', false, context);
+
+// Server evaluates flag with defaultVariant: null
+// No targeting match, so server returns:
+{
+  "reason": "DEFAULT",
+  "metadata": {}
+  // Note: both "value" and "variant" fields are omitted
+}
+
+// Client detects missing value field and uses its default (false)
+// Resolution details show reason: "DEFAULT" with undefined variant
+```
+
+### Consequences
+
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+
+## Proposal
+
+We propose implementing **Option 1: Allow `null` as Default Variant**, potentially combined with **Option 2: Make Default Variant Optional** for maximum flexibility.
+
+The implementation leverages field presence in evaluation responses across all protocols (in-process, RPC, and OFREP). When a flag configuration has `defaultVariant: null`, the evaluation response omits the value field entirely, which serves as a programmatic signal to the client to use its code-defined default value.
+
+This approach offers several key advantages:
+
+1. **No Protocol Changes**: RPC and OFREP protocols remain unchanged
```
toddbaert, line 57: This might not be true in the MOST strict sense IIUC, since I think we will mark the value as optional in the proto... but it's about as small a change as I can imagine.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+2. **Clear Semantics**: Omitted value field = "use your code default"
+3. **Backward Compatible**: Existing clients and servers continue to work
+4. **Universal Pattern**: Works consistently across all evaluation modes
+
+The absence of a value field provides an unambiguous signal that distinguishes between "the server evaluated to null/false/empty" (value field present) and "the server delegates to your code default" (value field absent).
```
toddbaert, line 62: I'm not 100% sure, but until they absorb this, a small risk here might be that old providers interpret missing `value` fields as zero values (empty string, numeric 0, etc). I don't think this is a blocker but it's something to note when implementing/releasing.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+   ```protobuf
+   message ResolveBooleanResponse {
+     bool value = 1;
+     string reason = 2;
+     string variant = 3;
+     google.protobuf.Struct metadata = 4;
+   }
+   ```
+
+   In proto3, all fields are optional by default. To implement code default delegation:
```
toddbaert, line 93: > In proto3, all fields are optional by default.

This is true, but it can cause issues like I mentioned above. That's why the `optional` proto3 keyword still exists, so clients can differentiate between intentional zero values and unset values: https://protobuf.dev/programming-guides/proto3/#field-labels.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+   - Do NOT set the `value` field (not even to false/0/"")
+   - Do NOT set the `variant` field
+   - Set `reason` to "DEFAULT"
+   - Optionally set `metadata`
+
+   **Critical implementation notes**:
+   - Ensure flagd omits fields rather than setting zero values
+   - Different languages detect field presence differently:
+     - Go: Check for zero values or use pointers
+     - Java: Use `hasValue()` / `hasVariant()` methods
```
toddbaert, line 103: > Use `hasValue()` / `hasVariant()` methods

These are only generated if you add `optional`, IIUC.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+2. **Provider Tests**: Each provider must have tests confirming they detect missing fields correctly in their language
+3. **Integration Tests**: End-to-end tests across different language combinations (e.g., Go flagd with Java provider)
+4. **OFREP Tests**: Verify JSON responses correctly omit fields (not set to null)
+5. **Backward Compatibility Tests**: Ensure old providers handle new responses gracefully
+6. **Consistency Tests**: Verify identical behavior across in-process, RPC, and OFREP modes
+
+### Open questions
+
+- How should providers handle responses with missing value fields in strongly-typed languages?
+- Should we support both `null` and absent `defaultVariant` fields, or choose one approach?
```
toddbaert, line 320: I think we should support both, I can't see a disadvantage.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+1. **Wire Format Tests**: Verify that protobuf messages with omitted fields are correctly serialized without the fields (not with zero values)
+2. **Provider Tests**: Each provider must have tests confirming they detect missing fields correctly in their language
+3. **Integration Tests**: End-to-end tests across different language combinations (e.g., Go flagd with Java provider)
+4. **OFREP Tests**: Verify JSON responses correctly omit fields (not set to null)
+5. **Backward Compatibility Tests**: Ensure old providers handle new responses gracefully
+6. **Consistency Tests**: Verify identical behavior across in-process, RPC, and OFREP modes
+
+### Open questions
+
+- How should providers handle responses with missing value fields in strongly-typed languages?
```
toddbaert, line 318: I believe all languages have some equivalent to "hasValue" etc, if the `optional` keyword is used.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+- Should we support both `null` and absent `defaultVariant` fields, or choose one approach?
+- What migration path should we recommend for users currently using workarounds?
+- Should this feature be gated behind a configuration flag during initial rollout?
+- How do we ensure consistent behavior across all provider implementations?
+- Should providers validate that the reason is "DEFAULT" when value is omitted, or accept any omitted value as delegation?
+- How do we handle edge cases where network protocols might strip empty fields?
+- When the client uses its code default after receiving a delegation response, what variant should be reported in telemetry/analytics?
```
toddbaert, line 332: I would vote none, just as we currently do for errors. We could still easily distinguish from errors via `DEFAULT`.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+6. **Consistency Tests**: Verify identical behavior across in-process, RPC, and OFREP modes
+
+### Open questions
+
+- How should providers handle responses with missing value fields in strongly-typed languages?
+- Should we support both `null` and absent `defaultVariant` fields, or choose one approach?
+- What migration path should we recommend for users currently using workarounds?
+- Should this feature be gated behind a configuration flag during initial rollout?
+- How do we ensure consistent behavior across all provider implementations?
+- Should providers validate that the reason is "DEFAULT" when value is omitted, or accept any omitted value as delegation?
```
toddbaert, line 328: I think providers should ensure that any resolution that has no value, `REASON=DEFAULT`, likely by just setting it themselves.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+  "metadata": {}
+}
+```
+
+Both the type-specific value field and variant field are omitted
+
+**Provider behavior**:
+
+1. Check response for presence of value field
+2. If value field is absent and reason is "DEFAULT", use code-defined default
```
toddbaert:
```suggestion
2. If value field is absent and error code is absent , use code-defined default
```

I _think_ this is the heuristic we want. Regardless of the DEFAULT reason. The provider can set DEFAULT itself in this case in the resolution, I think. :thinking:

docs/architecture-decisions/support-code-default.md
askpt: I am with @toddbaert here. Imagine the situation where the flag value is `null`. If that is the `signal` to use the default value from code instead, it might lead to the code using the default value in code, where the user intended the flag value to be `null`.
I like option 2 and 3's approach better, with the goal of completely deprecating and removing the default value in future versions.
Proposed:
```mermaid
sequenceDiagram
    Code->>OpenFeature: Evaluate("key", "defaultValue")
    OpenFeature->>flagd: Evaluate ("key")
    flagd-->>OpenFeature: return `null`
    OpenFeature-->>Code: return `defaultValue`
```
I would prefer:
```mermaid
sequenceDiagram
    Code->>OpenFeature: Evaluate("key", "defaultValue")
    OpenFeature->>flagd: Evaluate ("key")
    flagd-->>OpenFeature: return `null`
    OpenFeature-->>Code: return `null`
```
If we use default code:
```mermaid
sequenceDiagram
    Code->>OpenFeature: Evaluate("key", "defaultValue")
    OpenFeature->>flagd: Evaluate ("key")
    flagd-->>OpenFeature: return `__CODE_DEFAULT__`
    OpenFeature-->>Code: return `defaultValue`
```

guidobrei: @askpt, the `defaultValue` in the open feature API refers to the value, whereas the `null` and `__CODE_DEFAULT__` refer to flagd variant names.
I would prefer not to encode any special hints in the value, as we have an evaluation reason.

docs/architecture-decisions/support-code-default.md
guidobrei: We must have a concept already for non-existing flags that return a "FLAG_NOT_FOUND" evaluation response in all languages.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+4. **OFREP Tests**: Verify JSON responses correctly omit fields (not set to null)
+5. **Backward Compatibility Tests**: Ensure old providers handle new responses gracefully
+6. **Consistency Tests**: Verify identical behavior across in-process, RPC, and OFREP modes
+
+### Open questions
+
+- How should providers handle responses with missing value fields in strongly-typed languages?
+- Should we support both `null` and absent `defaultVariant` fields, or choose one approach?
+- What migration path should we recommend for users currently using workarounds?
+- Should this feature be gated behind a configuration flag during initial rollout?
```
guidobrei, line 324: Would go with documentation only and clearly state the required provider versions.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+- How do we ensure consistent behavior across all provider implementations?
```
guidobrei, line 326: Should be doable with our Gherkin test suite.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+To ensure correct implementation across all components:
+
+1. **Wire Format Tests**: Verify that protobuf messages with omitted fields are correctly serialized without the fields (not with zero values)
+2. **Provider Tests**: Each provider must have tests confirming they detect missing fields correctly in their language
+3. **Integration Tests**: End-to-end tests across different language combinations (e.g., Go flagd with Java provider)
+4. **OFREP Tests**: Verify JSON responses correctly omit fields (not set to null)
+5. **Backward Compatibility Tests**: Ensure old providers handle new responses gracefully
+6. **Consistency Tests**: Verify identical behavior across in-process, RPC, and OFREP modes
+
+### Open questions
```
guidobrei, line 316: How does this integrate with OFREP?

Afaik, this would require an `evaluationFailure`. Every successful evaluation would require a `value`. See [Open API spec](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml#L195)

guidobrei, line 316: Are there any side-effects on the [CLI](https://github.com/open-feature/cli) we should consider?

docs/architecture-decisions/support-code-default.md
guidobrei: Afaik, for OFREP this would require an `evaluationFailure`. Every successful evaluation would require a `value`. See [Open API spec](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml#L195)

docs/architecture-decisions/support-code-default.md
askpt: Thanks @guidobrei. Got confused with the `value` and `variant`.

beeme1mr: @guidobrei summed it up perfectly. Basically, flagd can't return a variant because there isn't one to return. I believe omitting this value is better than artificially creating one.

docs/architecture-decisions/support-code-default.md
beeme1mr: Not that I can think of.

docs/architecture-decisions/support-code-default.md
beeme1mr: Agreed. We can simply update the docs in the last step.

docs/architecture-decisions/support-code-default.md
beeme1mr: We may need to update OFREP to handle this more cleanly. I do not believe this should be treated like an evaluation error.

guidobrei: > I do not believe this should be treated like an evaluation error.

Totally agree. Just wanted to point out that we have code paths without a value already.

guidobrei (approved):

askpt (approved):

docs/architecture-decisions/support-code-default.md
toddbaert: @lukas-reining @thomaspoignant @jonathannorris I've @'d you here because we think we need to support empty/null/nil/undefined flag values in flagd responses, and therefor in OFREP. When flagd providers get these, it will interpret it as "use whatever the supplied code default is, and don't consider it an error". We don't have this functionality in flagd yet, but we have seen that it's needed both externally and internally at Dynatrace.

Do you think we can support this in OFREP?

jonathannorris: I'll check later in more detail, but I think we assumed that 404 responses from OFREP would be considered "flag not found".

docs/architecture-decisions/support-code-default.md
erenatas: In its absence, would it be a correct approach that defaultVariant would be assumed `null`?

erenatas:
Rust provider wise, this should not be a big impact, we already resolve flags as `Result`, I assume `null` would return as Result's `None`. Would appreciate if you create a follow up ticket if this is merged :)

docs/architecture-decisions/support-code-default.md
thomaspoignant: Yes this is how we manage that, 404 = flag not found.
Do we need more than that?

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+
+Currently, flagd requires a default variant to be specified in flag configurations. This creates a fundamental mismatch with the OpenFeature specification and common feature flag usage patterns where code-defined defaults serve as the ultimate fallback.
+
+The current behavior leads to confusion and operational challenges:
+
+1. **Two Sources of Truth**: Applications have default values defined in code (as per OpenFeature best practices), while flagd configurations require their own default variants. This dual-default pattern violates the principle of single source of truth.
+
+2. **State Transition Issues**: When transitioning a flag from DISABLED to ENABLED state, the behavior changes unexpectedly:
+   - DISABLED state: Flag evaluation falls through to code defaults
+   - ENABLED state: Flag evaluation uses the configured default variant
```
tangenti, line 23: Except for the case of no targeting rule, is there any other case where the configured default will be returned with NO error?

beeme1mr: Yes, this would also occur any time a targeting rule returns an invalid variant, null, or no variant at all.

In this example, emails not using the `@ingen.com` domain would use the default variant:

```json
{
  "flags": {
    "enable-mainframe-access": {
      "state": "ENABLED",
      "variants": {
        "true": true,
        "false": false
      },
      "targeting": {
        "if": [
          {
            "ends_with": [
              {
                "var": "email"
              },
              "@ingen.com"
            ]
          },
          "true"
        ]
      },
      "defaultVariant": "false"
    }
  }
}
```

Our problem is that a new flag starts out at "DISABLED" in our management system. That works well because it always uses the value defined in code. Then, when a user wants to enable the flag for a subset of users, the management tool controls the default variant (and value). This is more risky because there's no guarantee that the default variant matches what the user set in code.

docs/architecture-decisions/support-code-default.md
beeme1mr: That will work for supporting backwards compatibility but I think we should consider extending OFREP to handle this scenario a bit cleaner. I'll open a PR for that if/when this ADR is approved.

toddbaert:
@tangenti @guidobrei @cupofcat @alexandraoberaigner @chrfwow @dominikhaska 

As mentioned in our meeting, I've created some demos that try to explain some of the nuances related to this proposal in our current targeting semantics. Simply open the links below and click "Evalaute" for a demo:


First, in [this example](https://flagd.dev/playground/?flags=%7B%22%24schema%22%3A%22https%3A%2F%2Fflagd.dev%2Fschema%2Fv0%2Fflags.json%22%2C%22flags%22%3A%7B%22supports-one-hour-delivery%22%3A%7B%22state%22%3A%22ENABLED%22%2C%22defaultVariant%22%3A%22off%22%2C%22variants%22%3A%7B%22on%22%3Atrue%2C%22off%22%3Afalse%7D%2C%22targeting%22%3A%7B%22if%22%3A%5B%7B%22in%22%3A%5B%7B%22var%22%3A%22locale%22%7D%2C%5B%22us%22%2C%22ca%22%5D%5D%7D%2C%22on%22%2C%22off%22%5D%7D%7D%7D%7D&flag-key=supports-one-hour-delivery&return-type=boolean&eval-context=%7B%22locale%22%3A%22ca%22%7D), we have a rule that has two explicitly defined terminal variants (`"on"` and `"off"`), and which operates on the `locale` context property. When this is evaluated, we see the flag value resolved to `true` and the reason is a `"TARGETING_MATCH"`, indicating that we positively assigned a variant based on the context. Importantly, note that if you change the `locale` to `"pl"`, we see `value: false`, but still `reason: "TARGETING_MATCH"`, because our targeting still resulted in a variant.

Alternatively, [this rule](https://flagd.dev/playground/?flags=%7B%22%24schema%22%3A%22https%3A%2F%2Fflagd.dev%2Fschema%2Fv0%2Fflags.json%22%2C%22flags%22%3A%7B%22supports-one-hour-delivery%22%3A%7B%22state%22%3A%22ENABLED%22%2C%22defaultVariant%22%3A%22off%22%2C%22variants%22%3A%7B%22on%22%3Atrue%2C%22off%22%3Afalse%7D%2C%22targeting%22%3A%7B%22if%22%3A%5B%7B%22in%22%3A%5B%7B%22var%22%3A%22locale%22%7D%2C%5B%22us%22%2C%22ca%22%5D%5D%7D%2C%22on%22%5D%7D%7D%7D%7D&flag-key=supports-one-hour-delivery&return-type=boolean&eval-context=%7B%22locale%22%3A%22pl%22%7D), omits a variant for the negative case. This results in the `defaultVariant` being used (our targeting did not match, it "fell through"). This also results in `false` returned for the flag value, but our `reason` is now `"DEFAULT"` not `"TARGETING_MATCH"`. It's also worth noting that omitting the variant is the same as returning `null`, ie: [this](https://flagd.dev/playground/?flags=%7B%22%24schema%22%3A%22https%3A%2F%2Fflagd.dev%2Fschema%2Fv0%2Fflags.json%22%2C%22flags%22%3A%7B%22supports-one-hour-delivery%22%3A%7B%22state%22%3A%22ENABLED%22%2C%22defaultVariant%22%3A%22off%22%2C%22variants%22%3A%7B%22on%22%3Atrue%2C%22off%22%3Afalse%7D%2C%22targeting%22%3A%7B%22if%22%3A%5B%7B%22in%22%3A%5B%7B%22var%22%3A%22locale%22%7D%2C%5B%22us%22%2C%22ca%22%5D%5D%7D%2C%22on%22%2Cnull%5D%7D%7D%7D%7D&flag-key=supports-one-hour-delivery&return-type=boolean&eval-context=%7B%22locale%22%3A%22pl%22%7D) is semantically the same as the previously linked rule).

This is a powerful concept because it allows users to write rules that conditionally match, but otherwise fall through to some default, and provides very granular indications in telemetry that a user was assigned a particular variant due to a specifically defined rule, perhaps targeting only them or their cohort.

@beeme1mr 's proposed change builds on the concept by making the `defaultVariant` itself optional. When omitted, if the targeting "falls through" this communicates "use the provided code default" which might be different across references in code. Currently, since `defaultValue` is required, this is not possible if the flag is active.

docs/architecture-decisions/support-code-default.md
lukas-reining: Yeah as @beeme1mr said before, evaluation reason `flag not found` is good for when the flag does not exist. 
But if I get this right here, the intended semantics of this is to let the flag be existent, but without a value.
This does not exist. 

To me this feels relatively complicated for a problem that could often be reduced to `flag not found`.
What would be the motivation for having the semantics of an `existing flag without value` @beeme1mr?

I think if we add this, we should also add this to the of spec in some way. @beeme1mr @toddbaert

beeme1mr: @lukas-reining The motivation is to have more accurate telemetry. There's an important distinction between a flag that does not exist and a flag that defers resolution back to the client. However, the end result will essentially be the same.

docs/architecture-decisions/support-code-default.md
```diff
@@ -0,0 +1,333 @@
...
+     - JavaScript: Check for `undefined` (not null)
+   - Consider adding proto comments documenting this behavior
+   - Test wire format to confirm fields are actually omitted
+
+   **Example - Correct vs Incorrect Implementation**:
+
+   ```go
+   // INCORRECT - sets zero value
+   response := &ResolveBooleanResponse{
+       Value:   false,  // This sends 'false' on the wire
```
tangenti, line 114: For `optional` singular fields, this should be a pointer. 

Go example in https://github.com/protocolbuffers/protobuf/blob/main/docs/field_presence.md#go-example

tangenti (approved):

toddbaert:
@beeme1mr it looks like we largely have agreement on this proposal.

IMO, the exact way this is accomplished in OFREP is more of an implementation detail. As long as it can be supported, I would say we can accept and merge this.

docs/architecture-decisions/support-code-default.md
beeme1mr: We're going to try to avoid making changes at the protocol level. Inactive flags behave as though they don't exist and we will do something similar if there's no default variant.

docs/architecture-decisions/support-code-default.md
beeme1mr: Yes, exactly.

docs/architecture-decisions/support-code-default.md
beeme1mr: We will use a variation of flag not found with potentially a special error code.

toddbaert:
> @beeme1mr it looks like we largely have agreement on this proposal.
> 
> IMO, the exact way this is accomplished in OFREP is more of an implementation detail. As long as it can be supported, I would say we can accept and merge this.

I'm going to merge this on Monday unless I hear objections. Thanks all, and @beeme1mr

PR #1640, opened 2025-06-06
docs: adr for fractional

toddbaert:
Adds ADR for fractional.

_As written_ this is historical, however, we can modify this one in this PR if you would like. I know @tangenti @dominikhaska @cupofcat are interested in some changes here. Feel free to suggest them and if we have a consensus we can turn the changes into roadmap items.

### DIFF
--- a/docs/architecture-decisions/fractional.md
+++ b/docs/architecture-decisions/fractional.md
@@ -65,7 +65,7 @@ If the bucking value does not resolve to a string, or the `targeting key` is und
 
 #### Bucketing strategy implementation
 
-After retrieving the bucketing value, and hashing it to a [0, 100] range, the algorithm iterates through variants, accumulating their relative weights until finding the bucket containing the hash value.
+After retrieving the bucketing value, and hashing it to a [0, 99] range, the algorithm iterates through variants, accumulating their relative weights until finding the bucket containing the hash value.
 
 ```go
 // Simplified implementation structure


docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,120 @@
...
+for _, distribution := range variants {
+    currentWeight += (distribution.weight * 100) / sumOfWeights
+    if hashValue < currentWeight {
+        return distribution.variant
+    }
+}
+```
+
+This approach supports flexible weight ratios; weights of [25, 50, 25] translate to 25%, 50%, and 25% distribution respectively as do [1, 2, 1].
+It's worth noting that the maximum bucket resolution is 1/100, meaning that the maximum ratio between variant distributions is 1:99 (ie: a weight distribution of [1, 100000] behaves the same as [1, 100]).
```
toddbaert, line 83: IMO this is a limitation we should consider improving. We could make this dynamic based on the weights, to support a practically unlimited resolution (this might come with other costs, like reducing stability of assignments) or we could choose a higher number, like 1000, or even make this number an optional parameter.

toddbaert: We also might want to specify a max integer size in any case, considering precision and range constraints.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,120 @@
...
+    ["green", 30]
+  ]
+}
+```
+### Consequences
+
+* Good, because Murmur3 is fast, has good avalanche properties, and we don't need "cryptographic" randomness
+* Good, because we have flexibility but also simple shorthand
+* Good, because our bucketing algorithm is relatively stable when new variants are added
+* Bad, because we only support string bucketing values
```
toddbaert: I think non-string bucketing values were of interest?

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,120 @@
...
+* Bad, because we don't have bucket resolution finer than 1:99
+* Bad because we don't support JSONLogic expressions within bucket definitions
```
toddbaert: I think this was a limitation that was also questioned.

docs/architecture-decisions/fractional.md
askpt: We can default to `100` and then allow the user, by configuration, to change the ratios. We would need to make sure the sum is respected and what happens when it is not.

beeme1mr: I would prefer to change the bucket size to something larger (e.g. 10_000) as resizing the bucket size can lead to unexpected behavior.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+This non-cryptographic hash function provides excellent performance and good avalanche properties (small input changes produce dramatically different outputs) while maintaining deterministic behavior essential for sticky evaluations.
+Its wide language implementation ensures identical results across different flagd providers, no matter the language in question.
+
+#### Bucketing value
+
+The bucking value is an optional first value to the operator (it may be a JSONLogic expression, other than an array).
+This allows enables targeting based on arbitrary attributes (individual users, companies/tenants, etc).
+If not specified, the bucketing value is a JSONLogic expression concatenating the `$flagd.flagKey` and the extracted [targeting key](https://openfeature.dev/specification/glossary/#targeting-key) (`targetingKey`) from the context (the inclusion of the flag key prevents users from landing in the same "bucket index" for all flags with the same number of buckets).
+If the bucking value does not resolve to a string, or the `targeting key` is undefined, the evaluation is considered erroneous.
```
beeme1mr, line 54: The `targeting key` check is only valid when users haven't specified a custom bucketing value. In those situations, the users are responsible for checking if the value exists.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+```
+
+### Consequences
+
+- Good, because Murmur3 is fast, has good avalanche properties, and we don't need "cryptographic" randomness
+- Good, because we have flexibility but also simple shorthand
+- Good, because our bucketing algorithm is relatively stable when new variants are added
+- Bad, because we only support string bucketing values
+- Bad, because we don't have bucket resolution finer than 1:99
+- Bad because we don't support JSONLogic expressions within bucket definitions
```
beeme1mr:
```suggestion
- Bad, because we don't support JSONLogic expressions within bucket definitions
```

For consistency

beeme1mr (commented):
Is the goals of this ADR to capture the current state of the fractional operator or the ideal state? If it's the latter, I'd like to define the bucket size and hashing key constraints before approving.

docs/architecture-decisions/fractional.md
toddbaert: I think you're right but I wonder if we should improve this.

toddbaert:
> Is the goals of this ADR to capture the current state of the fractional operator or the ideal state? If it's the latter, I'd like to define the bucket size and hashing key constraints before approving.

@beeme1mr my thinking with this was the capture the current state, and we could immediately open a new PR to outline our hopes.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+Since flagd seeks to implement a full feature flag evaluation engine, such a feature is required.
+
+## Requirements
+
+- **Deterministic**: must be consistent given the same input (so users aren't re-assigned with each page view, for example)
+- **Performant**: must be quick; we want "predictable randomness", but with a relatively low performance cost
+- **Ease of use**: must be easy to use and understand for basic use-cases
+- **Customization**: must support customization, such as specifying a particular context attribute to "bucket" on
+- **Stability**: adding new variants should result in new assignments for as small a section of the audience as possible
+- **Strong avalanche effect**: slight input changes should result in relatively high chance of differential bucket assignment
```
tangenti, line 24: I feel that some of the requirements here are the consequences of the hash algorithm we picked, not necessarily requirements.

Wearing the hat of a user, I would like the actual distributions as close as what I specified in the weights.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Considered Options
+
+- We considered various "more common" hash algos, such as `sha1` and `md5`, but they were frequently slower than `Murmur3`, and didn't offer better performance for our purposes
```
tangenti, line 28: Is hashing the bottleneck of an evaluation? If not, this does not sound a good reason to not use other algorithms.

docs/architecture-decisions/fractional.md
tangenti, line 54: Could you help me understand more about only allowing string?

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+  "cat": [
+    {"var": "$flagd.flagKey"},
+    {"var": "targetingKey"}
+  ]
+}
+```
+
+#### Bucketing strategy implementation
+
+After retrieving the bucketing value, and hashing it to a [0, 100] range, the algorithm iterates through variants, accumulating their relative weights until finding the bucket containing the hash value.
```
tangenti: nit: [0, 100) or [0, 99]

docs/architecture-decisions/fractional.md
tangenti: > or even make this number an optional parameter.

This sounds a good approach - it's harder for users to accidentally change the distribution as they have to specify the total weight, and meanwhile we're not limiting the precision. We could start by allowing a list of values as the parameter (100, 1000, 100,000 etc) or setting a maximum value allowed.

docs/architecture-decisions/fractional.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+**Longhand format** allows precise weight control through two-element arrays:
+
+Note that in this example, we've also specified a custom bucketing value.
+
+```json
+{
+  "fractional": [
```
tangenti, line 105: Having variants as the only allowed value won't support some use cases of A/B testing straightforward (they could probably write some conversions to generate the expression), for example, 

I would like 30% of population to assign to group A, and 70% to group B. 
- For group A, if user is from OpenFeature, assign variant P, otherwise Q
- For group B, if user is from OpenFeature, assign variant M, otherwise N.

toddbaert: This makes a lot of sense. I think it's reasonable to want this enhancement.

docs/architecture-decisions/fractional.md
beeme1mr: There is an advantage to this variability from an A/B testing perspective. It wasn't a primary reason for choosing this algorithm but it was at least considered when evaluating the various options.

docs/architecture-decisions/fractional.md
beeme1mr: It could be a bottleneck since hashing can be fairly CPU-intensive, and this could run multiple times for every request.

docs/architecture-decisions/fractional.md
toddbaert:
```suggestion
After retrieving the bucketing value, and hashing it to a [0, 99] range, the algorithm iterates through variants, accumulating their relative weights until finding the bucket containing the hash value.
```

toddbaert:
Since this is historical, I'm merging this.

That doesn't mean it can't be edited or superseded, it's simply an explanation of a previous decision and reflects the current implementation.

PR #1641, opened 2025-06-10
feat: updating context using headers

### LINKED ISSUE #1583: [FEATURE] Header-based context injection
### Requirements

The `--context-value` / `-X` startup option allows flagd users to add arbitrary string key/values to all evaluations. This is great for statically configured values, such as a the server's geography or the cloud provider it's running on. For RPC evaluations (evaluations taking place on flagd) users can send other context dynamically in the evaluation request body... However, the ability to add context values from HTTP/connect-rpc headers is also useful, since it can be done without requiring users to set them directly in the client, but instead can be done via infrastructure and proxies. We should add the ability to map arbitrary headers to context values.

Requirements:

- support HTTP header mapping in [OFREP](https://github.com/open-feature/flagd/blob/779e8f929af3a0875a0c5d05c220077cc5f09973/flagd/pkg/service/flag-evaluation/ofrep/handler.go#L28) requests
- support HTTP header mapping in [evaluation service v2](https://github.com/open-feature/flagd/blob/779e8f929af3a0875a0c5d05c220077cc5f09973/flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go#L32) via [connect](https://connectrpc.com/docs/go/headers-and-trailers/).
- command should look something like `--context-from-header "X-Forwarded-For:remoteIp"` or `-H "X-Forwarded-For:remoteIp"` (map the contents of the `X-Forwarded-For` header into a context property called `remoteIp`).

Considerations:

- only string:string mapping is supported

Rahul-Baradol:
Able to update context map using headers present in 
- OFREP requests
- Connect Requests (via Flag Evaluator V2 service)

### Related Issues
Fixes #1583 

### Notes
Context values passed via headers is high priority

If same context key is updated via
- Headers
- Request Body
- Static Config

_Context via Headers will be considered_

### How to test

_Todo_


flagd/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -287,11 +296,13 @@ func (s *OldFlagEvaluationService) ResolveObject(
 		sCtx,
 		s.logger,
 		s.eval.ResolveObjectValue,
+		req.Header(),
 		req.Msg.GetFlagKey(),
 		req.Msg.GetContext(),
 		&objectResponse{schemaV1Resp: res},
 		s.metrics,
 		s.contextValues,
+		make(map[string]string),
```
aepfli, line 305: ~~[question] If we're adding this new map to all calls, do we need this parameter? It is always the same value, and always a fresh map.~~ 

I see why this is happening, but grpc streams can also contain headers, should we also utilize those?

flagd/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -301,28 +312,34 @@ func (s *OldFlagEvaluationService) ResolveObject(
...
+		if values, ok := headers[header]; ok {
+			merged[contextKey] = values[0]
+		}
+	}
 	return merged
 }
 
 // resolve is a generic flag resolver
-func resolve[T constraints](ctx context.Context, logger *logger.Logger, resolver resolverSignature[T], flagKey string,
+func resolve[T constraints](ctx context.Context, logger *logger.Logger, resolver resolverSignature[T], header http.Header, flagKey string,
```
aepfli, line 334: [nit] everytime i see another parameter added to this method my eyes is bleeding more and more. - I understand why we are extending this method, to have the logic centralized, but maybe we could also resolve this with a middleware on the whole connection, and somehow merging the contextValues with the headerContextValues already within the middleware, and removing the parameters to just one. (this is not blocking for me, it is just getting harder and harder to read this method and understand what all those parameters are actually doing)

aepfli (approved):
Thank you. The logic looks good to me for OFREP, I am wondering if we also want to add the header logic for the RPC-evaluations as GRPC requests can also contain headers. Additionally I am wondering if we could also resolve this with a middleware, and generate this information within the middleware. this would reduce the overall code and instrumentation for the contexts in general (but I am not sure if this is a feasible idea - eg. like suggested [here](https://drstearns.github.io/tutorials/gomiddleware/#secanalternative) as we do not have many endpoints this could be feasible)

flagd/pkg/service/flag-evaluation/flag_evaluator.go
Rahul-Baradol: I only changed `FlagEvaluationService`, and left `OldFlagEvaluationService` untouched because I believe the feature is asked in flag evaluation service - v2

But we can also add  this to `OldFlagEvaluationService`, if we would want to have it in old flag evaluation service as well!

flagd/pkg/service/flag-evaluation/flag_evaluator.go
Rahul-Baradol: How about making the call to `mergeContexts` at the caller of `resolve` (i.e at `Resolve*`)?

That way we can reduce 4 arguments at resolve (`evaluationContext`, `configContextValues`, `header`, `configHeaderToContextKeyMappings`), 
and they are not used by `resolve` other than for merging!

What do you think!?

toddbaert: I think less parameters passed is always better if it comes at no other cost or duplication.

If you try it and it adds other complexity, please just keep it as is, and we can perhaps due a subsequent refactor.

flagd/pkg/service/flag-evaluation/flag_evaluator.go
```diff
@@ -301,28 +312,34 @@ func (s *OldFlagEvaluationService) ResolveObject(
 	return res, err
 }
 
-// mergeContexts combines values from the request context with the values from the config --context-values flag.
-// Request context values have a higher priority.
-func mergeContexts(reqCtx, configFlagsCtx map[string]any) map[string]any {
+// mergeContexts combines context values from headers, static context (from cli) and request context. 
+// highest priority > header-context-from-cli > static-context-from-cli > request-context > lowest priority
```
toddbaert, line 316: Thanks for comments :heart:

flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
```diff
@@ -992,6 +1006,8 @@ func Test_mergeContexts(t *testing.T) {
 			args: args{
 				clientContext: map[string]any{"k1": "v1", "k2": "v2"},
 				configContext: map[string]any{"k2": "v22", "k3": "v3"},
+				headers:       http.Header{},
+				headerToContextKeyMappings: map[string]string{},
```
toddbaert, line 1010: Can we add a test or 2 here for the merging assertions involving the new params?

I think the is the primary functional test we'd need, unless I'm missing something.

toddbaert (commented):
Amazing work! :partying_face: 

Tested manually and works exactly as expected:

Started flagd with:

`go run main.go start -f file:../config/samples/example_flags.flagd.json  -H X-Color=color`

Tested using the isColorYellow flag, which returns true if color == yellow:

Connect CURL:

```
curl -X POST "localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" -d '{"flagKey":"isColorYellow","context":{}}' -H "Content-Type: application/json" -H "X-Color:yellow" | jq .value // true
```

```
curl -X POST "localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" -d '{"flagKey":"isColorYellow","context":{}}' -H "Content-Type: application/json" -H "X-Color:red" | jq .value // false
```

OFREP:

```
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/isColorYellow' -H "X-Color:yellow" | jq .value // true
```

```
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/isColorYellow' -H "X-Color:red" | jq .value // false
```

Furthermore, I tried it in combination with -X, which combined with the new -H , creates a kind of "default":

`go run main.go start -f file:../config/samples/example_flags.flagd.json  -H X-Color=color -X color=yellow`

When I do this, we get the expected "defaulting" behavior:

```
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/isColorYellow' -H "X-Color:red" | jq .value // false (we overrode the static context
```

```
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/isColorYellow'  | jq .value // true (since we didnt override the static context)
```

Lastly, the "casing" behavior seems right... The casing of the header doesn't matter, but the casing of the context attr does:

```
curl -X POST "localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" -d '{"flagKey":"isColorYellow","context":{}}' -H "Content-Type: application/json" -H "x-color:yellow" | jq .value  // true
```

```
curl -X POST "localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" -d '{"flagKey":"isColorYellow","context":{}}' -H "Content-Type: application/json" -H "X-Color:Yellow" | jq .value // false
```
I think the only blocker for my approval is a missing unit test [here](https://github.com/open-feature/flagd/pull/1641#discussion_r2140968396).

flagd/pkg/service/flag-evaluation/flag_evaluator_v2_test.go
Rahul-Baradol: Sure!

beeme1mr (approved):
The change looks good to me. As a follow up, we should update the documentation to mention this feature.

Rahul-Baradol:
If you are talking about `docs/reference/flagd-cli/flagd_start.md`, then it has been updated already!
Am I missing something else?

toddbaert:
> If you are talking about `docs/reference/flagd-cli/flagd_start.md`, then it has been updated already! Am I missing something else?

No, I think @beeme1mr is referring more to some "human readable" conceptual docs, but I'm happy to do that in a separate PR. I will ping you on it so you can review what I write.

toddbaert (approved):
I'll merge this in the next 24 hrs unless there's objections.

toddbaert (approved):

PR #1642, opened 2025-06-11
feat: add sync_context to SyncFlags

### LINKED ISSUE #1635: [FEATURE] add sync_context to SyncFlags payload
### Requirements

https://github.com/open-feature/flagd-schemas/pull/183 deprecated the GetMetadata call. Now the sync stream can return the same data, saving providers an RPC call. Providers will do exactly the same thing with this information as they currently do with the `GetMetadata` call.

Requirements:

- implement `sync_context` field in `SyncFlags`, which has the same payload as `GetMetadata`
- keep `GetMetadata` for now, for reverse compatibility
- add associated tests

bbland1:
## This PR

- adds comment about `GetMetadata` being deprecated in future release
- implements the `sync_context` field to the `SyncFlags` of `flag-sync/handler.go`

### Related Issues

Fixes #1635


toddbaert:
Hey @bbland1 ! Sorry for the delay. I will review this early next week or possibly over the weekend.

toddbaert:
> Thank you, I have nothing to add to the implementation, more like how we will proceed here, so we will not create breaking changes for everyone ;) - how do we know if the data is in the payload or a separate request. but this is something we might want to discuss on the issue rather here.

My thinking is we just leave this as it is (with no breaking changes and both working) for a long time* and update the providers to use the new feature as soon as possible. That will minimize the chance anyone is impacted when we actually remove this feature (probably in v1).

\* depending on how fast we can implement this in providers

flagd/pkg/service/flag-sync/handler.go
```diff
@@ -43,7 +44,24 @@ func (s syncHandler) SyncFlags(req *syncv1.SyncFlagsRequest, server syncv1grpc.F
...
+			metadataSrc := make(map[string]any)
+			maps.Copy(metadataSrc, s.contextValues)
+
+			if sources := s.mux.SourcesAsMetadata(); sources != "" {
+				metadataSrc["sources"] = sources
+			}
+
+			metadata, err := structpb.NewStruct(metadataSrc)
+			if err != nil {
+				s.log.Warn(fmt.Sprintf("error from struct creation: %v", err))
```
toddbaert: any reason why this is a warning log not an error?

toddbaert (approved):
@bbland1 great work! I tested this manually and it works as expected:

image


I've pushed 2 very small changes:

- `go mod tidy` in `/core`
- updated the schema submodule - I'm guessing you have this locally but didn't commit it, because it doesn't work otherwise. It can be tricky to update submodules, my method is usually to `cd` into it, and then checkout what I want (in this case the [0.6.4 proto release](https://github.com/open-feature/flagd-schemas/releases/tag/protobuf-v0.6.4)) and then commit the resultant submodule metadata change from the root directory (not the submodule dir you moved into)

toddbaert:
I'll merge this soon unless I hear objections.

flagd/pkg/service/flag-sync/handler.go
bbland1: @toddbaert no particular reason that I recall, happy to switch it to an error. I think at the time I was really trying to just move the logic from the `GetMetadata` call and minimize too many changes that I couldn’t track and didn’t ever circle back to weigh the option between the 2.

bbland1:
> 
> I've pushed 2 very small changes:
> 
> * `go mod tidy` in `/core`
> * updated the schema submodule - I'm guessing you have this locally but didn't commit it, because it doesn't work otherwise. It can be tricky to update submodules, my method is usually to `cd` into it, and then checkout what I want (in this case the [0.6.4 proto release](https://github.com/open-feature/flagd-schemas/releases/tag/protobuf-v0.6.4)) and then commit the resultant submodule metadata change from the root directory (not the submodule dir you moved into)

Happy to see it working on your end too! Yeah I updated it locally but I didn’t realize it wouldn’t push until you just said this so that makes total sense, thanks! Now I can make a note of it for the future. Nice catch on the `go mod tidy` in `/core`!

chrfwow (approved):

PR #1644, opened 2025-06-16
docs(ADR): decouple flag sources and flag sets 

tangenti:
The ADR proposes a different approach to solve the problems stated in https://github.com/open-feature/flagd/pull/1634. 

It's quite concise and simple at the point, assuming the reviewers already have the context from the original discussions.

Will add more details after the initial feedback and inputs.


docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Background
+
+Flagd daemon syncs flag configurations from multiple sources. A single source provides a single config, which has an optional flag set ID that may or may not change in the following syncs of the same source.
+
+The in-process provider uses `selector` to specify the desired source.  In order to get a desired flag set, a provider has to stick to a source that provides that flag set. In this case, the flagd daemon cannot remove a source without breaking the dependant flagd providers.
+
+Assumptions of the current model
+
+- `flagSetId`s must be unique across different sources or the configuration is considered invalid.
```
guidobrei, line 20: What does this mean exactly? Does it mean that a flag set id is only allowed to be contained in a single source?

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+- In-process providers can select based on flag sets.
+- No breaking changes for the current usage of `selector`
+
+## Proposal
+
+### API change
+
+#### Flag Configuration Schema
+
+Add an optional field `flagsetID` under `flag` or `flag.metadata`. The flag set ID cannot be specified if a flag set ID is specified for the config.
```
guidobrei, line 35: This nicely fits into the existing [flagSetId defined in the schema](https://github.com/open-feature/flagd-schemas/blob/bb763438abc5b0e97dc26a795bc72b7a9f3e4020/json/flags.json#L57).

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+```yaml
+# Flags from the source `override`
+selector: override
+
+# Flags from the source `override`
+selector: source=override
+
+# Flags from the flag set `project-42`
+selector: flagsetID=project-42
```
guidobrei, line 51: If we use the flagSetId from the schema it should be something like `metadata.flagSetId=project42`.

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+Selector will be extended for generic flags selection, starting with checking the equivalence of `source` and `flagsetID` of flags.
+
+Example
+
+```yaml
+# Flags from the source `override`
+selector: override
+
+# Flags from the source `override`
+selector: source=override
```
guidobrei, line 48: Wondering if we should inject the source as metadata prop into every flag. Then we wouldn't have to distinguish between selecting a source, or flagSetId, or anything else.

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: This is the assumption held in PR #1634. 

My interpretation for this is: it's an undefined behavior if multiple sources provides different flag set IDs for the same flag key.

guidobrei: According to flagd syncs [merge strategy](https://flagd.dev/concepts/syncs/#s3-sync) I would expect the last sync to win.

Maybe we should add an additional underlying assumption, that the `flagId` must be unique across different flag sets as well.

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: I've considered the option and didn't chose it, mostly to keep the syntax simple for now (so it can extended later in a flexible way).

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: This cannot be done in the flag config as the source itself does not know its name. In the flag store, it's currently under `model.flag` as a separate field and `metadata` is a map. On the technically perspective keeping it as a field (especially it's required) it is a better option to me. On the semantic level (for `selector`), `metadata.source` makes more sense to me.

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: Yes, this is the current behavior as there lacks a definition of order/priority of sources/flagsets. It's not a very good approach as the result is NOT deterministic and depends on the order of concurrent events.

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+updated: 2025-06-16
+---
+
+# Decouple flag sync sources and flag sets
+
+The goal is to support dynamic flag sets for flagd providers and decouple sources and flag sets.
+
+## Background
+
+Flagd daemon syncs flag configurations from multiple sources. A single source provides a single config, which has an optional flag set ID that may or may not change in the following syncs of the same source.
```
toddbaert, line 14: > which has an optional flag set ID that may or may not change in the following syncs of the same source.

This is a good point. I think we need to make sure we _do_ support changing this value and all the expected behavior associated with that.

docs/architecture-decisions/decouple-flag-source-and-set.md
toddbaert: Yes I agree this is a limitation in our current modal which we need to make more robust.

docs/architecture-decisions/decouple-flag-source-and-set.md
toddbaert, line 35:
```suggestion
Add an optional field `flagsetId` under `metadata` or `flags.{flagKey}.metadata`. The flag set ID cannot be specified if a flag set ID is specified for the config.
```

Feel free to accept or not, but this is what we currently support in our config, it just doesn't have most of the behavior you describe, but I think such behavior could be added using these existing fields. See [playground](https://flagd.dev/playground/?scenario-name=Flag+metadata).

Also [here](https://flagd.dev/schema/v0/flags.json).

image

toddbaert: Agreed, see [here](https://github.com/open-feature/flagd/pull/1644/files#r2150722702).

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+The semantic can later be extended with a more complex design, such as AIP-160 filter or Kubernetes selections. This is out of the scope of this ADR.
+
+### Flagd Daemon Storage
+
+1. Flagd will have separate stores for `flags` and `sources`
+
+2. `selector` will be removed from the store
+
+3. `flagSetID` will be added as part of `model.Flag` or under `model.Flag.Metadata` for better consistency with the API.
```
toddbaert, line 62: From an implementation perspective, do you see a benefit in changing flagd's internal representation of flags to support the `flagsetID=project-42` use-case optimally? If we expect this to be a common selection, we could benefit from actually "storing" flags in memory as a "map of maps" (a map of flagSets) so that sets can easily be "plucked" instead of iterating.

Not critical for this ADR, just wondering.

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: I'm aware of that, and this is the config-level flag set ID that I mentioned in the second sentence.

It will be used as the flag set ID for every flag in the config, if the flag level flag set ID is not set.

docs/architecture-decisions/decouple-flag-source-and-set.md
tangenti: > we could benefit from actually "storing" flags in memory as a "map of maps" (a map of flagSets) so that sets can easily be "plucked" instead of iterating.

I have the same idea as what Guido brought up in [this comment](https://github.com/open-feature/flagd/pull/1644#discussion_r2149723952) - flag set ID is just as the other metadata of a flag, and I don't see a good reason to structure the internal storage by flag set ID.

Indexing is a different problem than storing. We could build indexing over flag sets and sources if that iterating flags turns out to be a performance issue (which I doubt).

tangenti:
> I can support this proposal. I like the query expression idea. My main concern is if this can/should be done with the existing `flagSetId` field. I would certainly advocate for that to reduce breakage and redundancy. See https://github.com/open-feature/flagd/pull/1644/files#r2150722702.

Thanks Todd. This proposal is adding another more granularity to the existing flag set ID - it will be kept and used if the server would like to keep a config-level flag set ID.

toddbaert:
The more I think about this proposal, the more I like it. They don't need to be supported now, but I can think of a few very realistic "selector queries" beyond the examples you've posted:

- `selector: flagType=boolean` // select all boolean flags
- `selector: metadata.clientSide=true` // select all client side flags @beeme1mr pointed this out, for some of our use cases it would be great to be able to filter serverside/clientside flags

That said, I do also have some implementation concerns which I think should be discussed briefly before we can accept it, even if we are only currently supporting `source` and `flagSet` selection . Specifically, I'm worried that for evaluations actually done in flagd (OFREP and the `evaluation.proto`) that when large amounts of flags are in storage, an expression style selector could come at significant performance cost since for `flagSetId` (or any other property we might support selecting on later), we'd need to iterate over all the flags to find matches. This is something like what we're doing now, but the [alternative proposal](https://github.com/open-feature/flagd/pull/1634) implied a "pluck" based on `flag set` (retrieving from a hash).

However, this made me wonder if perhaps we should use some kind of in-memory DB (for example https://github.com/hashicorp/go-memdb)... this way we could offload all sorts of query code to that, benefiting from the performance of that DB, and accept various sorts of expressions as the selector.

cc @beeme1mr @chrfwow @alexandraoberaigner @guidobrei @aepfli

docs/architecture-decisions/decouple-flag-source-and-set.md
toddbaert: @tangenti for example, if the `flagSetId` is removed from a flag, we probably need to recompute the sets for listeners on that flagSet not to include that flag now. To put it another way, changes in the flag set could result in effective removal from listener sets. This isn't a problem, but we need to make sure we implement it this way, I think.

tangenti:
Thanks @toddbaert,

When you talk about a large amount of flags, do you have a solid number in mind of what this means? 1k flags and 1M flags require different optimizations for implementations.

I think a good approach here is to do a quick measurement with a reasonable number of flags on the critical path, especially the time it takes to recalculate the flags for each sub/listener.

For the idea of an in-memory DB, I think there are more questions to answered:
1. We need to have a well-defined interface for storage first so the storage can be switched. Right now the interface and the struct are used mixedly, which doesn't support different types of storage.
1. Currently the indexing for each sub is done separately from the flag storage. We should consider encapsulate that into another interface or with the flag store interface, so we can use the native indexing from the chosen database.
1. MVCC. I can think of two usages of MVCC: as another options compared to resyncs when a source is removed; (more for evaluation) a fallback mechanism if the latest flag is not valid.
1. Transaction. Currently there's a store level lock. We may want to use transactions instead to better handle the concurrent updates/queries.

docs/architecture-decisions/decouple-flag-source-and-set.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Flags Sync
+
+Sync server would count the extended syntax of `selector` and filter the list of flags on-the-fly answering the requests from the providers.
+
+The existing conflict resolving based on sources remains the same. Resyncs on removing flags remains unchanged as well.
```
chrfwow, line 68: > `The existing conflict resolving based on sources remains the same.`

I think this may be a problem. Iirc, our current conflict resolving strategy is to only use the last flag definition with a key, overriding flags from other flag sets if they share the same key. We could fix this by e.g. prepending the flag set id to the flag key, but I would like a solution that can handle duplicate flag keys in different flag sets natively.

toddbaert:
> Thanks @toddbaert,
> 
> When you talk about a large amount of flags, do you have a solid number in mind of what this means? 1k flags and 1M flags require different optimizations for implementations.

I'd say on the order of 10k flags for our use case.

> For the idea of an in-memory DB, I think there are more questions to answered:

> We need to have a well-defined interface for storage first so the storage can be switched. Right now the interface and the struct are used mixedly, which doesn't support different types of storage.
Currently the indexing for each sub is done separately from the flag storage. We should consider encapsulate that into another interface or with the flag store interface, so we can use the native indexing from the chosen database.
MVCC. I can think of two usages of MVCC: as another options compared to resyncs when a source is removed; (more for evaluation) a fallback mechanism if the latest flag is not valid.
Transaction. Currently there's a store level lock. We may want to use transactions instead to better handle the concurrent updates/queries.

Agreed.

But brings up another point: **What this ADR doesn't fully address is still a means to define multiple sets of colliding flags in a single sync resource**. At least for our internal purposes, while we can use the `flagSetId` to logically group and select flags in flagd as described in this proposal, we specifically need to support flags with the same key (from different projects) in a single source (http endpoint, file, etc). The current schema doesn't support that - even if the flags with duplicate keys have different `flagSetId`s which they can be _logically_ "namespaced" by, they cannot exist in the same JSON document and be parsed with most JSON parsers (or specifically Go's). This is technically allowed in JSON, but every parser I'm aware of will drop one of these flags:

```json
{
  "flags": {
    "flag1": {
      "state": "ENABLED",
      "defaultVariant": "true",
      "variants": {
        "true": true,
        "false": false
      },
      "targeting": {}
    },
    "flag1": {
      "state": "ENABLED",
      "defaultVariant": "false",
      "variants": {
        "true": true,
        "false": false
      },
      "targeting": {}
    }
  }
}
```

Currently I think others are dealing with this by prepending some sort of prefix to flag keys and then removing it, but that's exactly the thing we'd hoped the proposal [here](https://github.com/open-feature/flagd/pull/1634) would remove the need for.

cc @juanparadox as I believe this is your challenge too.

Could we support the selector expressions in this proposal, but the schema changes from #1634 ?

tangenti:
> But brings up another point: What this ADR doesn't fully address is still a means to define multiple sets of colliding flags in a single sync resource.

This is a new requirement and seems not brought up in #1634? I'm not so sure if this is broadly applicable to most user cases.  I think we'll need more discussions to understand the requirement better, in particular

1. It strongly implicates a multi-tenancy set up with isolation of flags between tenants. Each tenant will get different values for a flag. 
2. The implication also means that the flags with the same key in different flag sets SHOULD have the same type, but this is NOT reflected in the requirement itself or the schema proposed in #1634 .
3. As a result, it seems to enforce the provider to select one and only one flag sets. Let's say there a flag `new-feature` with the type boolean in flag set A, and `new-feature` with type string in flag set B, then how should we resolve the conflict?

I believe multi-tenancy support is needed, but the current approach increases the complexity a lot and doesn't provide full isolation from flagd: the resources are not isolated so a high QPS from one tenant can make flagd unavailable for all other tenants; data level isolation seems not guaranteed as well

toddbaert:
> > But brings up another point: What this ADR doesn't fully address is still a means to define multiple sets of colliding flags in a single sync resource.
> 
> This is a new requirement and seems not brought up in #1634? I'm not so sure if this is broadly applicable to most user cases. I think we'll need more discussions to understand the requirement better, in particular
> 
> 1. It strongly implicates a multi-tenancy set up with isolation of flags between tenants. Each tenant will get different values for a flag.

Perhaps we did a poor job communicating some of the requirements. This is one of the main challenges we're facing at Dynatrace, and I think @juanparadox is having the same issue with their adoption. We tried to describe that [here](https://github.com/open-feature/flagd/pull/1634/files#diff-6853e882f2176aa424d2231bb3338bd0d55333c63f7c69acaee86c9516b73969R24) but I agree it wasn't made very obvious.

> 2. The implication also means that the flags with the same key in different flag sets SHOULD have the same type, but this is NOT reflected in the requirement itself or the schema proposed in [docs(ADR): propose support for flag set selection #1634](https://github.com/open-feature/flagd/pull/1634) .

> 3. As a result, it seems to enforce the provider to select one and only one flag sets. Let's say there a flag `new-feature` with the type boolean in flag set A, and `new-feature` with type string in flag set B, then how should we resolve the conflict?

I'm not sure why it's implied that the flags must be the same type when their keys match across sets, maybe I'm missing something... I think the set you select comes with implicit assumptions about the flag types within that set. I'm not sure I see a problem with a situation where  `provider1` selects `flagSetA` and resolves `flag1` as a boolean, while `provider2` selects `flagSetB` and resolves `flag1` as a string; their selector has resulted in a different set of flags, so what's the issue with the types being different? If their selector results in a duplicate flag, we have to resolve it in any case.

This just seems like a special case of flag-keys existing in multiple sets. :thinking: If it were to happen, you'd have to correct your selector to get the flag you expect to evaluate.

> I believe multi-tenancy support is needed, but the current approach increases the complexity a lot and doesn't provide full isolation from flagd: the resources are not isolated so a high QPS from one tenant can make flagd unavailable for all other tenants; data level isolation seems not guaranteed as well

I think some kind of client rate-limiting could be discussed but it's out of scope of this proposal. "Tenants" in our case are just different teams and they are already sharing a central flagd implementation that's not isolating "noisy neighbors". All our clients are doing local evaluation anyway.

I think a possible solution could be to accept the selector proposal here, and the schema changes in #1634. With that, we'd have:

- a coherent and flexible selector semantic
- a backwards compatible schema that supports `flagSetIds` (ie: through a set of flag sets)

If you think a meeting is necessary I'm open to that.

docs/architecture-decisions/decouple-flag-source-and-set.md
toddbaert, line 35: Unless I'm misunderstanding, this already exists.

toddbaert (approved):
I agree with this approach overall, but @tangenti you may want to review it in light of our recent decisions.

PR #1645, opened 2025-06-21
feat: Add support for HTTP eTag header and 304 no change response

### LINKED ISSUE #1558: [FEATURE] Support E-Tag for HTTP sync
### Problem statement

Currently, "HTTP sync" fetches the body and decides whether to update the flags based on [SHA changes](https://github.com/open-feature/flagd/blob/main/core/pkg/sync/http/http_sync.go#L96).

This process is performed at every polling iteration, which may cause an overhead.

### Proposal

HTTP source can support [E-Tag](https://en.wikipedia.org/wiki/HTTP_ETag).

If E-Tag presented in response headers by server, then HTTP sync can add `If-None-Match` with the value of E-Tag in headers for further requests.
Server will return response with actual body, only if E-Tag of the content has changed (304 Not Modified otherwise).
This will help to avoid blank requests, when nothing changed in HTTP source.

ChihweiLHBird:
## This PR
- adds the support for the `eTag` request header and `304 Not Modified` response.

### Related Issues
Fixes #1558

### Notes
This proposal includes some significant behavior changes; therefore, any feedback, opinions, or objections are welcome and appreciated.

### How to test
```bash
make build
cd bin && ./flagd start --port 8013 --uri https://raw.githubusercontent.com/open-feature/flagd/main/samples/example_flags.flagd.json --debug
```

More specific test cases to be added when we all agree on proceeding with the implementation of the change in this PR.


beeme1mr (approved):

ChihweiLHBird:
@aepfli Thank you for the suggestion! May I know which README should be changed? The one in the root directory of the repository doesn't seem to be describing sync behaviors?

beeme1mr:
@ChihweiLHBird, it may be worth mentioning the eTag behavior [here](https://flagd.dev/concepts/syncs/#http-sync). Thanks!

beeme1mr:
I'll merge this now since it should be transparent to end users.

ChihweiLHBird:
Hi @beeme1mr, I just added some description, but I am not very good at doc writing. So, let me know if you want to rephrase/rewrite my doc or anything. Thank you!

toddbaert:
I'm going to review this myself today and fix the conflict, then I'll merge.

core/pkg/sync/http/http_sync_test.go
```diff
@@ -240,6 +241,85 @@ func TestHTTPSync_Fetch(t *testing.T) {
...
+						t.Fatalf("expected If-None-Match header to be '%s', got %s", expectedIfNoneMatch, actualIfNoneMatch)
+					}
+
+					newContent := "\"Hey there!\""
+					newETag := `"c2e01ce63d90109c4c7f4f6dcea97ed1bb2b51e3647f36caf5acbe27413a24bb"`
+
+					return &http.Response{
+						Header:     map[string][]string{
+							"Content-Type": {"application/json"},
+							"Etag":        {newETag},
```
ChihweiLHBird, line 295: One thing worth mention here is that the `http` package apparently doesn't handle case insensitive well and I had to make this testing header to be `Etag` (small case `t`), even though in real world usage (non-testing), `ETag` works well.

ChihweiLHBird: I guess building the header with the `Header().Set()` function rather than passing in a raw map will fix it but I didn't have time to try at this time.

PR #1648, opened 2025-06-24
docs: refines the code default adr

beeme1mr:
## This PR

- updates the code default ADR to use `FLAG_NOT_FOUND` errors codes instead of an undefined value.

### Notes

This update will have nearly identical end results but will be easier to implement and more consistent. Core to this change is switching from using an empty value as the indicator to using the `FLAG_NOT_FOUND` error code. While this may seem odd at first, it's in line with the behavior a user sees when first releasing a disabled flag. This will allow users to `create a new disabled flag` -> `enable the flag for a subset of users without impacting others` -> `enable the flag for everyone` -> `deprecate the flag`.


toddbaert:
@lukas-reining @thomaspoignant this is the result of some of the discussions you had with @beeme1mr as far as I know, specifically in terms of OFREP not supporting missing values.

toddbaert (approved):
I'm totally in favor of a more robust means of communicating `INACTIVE` flags at some point, including flags with no `defaultVariant`, but to properly fix that we'd need to revisit flagd's current behavior for that (`FLAG_NOT_FOUND).

toddbaert:
I've created updated issues [here](https://github.com/open-feature/flagd/issues/1646).

docs/architecture-decisions/support-code-default.md
```diff
@@ -197,10 +118,14 @@ flags:
...
 
+A single flag evaluation returns a `404` status code.
+
 ```json
 {
   "key": "my-feature",
-  "reason": "DEFAULT",
+  "errorCode": "FLAG_NOT_FOUND",
+  // Optional error details
+  "errorDetails": "Targeting not matched, using code default",
```
guidobrei, line 128: Should we extend the message with a hint that the flag did not specify a default value?
`Targeting not matched, no default value defined, using code default`

guidobrei (commented):
Sorry, I'm late to the party but have a minor comment on the error details.

PR #1659, opened 2025-06-29
feat: allowing null/missing defaultValue

### LINKED ISSUE #1647: [FEATURE] Update flagd-core, RPC and OFREP Evaluations to Support `null` or Undefined `defaultValue`
### Requirements

We want to support `null`/undefined `defaultValue` in flag configurations. See Parent issue for full details.

# Steps

- [ ] Update [JSON evaluator](https://github.com/open-feature/flagd/blob/main/core/pkg/evaluator/json.go) to return `FLAG_NOT_FOUND` for configurations missing a `defaultValue` and for which targeting doesn't resolve a variant (This is the critical behavioral change)
- [ ] Make sure OFREP and RPC handlers behave accordingly.

Rahul-Baradol:
## This PR
- adds support for null/missing default values 

### Related Issues
Fixes #1647 

### Notes
Points to be noted...
- If there is no `defaultValue` and no targetting rules, then `FlagNotFound` is returned
- If targetting doesn't resolve a variant, and there is no `defaultValue`, then `FlagNotFound` is returned


toddbaert:
Hey @Rahul-Baradol, thanks a lot!

This looks like it's on the right track. I think before we can merge it though, we need to update the schema as mentioned in [this issue](https://github.com/open-feature/flagd-schemas/issues/186), so configurations like this are seen as valid (that will be a different PR though).

Apart from that, could you add a test or two to assert that FLAG_NOT_FOUND is returned in both of the cases you mention in your description: 

> Points to be noted...
> 
> * If there is no `defaultValue` and no targetting rules, then `FlagNotFound` is returned
> * If targetting doesn't resolve a variant, and there is no `defaultValue`, then `FlagNotFound` is returned

EDIT: Oh, I see you also did https://github.com/open-feature/flagd-schemas/issues/186, great! I will review that as well.

Rahul-Baradol:
Sure will add tests!

toddbaert:
@Rahul-Baradol I will update this PR with the [new schema](https://github.com/open-feature/flagd-schemas/pull/187) once it's released, you won't have to do that, the submodule stuff can get a bit messy.

Just work on adding the test for now. It's no rush though, we might have to delay merging this one until we have some other related work done as well.

Rahul-Baradol:
Got it!

chrfwow (approved):

beeme1mr (approved):
Thanks,

toddbaert:
@Rahul-Baradol I've updated the schema submodule to https://github.com/open-feature/flagd-schemas/releases/tag/json%2Fjson-schema-v0.2.12, and I'm merging this. This means you will be able to use this feature locally in flagd. HOWEVER, because we have not yet published the new schema at the canonical location [here](https://flagd.dev/schema/v0/targeting.json), editors/validators will still see missing `defaultValue`'s as invalid. This is good for now, since this feature is not yet implemented in most in-process providers.

I've also removed the `!` from the title, since this is non-breaking.

Thanks a lot for your help!

PR #1660, opened 2025-06-30
docs(ADR): support flags with duplicate keys.

tangenti:
This is proposal for support flags with duplicate keys, as a follow up of the discussion on #1634 and #1644.

The selector semantics change proposed in #1644 will be addressed in a separate ADR.


docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+
+This ADR proposes allowing a single sync source to provide multiple flags that share the same key. This enables greater flexibility for modularizing flag configurations.
+
+## Background
+
+Currently, the `flagd` [flag configuration](https://flagd.dev/schema/v0/flags.json) stores flags in a JSON object (a map), where each key must be unique. While the JSON specification technically allows duplicate keys, it's not recommended and not well-supported in the implementations.
+
+This limitation prevents use cases for flag modularization and multi-tenancy, such as:
+
+- **Component-based Flags:** Two different services, each with its own in-process provider, cannot independently define a flag with the same key when communicating with the same `flagd` daemon.
```
toddbaert, line 18:
```suggestion
- **Component-based Flags:** Two different services, each with its own in-process provider, cannot independently define a flag with the same key in different sources when communicating with the same `flagd` daemon.
```

Nit: I think this is implied, but I think it's worth being explicit.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Proposal
+
+We will update the `flagd` flag configuration schema to support receiving flags as an **array of flag objects**. The existing schema will remain fully supported.
+
+### API Change
+
+#### Flag Configuration Schema
+
+We'll add a new schema as a [subschema](https://json-schema.org/learn/glossary#subschema) to the existing flagd flag configuration schema. It will be a composite of the original schema except `flags` (`#/definitions/base`), with a new schema for `flags` that allows flags array in addition to the currently supported flags object. The existing main schema will be the composite of the
```
toddbaert: > The existing main schema will be the composite of the

I think you've missed a word or two here.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+        },
+        {
+            "properties": {
+                "flags": {
+                    "oneOf": [
+                        {
+                            "$ref": "#/definitions/flagsArray"
+                        },
+                        {
+                            "$ref": "#/definitions/flagsMap"
```
toddbaert, line 77: I think there's multiple ways of structuring this schema for the same effective result, but what I'm getting from this is:

- any existing schema which is valid would remain valid
- the `flags` property could also be a JSON array of valid flags (an array of `#/definitions/flag`)

Is that correct?

toddbaert: Ahh, you've basically confirmed this [here](https://github.com/open-feature/flagd/pull/1660/files#diff-0e81c4866de16780a61beac4f6f4ad0088140f2eb70bdd8178f23a13eb0e3d9aR101).

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+
+1. The `flags` store will use the composite key for flags.
+
+1. `selector` will be removed from the store
+
+1. `flagSetID` will be moved from `source` metadata to `flag` metadata.
+
+### Flags Lifecycle
+
+Currently, the flags configurations from the latest update of a source will override the existing ones. If a flag was presented in the previous configuration but not in the current configuration, it will **NOT** get removed. Flags removals can only be triggerred when a source is removed, or a full resync is triggered.
```
toddbaert: > If a flag was presented in the previous configuration but not in the current configuration, it will **NOT** get removed. Flags removals can only be triggerred when a source is removed, or a full resync is triggered.

I could be misunderstanding this point, but if I'm not, this is a change from today. Today, sources are always treated as the current representation of all the flags they contain, and removals remove the flag. You can see this behavior if you start flagd from the repo in the dev startup make target:

In one terminal, do:

```sh
make run start // this uses config/samples/example_flags.flagd.json 
```

In a new terminal, do:

```sh
curl -X POST "http://localhost:8013/flagd.evaluation.v1.Service/ResolveString"       -d '{"flagKey":"myStringFlag","context":{}}' -H "Content-Type: application/json" | jq
```

And you will see: 

```json
{
  "value": "val1",
  "reason": "STATIC",
  "variant": "key1",
  "metadata": {
    "flagSetId": "example",
    "version": "v1"
  }
```

Then, delete the `myStringFlag` from the `config/samples/example_flags.flagd.json` file, save, and run the `curl` again, you will see:

```json
{
  "code": "not_found",
  "message": "Flag not found"
}
```

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+1. If a flag from a source no longer presents in the latest configuration of the same source, it will be kept.
+
+This behavior is not ideal and should be addressed in a separate ADR.
+
+### Consequences
+
+#### The good
+
+- One source can provide flags with the same keys.
+- Flag set ID no longer bound to a source, so one source can have multiple flag sets.
```
toddbaert, line 134: @beeme1mr this is perhaps more relevant to #1644, but right now when a flag evaluation fails (for instance, can't be found) we still use the `flagSetId` from the source we've selected in our telemetry (it shows up in our error metrics). We might have to change how we compute the flagSetId in telemetry for error evaluations with this change, but it doesn't seem like a deal-breaker.

toddbaert:
I see this proposal as totally workable, but in conflict with #1634 .

In my mind, the main choice we have here is between the more opinionated, less flexible hierarchical structure of #1634, which comes along with built-in assurances (such as the logical impossibility of having duplicate keys within a set) but makes "flagSetId" a special "key-of-keys" property and invalidates the previous usage of `flagSetId`, vs this proposal, which is less breaking and more flexible. With this proposal, the existing semantics of `flagSetId` don't change at all, but we now have to handle a new type of conflict (conflicts within the logical keys created by flagSetId + flagKey).

I _think_ I prefer this proposal slightly because of the non-breaking aspects.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+- No breaking change of the API definition and the API behaviors.
+- No significant change on the flagd stores and how selections work.
+
+#### The bad
+
+- The proposal still leverages the concept of flag set in the flagd storage.
+
+- The schema does not guarantee that flags of the same flag set from the same source will not have the same keys. This is guaranteed in the proposal of #1634.
+
+- The flag array is less readable compared to the flag sets object proposed in #1634.
```
toddbaert: I think this is bit subjective, but if everyone agrees it's less readable I'm OK to keep this point. It seems comparable to me.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+2. `{"flagKey": "enable-feature", "flagSetID": "default"}`
+3. `{"flagKey": "enable-feature", "flagSetID": "beta"}`
+
+### Flagd daemon
+
+Flagd daemon will perform the JSON schema checks with the reference to `https://flagd.dev/schema/v0/flags.json#flagsarray`, allowing both flags as an object and as an array.
+
+If the flag array contains two or more flags with the same composite key, the config will be considered as invalid.
+
+If the request from in-process flagd providers result in a config that has duplicate flag keys, the flagd daemon will only keep one of them in the response.
```
chrfwow, line 105: I think we could/should also change this behaviour in this ADR. If we keep one of the duplicate definitions, the service that expects the dropped flag will receive unexpected values, maybe even of unexpected type. I believe we should drop all flags that have the same composite key.

docs/architecture-decisions/duplicate-flag-keys.md
guidobrei: @toddbaert telemetry in flagd, or in the providers?

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: For the use case of our product, it will be the same source. Does this impact any of the requirements?

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Right :)

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: > If we keep one of the duplicate definitions, the service that expects the dropped flag will receive unexpected values, maybe even of unexpected type

True. I believe it's also the case for the current implementation - in-process flagd provider cannot describe the expectations for the received flags, and if the source or flagd provides a flag with the same key but different type, it will accept it.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: You're right. I had the impression that gRPC sync sources will trigger the sync type `sync.UPDATE`, and after searching the code base, it turns out all sync sources use `sync.ALL`, which will remove the flags that no longer present. 

Is `sync.UPDATE` something deprecated already?

Now it confuses me a bit - for the following scenario, it seems the flag A will be removed?

1. Source A sent a config that includes flag A
2. Source B sent a config that includes flag A as well (which will override the one from Source A).
3. Source B sent another config that doesn't include flag A.

So basically whoever offers the last state of the flag will take the ownership of that flag. If that's the case, I think we'll keep the same behavior with the composite key in this ADR.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: What I could imagine is that the source may organize the flag array like following

```
key:feature, flagSetID: primary
key:feature, flagSetID: secondary
key:another-feature, flagSetID: primary
...
```
and when the list goes longer, it's not easy to tell how many flag sets are there, and a flag set includes which flag.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Oops. Done.

docs/architecture-decisions/duplicate-flag-keys.md
toddbaert: No I guess not! Resolving.

docs/architecture-decisions/duplicate-flag-keys.md
toddbaert: > Is sync.UPDATE something deprecated already?

Yes. This is a "dead end" we want to remove. I will investigate the possibility of doing that. It's come up before as a point of confusion.

> So basically whoever offers the last state of the flag will take the ownership of that flag. If that's the case, I think we'll keep the same behavior with the composite key in this ADR.

Yes, I see no change here.

docs/architecture-decisions/duplicate-flag-keys.md
toddbaert: > @toddbaert telemetry in flagd, or in the providers?

The providers, technically, sorry!

docs/architecture-decisions/duplicate-flag-keys.md
toddbaert: Agreed on both counts.

I think we can drop these for now, and in future, we might allow some means of defining or using a conflict resolution strategy, with "drop both" (our current strategy) only being one, but I don't think it's needed right now.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+This ADR proposes allowing a single sync source to provide multiple flags that share the same key. This enables greater flexibility for modularizing flag configurations.
+
+## Background
+
+Currently, the `flagd` [flag configuration](https://flagd.dev/schema/v0/flags.json) stores flags in a JSON object (a map), where each key must be unique. While the JSON specification technically allows duplicate keys, it's not recommended and not well-supported in the implementations.
+
+This limitation prevents use cases for flag modularization and multi-tenancy, such as:
+
+- **Component-based Flags:** Two different services, each with its own in-process provider, cannot independently define a flag with the same key when communicating with the same `flagd` daemon.
+- **Multi-Tenant Targeting:** A single flagd daemon uses the same flag key with different targeting rules for different tenants
```
alexandraoberaigner:
```suggestion
- **Multi-Tenant Targeting:** A single flagd daemon cannot use the same flag key with different targeting rules for different tenants
```
NIT: to be consistent with the previous point phrasing

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+            "required": [
+                "flags"
+            ]
+        }
+    ]
+}
+...
+```
+
+If the config level flag set ID is not specified, `metadata.flagSetID` of each flag will be interpreted as its flag set ID.
```
alexandraoberaigner:
```suggestion
If the config level flag set Id is not specified, `metadata.flagSetId` of each flag will be interpreted as its flag set Id.
```
NIT: in the [current schema](https://flagd.dev/schema/v0/flags.json) it's `flagSetId`

docs/architecture-decisions/duplicate-flag-keys.md
guidobrei: If we just drop it, evaluation will result in `FLAG_NOT_FOUND` evaluation errors. Would be nice if we could report proper evaluation errors stating the reason.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+#### The good
+
+- One source can provide flags with the same keys.
+- Flag set ID no longer bound to a source, so one source can have multiple flag sets.
+- No breaking change of the API definition and the API behaviors.
+- No significant change on the flagd stores and how selections work.
+
+#### The bad
+
+- The proposal still leverages the concept of flag set in the flagd storage.
```
guidobrei, line 140: Yes, but I can see this only being a first step. 
In combination with the flexible selectors proposed in [ADR 1644](https://github.com/open-feature/flagd/pull/1644/files#diff-2d0251ee0f0d91e5d5064b192171405a39cad2dca4625762e0a6a3a5005b3139R39) we can rethink the storage layer later (e.g. split it in plain storage and indexing) in a non-breaking way.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+
+- The schema does not guarantee that flags of the same flag set from the same source will not have the same keys. This is guaranteed in the proposal of #1634.
```
guidobrei, line 142: Like JSON, also JSON schema does not define that property keys in a JSON object must be unique. The advantage of #1634 is that most parsers will immediately fail when they encounter duplicates. Whereas in this proposal it requires additional logic to detect duplicates.
But this could also be an advantage that enables more flexible selectors in the future.

docs/architecture-decisions/duplicate-flag-keys.md
guidobrei: I still see it neutral: we do not enforce how flag sources should organize their flags.

docs/architecture-decisions/duplicate-flag-keys.md
```diff
@@ -0,0 +1,144 @@
...
+### Consequences
+
+#### The good
+
+- One source can provide flags with the same keys.
+- Flag set ID no longer bound to a source, so one source can have multiple flag sets.
+- No breaking change of the API definition and the API behaviors.
+- No significant change on the flagd stores and how selections work.
+
+#### The bad
```
guidobrei, line 138: One additional downside we could add:

Compared to #1634 we cannot define flag set wide metadata that can be applied to all flags in a set.

docs/architecture-decisions/duplicate-flag-keys.md
toddbaert: I think we can report that in flagd with some loud warnings, but it's not so much a provider concern. If a back-end drops a flag, it seems like a "backend" problem.

toddbaert (approved):
After consideration and discussions internally, I think I'm in favor of accepting this over #1634 due to it's flexibility and "less-breaking" nature.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Thanks. I've updated the section to reflect this.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Thanks. Done

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Thanks. I kept `ID` and changed `flagSetID` to `flagSetId`.

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Added. Thanks

tangenti:
> After consideration and discussions internally, I think I'm in favor of accepting this over #1634 due to it's flexibility and "less-breaking" nature.

Thanks a lot Todd. As Guido pointed out in the comments, I think we could consider different options for the storage layer implementation. 


@guidobrei Could you approve the PR if there's no more concerns from you and you believe we're aligned here?

docs/architecture-decisions/duplicate-flag-keys.md
tangenti: Thanks. I've removed this point.

tangenti:
@toddbaert  It seems I don't have the permission to merge the PR. Could you merge the PR if everything looks good to you?

PR #1691, opened 2025-07-17
feat!: remove sync.Type

toddbaert:
The PR removes the obsolete `sync.Type`. This was intended to be used to support partial updates from sync sources, but it's never been used and it's lagging behind on features and would not work fully if implemented by a source (`metadata` is not properly implemented). Moreover, we're not convinced the value is worth the complexity it adds.

Specifically:

- removes `sync.Type` and all references
- removes tests for `sync.Type` and updates assertions for unrelated tests which previously asserted on `sync.Type` (now we use the payload/source in assertions)
- (unrelated) updates `CONTRIBUTING.md` to include a bunch of helpful manual testing steps

Note that this is a breaking change, but only for the `flagd/core` library, NOT for flagd or any sync source itself in terms of their behavior. Since there was no change in `flagd/`, this will not show up in the CHANGELOG.


toddbaert:
@djosephsen I hope you are relieved to see this! :sweat_smile:

aepfli (approved):

bbland1 (approved):

Issue #1690, opened 2025-07-17
[FEATURE] Specify and clarify configuration parsing for flag configurations


aepfli:
### Requirements

How we are parsing and handling flags is different from implementation to implementation. Simply based on the fact how we grew as a project and that a lot of implementations have been done by multiple contributors. Now it is the time to rethink our parsing and handling of the configurations, especially for in-process.

### Goals

- utilize schema validation as much as possible, instead of custom logic, on a flag and not on a configuration level
- define which file formats are supported (relates too https://github.com/open-feature/flagd/issues/1689)
- define what can cause a PARSE_ERROR: currently only an invalid targeting rule can cause this error, other schema invalidation result in a FLAG_NOT_FOUND
- define this in a way that we can test this via our testbed/test-harness

### issue relating to this

- https://github.com/open-feature/flagd/issues/1627
- https://github.com/open-feature/flagd/issues/1689
- https://github.com/open-feature/flagd/issues/1487
- https://github.com/open-feature/flagd/issues/1679

aepfli:
should we again create an ADR for this? or just do it?

aepfli:
this pr is on hold due to this foundings https://github.com/open-feature/python-sdk-contrib/pull/282

PR #1692, opened 2025-07-17
docs(ADR): standardize Flag configuration parsing and error handling

aepfli:
relates to: #1690

This is the starting point, i think we should talk about what we want to achieve and how. My simple mind did not manage to come to another option i could propose, but i am open to add more options. please let me know what you think of this.

### DIFF
--- a/docs/architecture-decisions/flag_configuration_parsing.md
+++ b/docs/architecture-decisions/flag_configuration_parsing.md
@@ -64,6 +64,39 @@ flowchart TD
     E@{ shape: processes, label: "each Flag" }
 ```
 
+#### Resolution flow
+
+> [!NOTE]
+> This was generated with AI and enhanced based on the java flagd provider
+
+```mermaid
+flowchart TD
+    A[Start Evaluation] --> B[Retrieve flag from store]
+    B --> B1{Is an entry in the store}
+    B1 -->|No| D[Return FLAG_NOT_FOUND with error message]
+    B1 -->|Yes| B2{Is the flag null}
+    B2 -->|Yes| B3[Return PARSE_ERROR]
+    B2 -->|No| E{Is flag state DISABLED?}
+    E -->|Yes| F[Return FLAG_NOT_FOUND with disabled error]
+    E -->|No| G{Is targeting string empty?}
+    G -->|Yes| H[Set resolvedVariant to defaultVariant and reason to STATIC]
+    G -->|No| I[Evaluate targeting rule]
+    I --> J{Targeting evaluation successful?}
+    J -->|No| K[Throw ParseError for targeting rule]
+    J -->|Yes| L{Is targeting result null?}
+    L -->|Yes| M[Set resolvedVariant to defaultVariant and reason to DEFAULT]
+    L -->|No| N[Set resolvedVariant to targeting result and reason to TARGETING_MATCH]
+    M --> O[Check variant existence]
+    N --> O[Check variant existence]
+    O --> P{Is variant found in flag?}
+    P -->|No| Q[Throw General for missing variant]
+    P -->|Yes| R[Check variant type compatibility]
+    R --> S{Is type compatible?}
+    S -->|No| T[Throw TypeMismatchError for invalid type]
+    S -->|Yes| U[Return ProviderEvaluation with resolved value, variant, and reason]
+
+```
+
 [Double click to switch code/render]
 
 ### API Changes
--- a/docs/architecture-decisions/flag_configuration_parsing.md
+++ b/docs/architecture-decisions/flag_configuration_parsing.md
@@ -1,0 +1,77 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: Simon Schrottner 
+created: 2025-07-17
+updated: YYYY-MM-DD
+---
+
+# Proposal for Unified Flag Parsing and Handling
+
+The goal of this proposal is to standardize how flags are parsed and handled across implementations. Currently, there is significant variation in how this is done, stemming from the organic growth of the project and contributions from multiple developers. This proposal aims to rethink and unify the parsing and handling of configurations, particularly for in-process scenarios.
+
+## Background
+
+Over time, the project has grown with contributions from various developers, leading to inconsistencies in how flags are parsed and handled. These inconsistencies have resulted in bugs, unexpected behaviors, and challenges in maintaining the codebase. For example, the file provider and JSON parsing behave differently across languages, and there is no clear definition of what constitutes a parsing error versus a flag-not-found error.
+
+Several issues have been raised that highlight the need for a unified approach:
+
+- [BUG] Inconsistency in file provider and JSON parsing for different languages (#1627)
+- [BUG] [File Provider] Support File Formats (#1689)
+- [FEATURE] Provider option to reject flag set with invalid rules (#1487)
+
+To address these issues, it is essential to establish a consistent and testable framework for flag parsing and handling.
+
+## Requirements
+
+- Utilize schema validation as much as possible, instead of custom logic, on a flag level rather than a configuration level.
+- Define what consistutes a `PARSE_ERROR` versus a `FLAG_NOT_FOUND`.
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,78 @@
...
+- [BUG] Inconsistency in file provider and JSON parsing for different languages (#1627)
+- [BUG] [File Provider] Support File Formats (#1689)
+- [FEATURE] Provider option to reject flag set with invalid rules (#1487)
+
+To address these issues, it is essential to establish a consistent and testable framework for flag parsing and handling.
+
+## Requirements
+
+* Utilize schema validation as much as possible, instead of custom logic, on a flag level rather than a configuration level.
+* Define what consistutes a `PARSE_ERROR` versus a `FLAG_NOT_FOUND`.
```
toddbaert:
```suggestion
* Define what constitutes a `PARSE_ERROR` versus what prevents the flag set from being updated.
```

docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,78 @@
...
+updated: YYYY-MM-DD
+---
+
+# Proposal for Unified Flag Parsing and Handling
+
+The goal of this proposal is to standardize how flags are parsed and handled across implementations. Currently, there is significant variation in how this is done, stemming from the organic growth of the project and contributions from multiple developers. This proposal aims to rethink and unify the parsing and handling of configurations, particularly for in-process scenarios.
+
+## Background
+
+Over time, the project has grown with contributions from various developers, leading to inconsistencies in how flags are parsed and handled. These inconsistencies have resulted in bugs, unexpected behaviors, and challenges in maintaining the codebase. For example, the file provider and JSON parsing behave differently across languages, and there is no clear definition of what constitutes a parsing error versus a flag-not-found error.
```
toddbaert:
```suggestion
Initially we did not have a JSON schema, and validation was done manually after the configuration was parsed; in this case parsing was not consistent at all.
Later, we implemented a basic schema for the flag structure, which could be used to consistently validate the everything but the targeting rules.
Finally, we wrote a full schema for the targeting rules as well, which ensure that all `targeting` was valid.
These inconsistencies have resulted in bugs, divergent behaviors, and challenges in maintaining the codebase. For example, the file provider and JSON parsing behave differently across languages, and there is no clear definition of what constitutes a parsing error versus a flag-not-found error, and what exactly happens when an invalid configuration is loaded.
```

docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,78 @@
...
+- [BUG] [File Provider] Support File Formats (#1689)
+- [FEATURE] Provider option to reject flag set with invalid rules (#1487)
+
+To address these issues, it is essential to establish a consistent and testable framework for flag parsing and handling.
+
+## Requirements
+
+* Utilize schema validation as much as possible, instead of custom logic, on a flag level rather than a configuration level.
+* Define what consistutes a `PARSE_ERROR` versus a `FLAG_NOT_FOUND`.
+* Define if an invalid flag entry within a configuration can invalidate the whole configuration.
```
toddbaert: I think a JSON document is either invalid or valid, so we don't have unlimited flexibility here unless we want to do some fairly annoying manipulation of the payload, either validating only 1 flag at a time (this might require breaking single flags into a subschema).

It _might_ be possible to identify the specific sub-entity of an invalid configuration with the validation library, and remove it, but I think that might not be worth it or possible everywhere.

docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,78 @@
...
+* Define which file formats are supported (e.g., JSON, YAML, etc.).
```
toddbaert: We should make this consistent, I think. It's pretty trivial to support YAML/JSON conversion in most cases, though there might be some edge cases to consider.

docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,78 @@
...
+* Ensure the framework is testable via the existing testbed/test-harness.
```
toddbaert: :100:

docs/architecture-decisions/flag_configuration_parsing.md
aepfli: we can select subsets for validating entities within the json schema, it will be one validation per flag, but i think it is better than invalidating the whole configuration.

docs/architecture-decisions/flag_configuration_parsing.md
```diff
@@ -0,0 +1,77 @@
...
+## Proposal
+
+### Standardize Flag Parsing and Handling Using Schema Validation (Option 1)
+
+This proposal recommends adopting schema validation as the primary method for parsing and handling flags. The key elements of this approach include:
+
+1. **Schema Validation**: Use schema validation to ensure that flags meet predefined criteria. This reduces the reliance on custom logic and improves consistency across implementations.
+2. **Supported File Formats**: Define and document the file formats that are supported (e.g., JSON, YAML). This aligns with the issue raised in [BUG] [File Provider] Support File Formats (#1689).
+3. **Error Definitions**: Clearly define what constitutes a `PARSE_ERROR` versus a `FLAG_NOT_FOUND` error. Each inconsistent flag should evaluate as a `PARSE_ERROR`, only if the flag is not part of the configuration (or due to other definitions), it should evaluate to a `FLAG_NOT_FOUND`
+4. **Invalid Flag Handling**: A invalid flag configuration should not affect the whole configuration. Other valid flags might still be important, and an error on one flag should not hinder updates or changes to the flag configuration.
```
toddbaert, line 51: I have to admit, I have a strong preference for failing to start up if the config is invalid in any way, and fully rejecting an update if it's invalid in any way. I think this is what most users expect from invalid configurations. I would be open to a switch that would ignore validation errors, in which case the behavior might not be well-defined (but it would allow us to test some of our edge cases).

I think removing invalid flags actually leads to more confusing behaviors, since flagd will start up but certain flags might just be missing due to a typo. I think it's better to simply not start up if any single flag is invalid. When an update is received, I think similarly - it's better if the entire update fails than a single flag is "silently" removed (only showing up in a log or something).

I also think implementing removal of only invalid flags will be tricky.

I'm really interested in other's opinions here though. Maybe I'm in the minority here.

aepfli: I would not remove them, I would persist them with a parse_error in the flag storage. this is now implementation specific for eg a hashmap:

- an existing key without a value is a parse_error reason
-  an existing key with a flag, will trigger the resolution process
- a non existing key, is a flag not found

This way, we keep the user informed about possible flag issues, but ensure that updates for other flags are working.

aepfli: enhanced the proposal with flow diagrams for easier understanding of the proposed flow

toddbaert:
@aepfli sorry, I started typing without selecting the textbox first and accidently closed the issue :sweat_smile: 

To summarize my alternative proposal:

- use https://flagd.dev/schema/v0/flags.json to validate the whole document, on each sync
- at startup, if anything is invalid, flagd fails to start (this is already partially implemented; if we can't parse the flags configuration, so we'd just follow that same path)
- on any sync update, if anything is invalid, the update is ignored and an error/warning message is logged louded (again, this is already partially implemented; if we can't parse the flags configuration, so we'd just follow that same path)

Advantages:

- very easy to implement, just add a validation step in the exact same code-location we already parse
- we can remove many of our "error-edge-cases" in our test suite, since they will never be allowed
- we rely fully on a schema for validation so we don't have to implement any detailed error handling code

Disadvantages:

- 1 bad flag will prevent a whole update - however, this is already true if a single flag cannot be parsed (invalid JSON or nonsense variant configuration, etc) so we can never fully remove this situation, so I think we should just embrace this flow

aepfli:
@toddbaert I also added your proposal.

I think it would be time to start some kind of vote - or give people the chance to add their remark - @open-feature/flagd-triagers please take a look

Issue #1688, opened 2025-07-17
[FEATURE] Temporary Context Enrichment toggle


aepfli:
### Requirements

With https://github.com/open-feature/flagd/issues/1635 we added the context enrichment payload into our sync payload. Which is a great way to reduce complexity. But we are currently lacking a way to test this during implementations of our providers. Currently the info is sent via both "channels" (getMetadata & Sync). 

This brings us to a problematic situation while implementing the new approach in providers. We can not test this properly because the `getMetadata` request is sent first.

I propose to add a temporary config flag in flagd which decides which path to use. This allows us to define clear strategies within gherkin for testing.

Furthermore it allows us to define a specified migration and deprecation plan:
1. we only use getMetadata by default and Sync payload needs to be activated
2. we use Sync by default and getMetadata as a fallback
3. we remove getMetadata after a grace period

Additionally we could think of propagating this information/flag also via the flag configuration as additional information. (to be discusses)

leakonvalinka:
After a discussion with @aepfli and @guidobrei we figured it makes more sense to add a toggle that only en- or disables the `getMetadata` request and always send the `syncContext` field. This way, future migration away from the `getMetadata` request is possible.

PR #1693, opened 2025-07-23
feat: Add toggle for disabling getMetadata request

leakonvalinka:
## This PR

- adds a toggle `disable-sync-metadata` for the Sync Service to disable or enable the deprecated `getMetadata` request

### Related Issues
[#1688 [FEATURE] Temporary Context Enrichment toggle](https://github.com/open-feature/flagd/issues/1688)


flagd/cmd/start.go
```diff
@@ -89,6 +90,7 @@ func init() {
 	flags.StringToStringP(headerToContextKeyFlagName, "H", map[string]string{}, "add key-value pairs to map "+
 		"header values to context values, where key is Header name, value is context key")
 	flags.Duration(streamDeadlineFlagName, 0, "Set a server-side deadline for flagd sync and event streams (default 0, means no deadline).")
+	flags.Bool(enableSyncContext, false, "Enable or disable sync context. Defaults to false.")
```
toddbaert:
```suggestion
	flags.Bool(enableSyncContext, false, "Enable or disables including static context values (see the -X flag) in the sync payload, instead of in the GetMetadataResponse. Defaults to false, but will be default to true in later versions.")
```

docs/reference/flagd-cli/flagd_start.md
```diff
@@ -14,6 +14,7 @@ flagd start [flags]
   -H, --context-from-header stringToString   add key-value pairs to map header values to context values, where key is Header name, value is context key (default [])
   -X, --context-value stringToString         add arbitrary key value pairs to the flag evaluation context (default [])
   -C, --cors-origin strings                  CORS allowed origins, * will allow all origins
+      --enable-sync-context                  Enable or disable sync context. Defaults to false.
```
toddbaert:
```suggestion
      --enable-sync-context                  Enable or disables including static context values (see the -X flag) in the sync payload, instead of in the GetMetadataResponse. Defaults to false, but will be default to true in later versions.
```

flagd/pkg/service/flag-sync/handler.go
```diff
@@ -97,6 +99,9 @@ func (s syncHandler) FetchAllFlags(_ context.Context, req *syncv1.FetchAllFlagsR
 func (s syncHandler) GetMetadata(_ context.Context, _ *syncv1.GetMetadataRequest) (
 	*syncv1.GetMetadataResponse, error,
 ) {
+	if s.enableSyncContext {
+		return nil, fmt.Errorf("metadata endpoint disabled")
```
toddbaert: Can we use a "proper" error code for this? Something from [here](https://grpc.io/docs/guides/status-codes/) should work; maybe "UNIMPLEMENTED"?.

I think you can return it like we do here: https://github.com/open-feature/flagd/blob/ea3be4f9010644132795bb60b36fb7705f901b62/flagd/pkg/service/flag-sync/handler.go#L74

As written, we get:

```
ERROR:
  Code: Unknown
  Message: metadata endpoint disabled
```

Which is good, but not ideal since the it's not a helpful error code.

toddbaert:
Looks like you're heading in the right direction!

There's an e2e test that needs updating because of the submodule update, and I left some text suggestions and a suggestion about using an informative error code.

flagd/pkg/service/flag-sync/handler.go
leakonvalinka: You're right, I added this with the UNIMPLEMENTED error code

aepfli (approved):

PR #1697, opened 2025-07-28
chore(refactor): use memdb for flag storage

toddbaert:
This PR contains no behavioral changes, and no breaking changes.

It lays the groundwork for some of the upcoming improvements we want in flagd, as specified in recent ADRs. It does this by re-implementing our storage layer to use [go-memdb](https://github.com/hashicorp/go-memdb), an open source in-memory database developed by Hashicorp and used in [Consul](https://developer.hashicorp.com/consul) (as well as MANY other projects).

### Why?

This will allow us to easily support:

- duplicate flag keys (namespaced by flagSetId), as mentioned in [this ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/duplicate-flag-keys.md)
- robust flag selectors, as mentioned in [this ADR](https://github.com/open-feature/flagd/pull/1644), by supporting "watchers" which allow us to "listen" to change in flags returned by a given query :man_dancing: 
- robust (and possibly, in the future, even customizable) indexing on arbitrary attributes (to easily support fetching "all boolean flags", or "flags with metadata = xyz", etc)
- cross-"table" transactions

I have already PoC'd that these are all practical.

### Changes in implementation

- the `store` package's `State` is no longer just a struct; it's a object with methods wrapping the internal db 
- the `store` package's `State` was renamed to `Store` and the file was renamed from `flags.go` to `store.go`, since it's ceased to be a simple stateful object, and for consistency
- a non-serialized (used only internally) `Key` field was added to the flag type (for indexing)
- a new constructor for the `Store` was added which takes a logger, the old was deprecated to avoid breaking changes in consumers (the Go flagd provider, mostly)

Note that the go-memdb dependency is MPL2, which is not allowed by the CNCF, however, go-memdb is already used in CNCH projects and has a [special exception](https://github.com/cncf/foundation/blob/347a55dc0a6c6dc3d55a9a782e5701080a8ec43b/license-exceptions/cncf-exceptions-2023-08-31.json#L7-L11).


Makefile
```diff
@@ -47,8 +47,11 @@ test-flagd:
 	go test -race -covermode=atomic -cover -short ./flagd/pkg/... -coverprofile=flagd-coverage.out
 test-flagd-proxy:
 	go test -race -covermode=atomic -cover -short ./flagd-proxy/pkg/... -coverprofile=flagd-proxy-coverage.out
-flagd-integration-test: # dependent on ./bin/flagd start -f file:test-harness/flags/testing-flags.json -f file:test-harness/flags/custom-ops.json -f file:test-harness/flags/evaluator-refs.json -f file:test-harness/flags/zero-flags.json
-	go test -cover ./test/integration $(ARGS)
+flagd-integration-test-harness:
```
toddbaert, line 54: Just added to make local integration testing easier.

core/pkg/evaluator/fractional_test.go
```diff
@@ -459,7 +459,7 @@ func TestFractionalEvaluation(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			log := logger.NewLogger(nil, false)
 			je := NewJSON(log, store.NewFlags())
-			je.store.Flags = tt.flags.Flags
+			je.store.Update("", "", tt.flags.Flags, model.Metadata{})
```
toddbaert, line 462: Many tests have been updated to not directly mutate the `State`.

core/pkg/model/flag.go
```diff
@@ -3,6 +3,7 @@ package model
 import "encoding/json"
 
 type Flag struct {
+	Key            string          `json:"-"`
```
toddbaert, line 6: New field used for indexing (not serialized).

Any other fields we add here (copies of certain metadata, flag type, etc, etc) could be used for indexing.

core/pkg/store/store_test.go
```diff
@@ -0,0 +1,207 @@
+package store
```
toddbaert, line 1: Unfortunately, the large amount of changes here, plus the file rename from `flags_test.go` to `store_test.go` makes this look like a re-write - bot nothing has fundamentally changed - we are just using the public methods now instead of messing with the `State` struct directly.

core/pkg/store/store.go
```diff
@@ -50,49 +52,76 @@
...
 }
 
-func (f *State) Set(key string, flag model.Flag) {
-	f.mx.Lock()
-	defer f.mx.Unlock()
-	f.Flags[key] = flag
+// Deprecated: use NewStore instead
+func NewFlags() *Store {
+	state, _ := NewStore(logger.NewLogger(nil, false))
+	return state
```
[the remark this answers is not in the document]
toddbaert: This was intentional since we have to maintain backwards compatibility with the go-flagd-provider which uses the old errorless constructor... I'm not sure panicking is really better here, but I guess it doesn't really matter since this is deprecated anyway. Once this is published we can update the provider so this won't matter for long.

flagd/pkg/service/flag-sync/util_test.go
```diff
@@ -6,43 +6,42 @@ import (
...
 // getSimpleFlagStore returns a flag store pre-filled with flags from sources A & B & C, which C empty
-func getSimpleFlagStore() (*store.State, []string) {
+func getSimpleFlagStore() (*store.Store, []string) {
 	variants := map[string]any{
 		"true":  true,
 		"false": false,
 	}
 
-	flagStore := store.NewFlags()
+	flagStore, _ := store.NewStore(logger.NewLogger(nil, false))
```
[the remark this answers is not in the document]
toddbaert: I can pass `testing` in.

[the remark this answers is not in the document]
toddbaert: I've fixed this by passing `t testing.T` in, and failing the test if there's an error here.

core/pkg/store/store.go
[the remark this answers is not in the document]
toddbaert: I've added the panic as suggested.

beeme1mr:
gemini /review

core/pkg/store/store.go
```diff
@@ -50,49 +52,79 @@ func (f *State) hasPriority(stored string, new string) bool {
...
+		Tables: map[string]*memdb.TableSchema{
+			"flags": {
+				Name: "flags",
+				Indexes: map[string]*memdb.IndexSchema{
+
+					"id": {
+						Name:    "id",
+						Unique:  true,
+						Indexer: &memdb.StringFieldIndex{Field: "Key", Lowercase: false},
+					},
```
toddbaert, line 67: Currently we just have a simple index on `Key`, to mimic the previous map behavior, but `go-memdb` supports all kinds of index types very useful to us, including compound indexes, numeric indexes, conditional indexes, etc. We can use these to great effect.

Note that `Lowercase` converts the field to lower before building the index - I've disabled this to maintain the previous, case-sensitive behavior, and for consistency with JSON map index behavior.

aepfli (approved):
Thank you, this looks really cool, and i cant wait to use it. I think it will reduce our internal efforts heavily, as the memdb will offer a lot of flexibility. Thank you

core/pkg/store/store.go
```diff
@@ -50,49 +52,79 @@ func (f *State) hasPriority(stored string, new string) bool {
...
+	if err != nil {
+		panic(fmt.Sprintf("unable to create flag store: %v", err))
+	}
+	return state
 }
 
-func (f *State) Get(_ context.Context, key string) (model.Flag, model.Metadata, bool) {
+func (f *Store) Get(_ context.Context, key string) (model.Flag, model.Metadata, bool) {
+	f.logger.Debug(fmt.Sprintf("getting flag %s", key))
 	f.mx.RLock()
```
tangenti: We could move the lock operations to `GetMetadataForSource`.

core/pkg/store/store.go
```diff
@@ -101,95 +133,80 @@ func (f *State) String() (string, error) {
 }
 
 // GetAll returns a copy of the store's state (copy in order to be concurrency safe)
-func (f *State) GetAll(_ context.Context) (map[string]model.Flag, model.Metadata, error) {
+func (f *Store) GetAll(_ context.Context) (map[string]model.Flag, model.Metadata, error) {
 	f.mx.RLock()
```
tangenti: Same here.

tangenti (approved):

flagd/pkg/service/flag-sync/util_test.go
```diff
@@ -5,44 +5,47 @@ import (
...
+	"github.com/open-feature/flagd/core/pkg/logger"
 	"github.com/open-feature/flagd/core/pkg/model"
 	"github.com/open-feature/flagd/core/pkg/store"
 	"google.golang.org/grpc/credentials"
 )
 
-// getSimpleFlagStore returns a flag store pre-filled with flags from sources A & B & C, which C empty
-func getSimpleFlagStore() (*store.State, []string) {
+// getSimpleFlagStore is a test util which returns a flag store pre-filled with flags from sources A & B & C, which C empty
+func getSimpleFlagStore(t *testing.T) (*store.Store, []string) {
```
erka: You could use `testing.TB` instead of `*testing.T` and it would be nice to add `t.Helper()` call

core/pkg/store/store.go
toddbaert: @tangenti You are correct! I will make this change; we also the RLock need it on `getMetadata` (a similar function needed at least for now, to fill metadata in error situations). I want to make improvements around these (and perhaps even use the DB to store metadata) but I want to keep this PR limited.

I will implement your suggestion.

flagd/pkg/service/flag-sync/util_test.go
toddbaert: TIL about [t.Helper](https://pkg.go.dev/testing#B.Helper). I will add that, as well as use `TB`.

core/pkg/store/store.go
toddbaert: Done

core/pkg/store/store.go
toddbaert: https://github.com/open-feature/flagd/pull/1697#discussion_r2240427359

flagd/pkg/service/flag-sync/util_test.go
toddbaert: Done.

core/pkg/store/store.go
```diff
@@ -50,49 +52,77 @@ func (f *State) hasPriority(stored string, new string) bool {
...
 
-func (f *State) Delete(key string) {
-	f.mx.Lock()
-	defer f.mx.Unlock()
-	delete(f.Flags, key)
-}
-
-func (f *State) String() (string, error) {
+func (f *Store) String() (string, error) {
+	f.logger.Debug(fmt.Sprintf("dumping flags to string"))
```
beeme1mr: Can you address the static check error, please?

beeme1mr (approved):

core/pkg/store/store.go
toddbaert: Yep, done.

Issue #1700, opened 2025-07-30
[BUG] Race condition in test suite


toddbaert:
It looks [like this change](https://github.com/open-feature/flagd/pull/1697) created a very subtle (doesn't show up on most test runs) [race condition](https://github.com/open-feature/flagd/actions/runs/16623441044/job/47033372823) in tests.

I think [upcoming changes](https://github.com/open-feature/flagd/issues/1675) in how we handle subscriptions (using the new DB) will fix this indirectly, but I'm making a note of it here.

Issue #1695, opened 2025-07-23
[REFACTOR] Migrate storage to go-memdb


toddbaert:
In order to support flag arrays, selection/querying of flags based on flagSetId and other properties, we can migrate to [GitHub - hashicorp/go-memdb: Golang in-memory database built on immutable radix trees](https://github.com/hashicorp/go-memdb) 

This supports in-memory storage and index-based queries, which will allow us to more gracefully handle flags with conflicting keys across sets, and querying/evaluation based on flagSetId.

It also features “watchers” which enable “listening” for changes within sets of flags returned by arbitrary queries. 

In this task, we’ll only be migrating to the mem-db, without adding any new features.

Issue #1675, opened 2025-07-10
[FEATURE] Support `selector` Expression to Retrieve All Flags Based on `flagSetId`


toddbaert:
Up to know, flagd has not handled flags with conflicting keys from different sources, or even the same source.

As described in [these](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/duplicate-flag-keys.md) 2 [documents](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/decouple-flag-source-and-set.md), we want to support a sort of "flag namespacing" via our existing `flagSetId` property. Logically, the unique identifier for a flag would be represented by a combination of `flagSetId` and `key`, irrespective of the sync-source.

Definition of done:

- flags are uniquely identified by `flagSetId` and `key` in all endpoints/protocols (evaluation.proto, sync.proto, OFREP); clients identify their flagSet (either at steam initialization or during the RPC, depending on the ptotocol) and then get responses according to that flagSet.
- flag metadata and flag set metadata are restricted similarly to above
- the means of expressing the `flagSetId` must be extensible, so that other criteria and filters might be supported later

Issue #1611, opened 2025-03-28
[FEATURE] selector param/header for OFREP


toddbaert:
flagd supports selectors, so separate flags into different buckets, eg. each file source, http source, etc, is its own selector.
If flagd is started with multiple sources, the `selector` param can be used to evaluate/sync only the set of flags from the source in question:

<details>
  <summary>Example</summary>

[a.json](https://github.com/user-attachments/files/19340956/a.json) and 
[b.json](https://github.com/user-attachments/files/19340957/b.json):
```json
{
    "flags": {
        "flagb": {
            "state": "ENABLED",
            "variants": {
                "foo": "foo",
                "bar": "bar"
            },
            "defaultVariant": "foo"
        }
    },
    "metadata": {
        "flagSetId": "b"
    }
}
```
```json
{
    "flags": {
        "flaga": {
            "state": "ENABLED",
            "variants": {
                "foo": "foo",
                "bar": "bar"
            },
            "defaultVariant": "foo"
        }
    },
    "metadata": {
        "flagSetId": "a"
    }
}
```

start flagd with those two files as sources:

```shell
docker run \
  --rm -it \
  --name flagd \
  -p 8015:8015 \
  -v $(pwd):/flags \
  ghcr.io/open-feature/flagd:latest start \
  --uri file:./flags/a.json \
  --uri file:./flags/b.json
```
 then, you can use: 

```shell
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/flaga'
```
and 

```shell
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags/flagb'
```

and 

```shell
curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags'
```

</details>

We want to support a "selector" field similar to what's in the sync protocol, as a header, which would restrict evaluations to only the flag set identified by the header. ie: 

```shell
curl -H 'X-Flag-Set-Id:a' -X POST 'http://localhost:8016/ofrep/v1/evaluate/flagb' // not found since flagb is not in set a.
```

Note, this requires https://github.com/open-feature/flagd/issues/1610 to be done first.

The precise name of the header is subject to discussion, but feel free to make a proposal or use what I've suggested above.

guidobrei:
How would this be tied up on the SDK side using an OFREP provider? Can custom headers be configured on the provider side?

toddbaert:
[...]

PR #1702, opened 2025-08-05
feat: multi-project support via selectors and flagSetId namespacing

toddbaert:
Sorry for how big this PR seems (a lot of the change lines are not significant (DB schema or tests), so I've tried to highlight the important parts. This is a substantial PR that builds on our recent changes to use go-memdb for storage. It primarily supports 2 new crucial features:

- support for duplicate flag keys from multiple sources (namespaced by `flagSetId`) as described in [this ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/duplicate-flag-keys.md)
- support for a robust query syntax in the "selector" fields and headers, as proposed by @tangenti in [this ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/decouple-flag-source-and-set.md)

Both of these were accomplished using the new go-memdb module. **The built-in "watcher" functionality also allows us to _completely delete_ our "mux" layer, which was responsible for fanning out changes from sync-sources to listeners**; this is something the go-memdb module providers for free (the ability to watch a query for changes). Now, we completely rely on this feature for all change notifications. Additionally, unlike before, change notifications are now scoped to particular _selectors_ (ie: if a client is only interested in changes for flags from `flagSetId: x` or `source: y`, they will only get change notifications pertaining to that selection. Currently, the only supported query fields for the `selector` are `"source"` and `"flagSetId"`, but this functionality can easily be extended. By default, if no `selector` is specified, the previous key-overwrite by source priority apples (this logic has also been simplified using the new database). Most of the new functionality is tested [here](https://github.com/open-feature/flagd/pull/1702/files#diff-0cdd4fe716f4b8a94466279fd1b11187fcf4d74e6434727c33d57ed78c89fe27R163-R478).

Selector in action for `Sync` API:

demo

Selector in action for `Evaluation` API:

demo2

To test, run the new make target `make run-flagd-selector-demo`, then use the OFREP or gRPC endpoints to experiment. This new functionality is available on all APIs and endpoints. The command I ran in the gifs above are:

streaming:
```shell
grpcurl -d '{"selector":"flagSetId=example"}' -import-path schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags | jq
grpcurl -d '{"selector":"flagSetId=example,source=../config/samples/example_flags.flagd.json"}' -import-path schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags | jq
grpcurl -d '{"selector":"flagSetId=other"}' -import-path schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags | jq
```

single-evaluations:
```shell
curl -X POST  -d '{"context":{}}' 'http://localhost:8016/ofrep/v1/evaluate/flags' | jq
curl -X POST -H 'flagd-selector:flagSetId=other'  -d '{"context":{}}' 'http://localhost:8016/ofrep/v1/evaluate/flags' | jq
```

:warning: There's no breaking changes here. Besides the new features, there is one behavioral change - the top level "metadata" object returned for bulk evaluations (and failed evaluations) was previously very nonsensical in it's behavior (we basically just aggregated the metadata from all sources, discarding duplicates, and sent it back. This field was used by providers for telemetry purposes. Now, since flag evaluations and subscripts are "query based" and can aggregate data from multiple sources, we've opted to simply reflect the selector queries contents here.

So if you used a selector like `"flagSetId=1234,source=../my/source"`, the top-level metadata object in the response would be:

```
"metadata": {
  "flagSetId": 1234,
  "source": "../my/source"
}
```

This is useful for the provider's ability to report errors, etc in telemetry.

Fixes: https://github.com/open-feature/flagd/issues/1675
Fixes: https://github.com/open-feature/flagd/issues/1695
Fixes: https://github.com/open-feature/flagd/issues/1611
Fixes: https://github.com/open-feature/flagd/issues/1700
Fixes: https://github.com/open-feature/flagd/issues/1610

### DIFF
--- a/core/pkg/evaluator/json.go
+++ b/core/pkg/evaluator/json.go
@@ -153,8 +153,6 @@ func (je *Resolver) ResolveAllValues(ctx context.Context, reqID string, context 
 	s := ctx.Value(store.SelectorContextKey{})
 	if s != nil {
 		selector = s.(store.Selector)
-	} else {
-		selector = store.NewSelector("")
 	}
 	allFlags, flagSetMetadata, err := je.store.GetAll(ctx, &selector)
 	if err != nil {


Makefile
```diff
@@ -47,8 +47,6 @@ test-flagd:
 	go test -race -covermode=atomic -cover -short ./flagd/pkg/... -coverprofile=flagd-coverage.out
 test-flagd-proxy:
 	go test -race -covermode=atomic -cover -short ./flagd-proxy/pkg/... -coverprofile=flagd-proxy-coverage.out
-flagd-integration-test: # dependent on ./bin/flagd start -f file:test-harness/flags/testing-flags.json -f file:test-harness/flags/custom-ops.json -f file:test-harness/flags/evaluator-refs.json -f file:test-harness/flags/zero-flags.json
```
toddbaert, line 50: This was a duped recipe.

core/pkg/evaluator/json.go
```diff
@@ -150,8 +150,12 @@ func (je *Resolver) ResolveAllValues(ctx context.Context, reqID string, context
...
 	defer span.End()
 
-	var err error
-	allFlags, flagSetMetadata, err := je.store.GetAll(ctx)
+	selector := store.NewSelector("")
+	s := ctx.Value(store.SelectorContextKey{})
+	if s != nil {
+		selector = s.(store.Selector)
+	}
+	allFlags, flagSetMetadata, err := je.store.GetAll(ctx, &selector, nil)
```
toddbaert: The `IStore` interface (and implementation) is also used by the in-process provider in the go-sdk-contribs. The in-process provider will not need to support selectors/quering as flagd does, so the interface was not modified. The selector is instead propagated in a special field in the `context.Context`, and if it's set, used in the storage layer.

Please note that we can't make breaking changes to interfaces used by the go-provider (without updating and releasing it) as it's purposely used for integration testing in this repo (ie: breaking any contracts the go in-process provider uses will cause the integration test runs here to fail).

flagd/pkg/service/flag-sync/sync-multiplexer.go
```diff
@@ -1,213 +0,0 @@
-package sync
```
toddbaert, line 1: The entire mux layer was deleted! This is now handled by the mem-db as mentioned.

core/pkg/store/store.go
```diff
@@ -76,48 +122,76 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
+			}
+			if nextFlag.Priority >= flag.Priority {
+				flag = nextFlag
+			} else {
+				s.logger.Debug(fmt.Sprintf("discarding flag %s from lower priority source %s in favor of flag from source %s", nextFlag.Key, s.sources[nextFlag.Priority], s.sources[flag.Priority]))
+			}
 
-	return f.SourceDetails[flag.Source].Selector
+		}
+		return flag, queryMeta, true
```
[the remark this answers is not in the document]
toddbaert: fixed!

core/pkg/utils/selector_to_map.go
```diff
@@ -0,0 +1,21 @@
...
+// 		// Split each pair by the first equal sign
+// 		parts := strings.Split(pair, "=")
+// 		if len(parts) == 2 {
+// 			key := parts[0]
+// 			value := parts[1]
+// 			selectorMap[key] = value
+// 		}
+// 	}
+// 	return selectorMap
+// }
```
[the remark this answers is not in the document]
toddbaert: Yes this was moved into the `Selector`. Thanks!

flagd/pkg/runtime/from_config.go
```diff
@@ -79,32 +79,33 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
...
-		s.FlagSources = append(s.FlagSources, provider.URI)
-		s.SourceDetails[provider.URI] = store.SourceDetails{
-			Source:   provider.URI,
-			Selector: provider.Selector,
-		}
+		// s.FlagSources = append(s.FlagSources, provider.URI)
+		// s.SourceDetails[provider.URI] = store.SourceDetails{
+		// 	Source:   provider.URI,
+		// 	Selector: provider.Selector,
+		// }
```
toddbaert:
```suggestion
```

core/pkg/evaluator/json_test.go
```diff
@@ -476,62 +476,6 @@ func TestResolveAllValues(t *testing.T) {
 	}
 }
 
-func TestMetadataResolveType(t *testing.T) {
```
toddbaert, line 479: I mentioned the reason for deleting this in the description. This is covered [here](https://github.com/open-feature/flagd/pull/1702/files#diff-0cdd4fe716f4b8a94466279fd1b11187fcf4d74e6434727c33d57ed78c89fe27R452) now.

flagd/pkg/service/constants.go
```diff
@@ -0,0 +1,3 @@
+package service
+
+const FLAGD_SELECTOR_HEADER = "Flagd-Selector"
```
toddbaert, line 3: This is used for the selector in OFREP and connect (gRPC over HTTP/1.1). Apparently `X-` prefixes are not to be used for custom headers anymore.

Let me know if you don't like this header name.

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
+					Flags: bytes,
+				}
+
+			case <-ctx.Done():
+				close(watcher)
+			}
+		}()
+	}
+
+	return flags, queryMeta, nil
```
toddbaert: This enables the new "Watcher" functionality. Note that the `WatchCh()` returns a "one-time", channel, immediately closed after one message, which is fired if anything impacting the associated query was changed, and the channel doesn't return any data - it's merely used as a signal to call the query again to see the changes, which we do with the recursive call in the new goroutine.

core/pkg/store/store.go
```diff
@@ -4,66 +4,117 @@ import (
...
-		}
-	}
-	return true
-}
-
-func NewStore(logger *logger.Logger) (*Store, error) {
+func NewStore(logger *logger.Logger, sources []string) (*Store, error) {
 
+	// a unique index must exist for each set of constraints - for example, to look up by key and source, we need a compound index on key+source, etc
+	// we maybe want to generate these dynamically in the future to support more robust querying, but for now we will hardcode the ones we need
```
toddbaert, line 51: As the comment here mentions, we need a distinct index for every possible combination of constraint arguments. If we move beyond `selector` and `source`, we will probably need to generate these dynamically, but these are all that's needed for now.

core/pkg/utils/build_notifications.go
```diff
@@ -0,0 +1,46 @@
+package utils
```
toddbaert: This was extracted from the store, since we also need the same logic [here](https://github.com/open-feature/flagd/pull/1702/files#diff-23c9326d5b7c4e77c939963ea1010025ca5faf16eb9db96b348d7c19cb921fbeR41) now, to diff flag changes on the RPC event stream.

core/pkg/store/store.go
```diff
@@ -4,66 +4,117 @@ import (
...
-						Unique:  true,
-						Indexer: &memdb.StringFieldIndex{Field: "Key", Lowercase: false},
+					// primary index; must be unique and named "id"
+					idIndex: {
+						Name:   idIndex,
+						Unique: true,
+						Indexer: &memdb.CompoundIndex{
+							Indexes: []memdb.Indexer{
+								&memdb.StringFieldIndex{Field: model.FlagSetId, Lowercase: false},
+								&memdb.StringFieldIndex{Field: model.Key, Lowercase: false},
```
toddbaert, line 64: :mega: This is a significant change, worth noting. The new primary index for our store is flagdSetId+key. This index is unique and is fundamentally what decouples flag sources from logical flag sets. We don't reject changes that overwrite this key, so users should understand that no matter what source a flag comes from, it's fundamental "id" is a combination of the `flagSetId` and `key`. Note that, the nil `flagSetId` is assigned to every flag without a flagSetId, though this `flagSetId` cannot be specifically selected for (this can be changed, but I'm not sure it's needed).

core/pkg/store/query.go
```diff
@@ -0,0 +1,122 @@
...
+const keySourceCompoundIndex = keyIndex + "+" + sourceIndex
+const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" + sourceIndex
+
+// flagSetId defaults to a UUID generated at startup to make our queries consistent
+// any flag without a "flagSetId" is assigned this one; it's never exposed externally
+var nilFlagSetId = uuid.New().String()
+
+// A selector represents a set of constraints used to query the store.
+type Selector struct {
+	indexMap map[string]string
```
toddbaert, line 32: I didn't add any tests directly for this struct (it's indirectly tested in many places) but I think I probably should (and I can do that in this PR if it's preferred).

core/pkg/store/query.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// NewSelector creates a new Selector from a selector expression string.
+// For example, to select flags from source "./mySource" and flagSetId "1234", use the expression:
+// "source=./mySource,flagSetId=1234"
+func NewSelector(selectorExpression string) Selector {
+	return Selector{
+		indexMap: expressionToMap(selectorExpression),
+	}
+}
```
toddbaert, line 42: :mega: One thing I think we should do, probably before merging this, is enhance this to interpret selectors without any expressions as sources, ie: `"source=../my/source"` and  `"../my/source"` would mean the same thing. This is to preserve the existing behavior of the selector in the `sync.proto`, which currently assumes the selector is a reference to a source. If we don't do so, this will be a breaking behavioral change we will have to call out loudly.

core/pkg/store/store.go
```diff
@@ -76,48 +127,76 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
+		flag, ok := raw.(model.Flag)
+		if (err != nil) || !ok {
+			return model.Flag{}, queryMeta, false
+		}
+		return flag, queryMeta, true
 
-func (f *Store) SelectorForFlag(_ context.Context, flag model.Flag) string {
-	f.mx.RLock()
-	defer f.mx.RUnlock()
+	} else {
```
tangenti: nit: `else` can be removed as there's a return in the block above

core/pkg/store/store.go
```diff
@@ -76,48 +127,76 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
+		// get all flags with the given key, and keep the one with the highest priority
+		s.logger.Debug(fmt.Sprintf("getting highest priority flag with key: %s", key))
+		it, err := txn.Get(flagsTable, keyIndex, key)
+		if err != nil {
+			return model.Flag{}, queryMeta, false
+		}
+		flag := model.Flag{}
+		var found bool
+		for raw := it.Next(); raw != nil; raw = it.Next() {
+			found = true
```
tangenti: Should `found` be set after type casting?

core/pkg/store/store.go
```diff
@@ -4,66 +4,117 @@ import (
...
+	Flags map[string]model.Flag
+}
 
 type IStore interface {
-	GetAll(ctx context.Context) (map[string]model.Flag, model.Metadata, error)
-	Get(ctx context.Context, key string) (model.Flag, model.Metadata, bool)
-	SelectorForFlag(ctx context.Context, flag model.Flag) string
+	GetAll(ctx context.Context, selector *Selector, watcher chan Payload) (map[string]model.Flag, model.Metadata, error)
+	Get(ctx context.Context, key string, selector *Selector) (model.Flag, model.Metadata, bool)
 }
```
tangenti, line 29: I'd suggesting adding an interface check here:

```
var _  IStore = (*Store)(nil)
```

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
-	it, err := txn.Get("flags", "id")
+	queryMeta := model.Metadata{}
+
+	var it memdb.ResultIterator
+	var err error
+	if selector != nil && !selector.isEmpty() {
+		queryMeta = selector.SelectorToMetadata()
+		indexId, constraints := selector.SelectorMapToQuery()
+		s.logger.Debug(fmt.Sprintf("getting all flags with query: %s, %v", indexId, constraints))
+		it, err = txn.Get("flags", indexId, constraints...)
```
tangenti: nit: `txn.Get(flagsTable...` to keep the consistency

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
 
-	for obj := it.Next(); obj != nil; obj = it.Next() {
-		flag := obj.(model.Flag)
+	for raw := it.Next(); raw != nil; raw = it.Next() {
+		flag := raw.(model.Flag)
+		if existing, ok := flags[flag.Key]; ok {
+			if flag.Priority < existing.Priority {
+				s.logger.Debug(fmt.Sprintf("discarding duplicate flag %s from lower priority source %s in favor of flag from source %s", flag.Key, s.sources[flag.Priority], s.sources[existing.Priority]))
+				continue // we already have a higher priority flag
+			} else {
```
tangenti: nit: `else` is not needed with `continue`.

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
 		flags[flag.Key] = flag
 	}
 
-	return flags, f.getMetadata(), nil
+	if watcher != nil {
+
+		// a "one-time" watcher that will be notified of changes to the query
+		changes := it.WatchCh()
+
+		go func() {
```
tangenti: Can we have a separate function for watching? It can be modified to a for(ever) loop with the `WatchSet` provided by memdb.

```
    for {
        txn := db.Txn(false)
        ws  := memdb.NewWatchSet()

        iter, _ := txn.Get(...)
        ws.Add(iter.WatchCh())
        
        watcher <- ...
        
        if err := ws.WatchCtx(ctx); err != nil {
            close(watcher)
            return err                    // ctx cancelled or deadline
        }
        // loop restarts with a brand-new snapshot & watch set
    }
```

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
+				s.logger.Debug("flags store has changed, notifying watchers")
+
+				// recursively get all flags again in bytes new goroutine to keep the watcher responsive
+				// as long as we do this in bytes new goroutine, we don't risk stack overflow
+				bytes, _, err := s.GetAll(ctx, selector, watcher)
+				if err != nil {
+					s.logger.Error(fmt.Sprintf("error getting flags in watcher: %v", err))
+					break
+				}
+				watcher <- Payload{
```
tangenti: `s.GetAll` will start another goroutine, which may close the `watcher` channel if the context is cancelled, and if it happens before L265 is executed, the program will panic

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
+	return utils.BuildNotifications(oldFlags, flags), resyncRequired
 }
 
-func (f *Store) GetMetadataForSource(source string) model.Metadata {
-	f.mx.RLock()
-	defer f.mx.RUnlock()
-	perSource, ok := f.MetadataPerSource[source]
-	if ok && perSource != nil {
-		return maps.Clone(perSource)
+func mergeMetadata(m1, m2 model.Metadata) model.Metadata {
```
tangenti: I'd call it `patch...(original, patch)` or `overlay(old, new)` as m2 can override some values.

core/pkg/utils/build_notifications.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"reflect"
+
+	"github.com/open-feature/flagd/core/pkg/model"
+)
+
+func BuildNotifications(oldFlags, newFlags map[string]model.Flag) map[string]interface{} {
```
tangenti: Does it make sense to 

1. Create a package called `notifications`
2. Define the type Notifications as `map[string]any`
3. Rename this function to `NewFromFlags` or `New` and returns the defined type

?

core/pkg/store/query.go
```diff
@@ -0,0 +1,122 @@
...
+
+func expressionToMap(selector string) map[string]string {
+	selectorMap := make(map[string]string)
+	if selector == "" {
+		return selectorMap
+	}
+
+	// Split the selector by commas
+	pairs := strings.Split(selector, ",")
```
tangenti: It seems both `,` and `=` are allowed in the flag set IDs, so the parsing here may not work as intended for corner cases.

tangenti (commented):
There was a concern over the selection performance - do we have any benchmark results showing the impact of this change?

core/pkg/store/query.go
toddbaert: I considered this, but wasn't sure how to proceed! Thanks for the eye for detail. ..

Do you have a suggestion? In general there's not much validation here. I can't decide on the best course. I'm not sure I want to implement escape characters or something.

toddbaert:
> There was a concern over the selection performance - do we have any benchmark results showing the impact of this change?

I will run or benchmark suite again, like I did when we did the memdb migration. I expect things will be slower, especially for "selector-less" requests, but IMO as long as it's not an order of magnetude  shower or more, it's acceptable, considering how fast flagd already is.

core/pkg/store/store.go
toddbaert: This sounds much nicer than my optional recursion. Let me see...

core/pkg/store/query.go
```diff
@@ -0,0 +1,122 @@
...
+// flags table and index constants
+const flagsTable = "flags"
+
+const idIndex = "id"
+const keyIndex = "key"
+const sourceIndex = "source"
+const priorityIndex = "priority"
+const flagSetIdIndex = "flagSetId"
+
+// compound indices; maintain sub-indexes alphabetically; order matters; these must match what's generated in
```
chrfwow: > these must match what's generated in [...]

a few words might have gone lost here

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
 				continue
 			}
+
+			s.logger.Debug(
+				fmt.Sprintf(
+					"store resync triggered: flag %s has been deleted from source %s",
+					key, source,
+				),
+			)
+			continue
```
chrfwow: Do we need a second log?

core/pkg/store/store.go
```diff
@@ -131,146 +210,180 @@ func (f *Store) String() (string, error) {
...
+
+		raw, err := txn.First(flagsTable, keySourceCompoundIndex, key, source)
+		if err != nil {
+			s.logger.Error(fmt.Sprintf("unable to get flag %s from source %s: %v", key, source, err))
+			continue
+		}
+		oldFlag, ok := raw.(model.Flag)
+		// if we already have a flag with the same key and source, we need to check if it has the same flagSetId
+		if ok {
+			if oldFlag.FlagSetId != newFlag.FlagSetId {
```
chrfwow, line 304: why do we delete the flag if the flagSets are different?

toddbaert: Good question.

It's because logically the primary key is flagSetId+key. We are looping through all the flags for a source here; if the flag set has changed, the previous flag must be deleted (any clients selecting on the old flag set should get a message now missing this flag). This is an important case some of my new tests cover.

Does this make sense? 

I can add a comment about this.

chrfwow: That makes sense

core/pkg/evaluator/json.go
```diff
@@ -118,7 +118,7 @@ func (je *JSON) SetState(payload sync.DataSync) (map[string]interface{}, bool, e
 	var events map[string]interface{}
 	var reSync bool
 
-	events, reSync = je.store.Update(payload.Source, payload.Selector, definition.Flags, definition.Metadata)
+	events, reSync = je.store.Update(payload.Source, definition.Flags, definition.Metadata)
 
 	// Number of events correlates to the number of flags changed through this sync, record it
 	span.SetAttributes(attribute.Int("feature_flag.change_count", len(events)))
```
thisthat, line 124: unrelated to this PR, but this looks like it could also be a metric

core/pkg/evaluator/json.go
```diff
@@ -150,8 +150,12 @@ func (je *Resolver) ResolveAllValues(ctx context.Context, reqID string, context
 	_, span := je.tracer.Start(ctx, "resolveAll")
 	defer span.End()
 
-	var err error
-	allFlags, flagSetMetadata, err := je.store.GetAll(ctx)
+	selector := store.NewSelector("")
+	s := ctx.Value(store.SelectorContextKey{})
+	if s != nil {
+		selector = s.(store.Selector)
+	}
```
thisthat: I would avoid allocating memory if we have it already in the ctx
```suggestion
	var selector store.Selector
	s := ctx.Value(store.SelectorContextKey{})
	if s != nil {
		selector = s.(store.Selector)
	} else {
		selector = store.NewSelector("")
	}
```

core/pkg/evaluator/json.go
```diff
@@ -302,19 +306,19 @@ func resolve[T constraints](ctx context.Context, reqID string, key string, conte
 func (je *Resolver) evaluateVariant(ctx context.Context, reqID string, flagKey string, evalCtx map[string]any) (
 	variant string, variants map[string]interface{}, reason string, metadata map[string]interface{}, err error,
 ) {
-	flag, metadata, ok := je.store.Get(ctx, flagKey)
+
+	selector := store.NewSelector("")
```
thisthat: ditto

core/pkg/store/store.go
toddbaert: Yes, this is handy in Go, and something I always forget to do.

core/pkg/store/store.go
toddbaert: This is fixed with your recommended addition of a `Watch` func.

core/pkg/store/store.go
toddbaert: I've added a comment to this effect.

core/pkg/store/store.go
toddbaert: No.

core/pkg/utils/build_notifications.go
toddbaert: Nice, done.

core/pkg/store/store.go
toddbaert: This is much nicer. Excellent recommendation. Please see [this commit](https://github.com/open-feature/flagd/pull/1702/commits/a04799e9360f38d1ccfd297b1b13557c0e41a066) for my changes (mostly related to this comment, but also some of your others).

Please note that to add the `Watch` func you recommended here, I extracted some private functions to reduce repitition (otherwise `GetAll` and `Watch` would have contained some trivial duplication). Despite that, I still think the ergonomics of your proposal are better, since in many cases we did a `GetAll` without watchng, and vice versa.

I also extracted the notifications as you suggested and added tests for that.

cc @chrfwow

core/pkg/notifications/notifications.go
```diff
@@ -1,19 +1,25 @@
-package utils
+package notifications
```
toddbaert, line 1: @tangenti as you recommended.

core/pkg/store/store.go
```diff
@@ -212,68 +213,19 @@ func (f *Store) String() (string, error) {
...
-	if selector != nil && !selector.isEmpty() {
+	if selector != nil && !selector.IsEmpty() {
 		queryMeta = selector.SelectorToMetadata()
-		indexId, constraints := selector.SelectorMapToQuery()
-		s.logger.Debug(fmt.Sprintf("getting all flags with query: %s, %v", indexId, constraints))
-		it, err = txn.Get("flags", indexId, constraints...)
-	} else {
-		// no selector, get all flags
-		it, err = txn.Get(flagsTable, idIndex)
 	}
```
toddbaert: Extracted into `selectOrAll` (because we also need it in `Watch`.

core/pkg/store/store.go
```diff
@@ -212,68 +213,19 @@ func (f *Store) String() (string, error) {
...
-		flag := raw.(model.Flag)
-		if existing, ok := flags[flag.Key]; ok {
-			if flag.Priority < existing.Priority {
-				s.logger.Debug(fmt.Sprintf("discarding duplicate flag %s from lower priority source %s in favor of flag from source %s", flag.Key, s.sources[flag.Priority], s.sources[existing.Priority]))
-				continue // we already have a higher priority flag
-			} else {
-				s.logger.Debug(fmt.Sprintf("overwriting duplicate flag %s from lower priority source %s in favor of flag from source %s", flag.Key, s.sources[existing.Priority], s.sources[flag.Priority]))
-			}
-		}
-		flags[flag.Key] = flag
```
toddbaert: Extracted into `collect` (because we also need it in `Watch`.

core/pkg/store/store.go
```diff
@@ -212,68 +213,19 @@ func (f *Store) String() (string, error) {
...
-				watcher <- Payload{
-					Flags: bytes,
-				}
-
-			case <-ctx.Done():
-				close(watcher)
-			}
-		}()
-	}
-
```
toddbaert: Moved to `Watch`.

flagd/pkg/service/flag-sync/handler.go
```diff
@@ -49,19 +49,7 @@ func (s syncHandler) SyncFlags(req *syncv1.SyncFlagsRequest, server syncv1grpc.F
...
-	}
-	flagsString, err := json.Marshal(flags)
-	if err != nil {
-		return status.Error(codes.DataLoss, "error marshalling flags")
-	}
-	err = server.Send(&syncv1.SyncFlagsResponse{FlagConfiguration: string(flagsString), SyncContext: syncContext})
-
-	if err != nil {
-		return err
-	}
```
toddbaert: `Watch` makes this much cleaner; we don't need to call `GetAll` for an initial payload.

core/pkg/store/query.go
```diff
@@ -28,15 +28,24 @@ const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" +
 var nilFlagSetId = uuid.New().String()
 
 // A selector represents a set of constraints used to query the store.
-type Selector struct {
+type Selector interface {
```
toddbaert: I made this an interface because I was passing it as a pointer everywhere anyway.

core/pkg/store/query.go
```diff
@@ -0,0 +1,131 @@
...
+	sort.Strings(keys)
+
+	for _, key := range keys {
+		indexId += key + "+"
+		qs = append(qs, s.indexMap[key])
+	}
+
+	indexId = strings.TrimSuffix(indexId, "+")
+	// Convert []string to []interface{}
+	c := make([]interface{}, 0, len(qs)+1)
```
thisthat: why the `+1`? 🤔

core/pkg/store/store.go
```diff
@@ -76,48 +131,75 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
-		db:                db,
-		logger:            logger,
+		sources: s,
+		db:      db,
+		logger:  logger,
 	}, nil
 }
 
-// Deprecated: use NewStore instead
+// Deprecated: use NewStore instead - will be removed very soon.
```
thisthat, line 144: is there already a ticket to track this deprecation?

core/pkg/store/store.go
```diff
@@ -4,66 +4,121 @@ import (
...
+
+type FlagQueryResult struct {
+	Flags map[string]model.Flag
+}
 
 type IStore interface {
-	GetAll(ctx context.Context) (map[string]model.Flag, model.Metadata, error)
-	Get(ctx context.Context, key string) (model.Flag, model.Metadata, bool)
-	SelectorForFlag(ctx context.Context, flag model.Flag) string
+	Get(ctx context.Context, key string, selector Selector) (model.Flag, model.Metadata, bool)
```
thisthat: we should unify this such that both getters return `error`

core/pkg/store/store.go
```diff
@@ -76,48 +131,75 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
+	it, err := txn.Get(flagsTable, keyIndex, key)
+	if err != nil {
+		return model.Flag{}, queryMeta, false
+	}
+	flag := model.Flag{}
+	var found bool
+	for raw := it.Next(); raw != nil; raw = it.Next() {
+		s.logger.Debug(fmt.Sprintf("got range scan: %v", raw))
+		nextFlag, ok := raw.(model.Flag)
+		found = true
```
thisthat: I would feel more confident moving this after the cast check so we can be sure `flag` has some value 🤔

core/pkg/store/store.go
```diff
@@ -76,48 +131,75 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
 	}
-	return flag, f.GetMetadataForSource(flag.Source), true
-}
-
-func (f *Store) SelectorForFlag(_ context.Context, flag model.Flag) string {
-	f.mx.RLock()
-	defer f.mx.RUnlock()
-
-	return f.SourceDetails[flag.Source].Selector
+	// otherwise, get all flags with the given key, and keep the one with the highest priority
```
thisthat: NIT, just to highlight the sorting in case of same prio
```suggestion
	// otherwise, get all flags with the given key, and keep the last one with the highest priority
```

core/pkg/store/store.go
```diff
@@ -131,146 +213,176 @@ func (f *Store) String() (string, error) {
...
-		newFlag.Selector = selector
+		s.logger.Debug(fmt.Sprintf("got metadata %v", metadata))
+
 		newFlag.Key = key
-		storedFlag, _, ok := f.Get(context.Background(), key)
+		newFlag.Source = source
+		newFlag.Priority = priority
+		newFlag.Metadata = patchMetadata(metadata, newFlag.Metadata)
+
+		// flagSetId defaults to a UUID generated at startup to make our quires isomorphic
```
thisthat: nit
```suggestion
		// flagSetId defaults to a UUID generated at startup to make our queries isomorphic
```

core/pkg/store/store.go
```diff
@@ -131,146 +213,176 @@ func (f *Store) String() (string, error) {
...
-				continue
+	// get all flags for the source we are updating
+	selector := NewSelector(sourceIndex + "=" + source)
+	oldFlags, _, _ := s.GetAll(context.Background(), selector)
+
+	s.mx.Lock()
+	for key := range oldFlags {
+		if _, ok := flags[key]; !ok {
+			// flag has been deleted
+			s.logger.Debug(fmt.Sprintf("flag %s has been deleted from source %s", key, source))
```
thisthat, line 267: this could be at INFO level 🤔

core/pkg/store/store.go
```diff
@@ -131,146 +213,176 @@ func (f *Store) String() (string, error) {
...
-	defer f.mx.RUnlock()
-	perSource, ok := f.MetadataPerSource[source]
-	if ok && perSource != nil {
-		return maps.Clone(perSource)
+// Watch the result-set of a selector for changes, sending updates to the watcher channel.
+func (s *Store) Watch(ctx context.Context, selector Selector, watcher chan FlagQueryResult) {
+	go func() {
+		for {
+			ws := memdb.NewWatchSet()
+			it, err := s.selectOrAll(selector)
```
thisthat, line 332: I would log at ERROR level and return if `err != nil`

core/pkg/store/store.go
```diff
@@ -353,9 +297,11 @@ func (s *Store) Update(
 			continue
 		}
 		oldFlag, ok := raw.(model.Flag)
-		// if we already have a flag with the same key and source, we need to check if it has the same flagSetId
+		// If we already have a flag with the same key and source, we need to check if it has the same flagSetId
 		if ok {
 			if oldFlag.FlagSetId != newFlag.FlagSetId {
+				// If the flagSetId is different, we need to delete the, since flagSetId+key represents the primary index, and it's now been changed.
```
chrfwow:
```suggestion
				// If the flagSetId is different, we need to delete the entry, since flagSetId+key represents the primary index, and it's now been changed.
```
missed a word?

core/pkg/store/store.go
```diff
@@ -353,9 +297,11 @@ func (s *Store) Update(
...
+				// Yhis is important especially for clients listening to flagSetId changes, as they expect the flag to be removed from the set in this case.
```
chrfwow:
```suggestion
				// This is important especially for clients listening to flagSetId changes, as they expect the flag to be removed from the set in this case.
```

chrfwow (approved):
I'm not an expert here, but as far as I can tell there are no obvious problems for me

core/pkg/store/query.go
toddbaert: Oops. I'm 99% sure this was a copilot tab that I didn't catch.

toddbaert:
```suggestion
	c := make([]interface{}, 0, len(qs))
```

core/pkg/store/store.go
toddbaert: Nope but I'll make one if this is merged

core/pkg/store/query.go
toddbaert: @thisthat 

Any thoughts on this?

toddbaert:
~~My recent commits created a race condition (just in a test assertion) I think it's an easy fix, I'll get to it later today.~~

EDIT:

Fixed.

core/pkg/store/store.go
toddbaert: Good catch, added a return and close of the `watcher` chan.

core/pkg/store/store.go
toddbaert: I'm not sure, I think this is perfectly normal behavior so I'm on the side of debug.

core/pkg/store/store.go
toddbaert: :100:

core/pkg/evaluator/json.go
toddbaert: E F F I C I E N C Y.

core/pkg/store/store.go
toddbaert: Yep, done.

core/pkg/evaluator/json.go
toddbaert: Ya this is a good point. We should probably review all the metrics as part of pre-v1.

toddbaert:
@tangenti @beeme1mr after running the benchmark suite, there's a small reduction in speed (some tests are identical to before the change, or even slightly faster, and the most negatively impacted are ~33% slower) but this is not significant in my opinion; we're still talking about 1/100th of a millisecond for our _slowest_ operations.

There's also a small increase in allocations (something like 10%, eyeballing it). See below:

<details>
  <summary>Bench diff</summary>

```diff
-BenchmarkFractionalEvaluation/test_a@faas.com-16         	  423930	     13316 ns/op	    7229 B/op	     135 allocs/op
-BenchmarkFractionalEvaluation/test_b@faas.com-16         	  469594	     13677 ns/op	    7229 B/op	     135 allocs/op
-BenchmarkFractionalEvaluation/test_c@faas.com-16         	  569103	     13286 ns/op	    7229 B/op	     135 allocs/op
-BenchmarkFractionalEvaluation/test_d@faas.com-16         	  412386	     13023 ns/op	    7229 B/op	     135 allocs/op
-BenchmarkResolveBooleanValue/test_staticBoolFlag-16      	 3106903	      1792 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveBooleanValue/test_targetingBoolFlag-16   	  448164	     11250 ns/op	    6065 B/op	      87 allocs/op
-BenchmarkResolveBooleanValue/test_staticObjectFlag-16    	 3958750	      1476 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveBooleanValue/test_missingFlag-16         	 5331808	      1353 ns/op	     784 B/op	      12 allocs/op
-BenchmarkResolveBooleanValue/test_disabledFlag-16        	 4530751	      1301 ns/op	    1072 B/op	      13 allocs/op
-BenchmarkResolveStringValue/test_staticStringFlag-16     	 4583056	      1525 ns/op	    1040 B/op	      13 allocs/op
-BenchmarkResolveStringValue/test_targetingStringFlag-16  	  839954	     10388 ns/op	    6097 B/op	      89 allocs/op
-BenchmarkResolveStringValue/test_staticObjectFlag-16     	 4252830	      1677 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveStringValue/test_missingFlag-16          	 3743324	      1495 ns/op	     784 B/op	      12 allocs/op
-BenchmarkResolveStringValue/test_disabledFlag-16         	 3495699	      1709 ns/op	    1072 B/op	      13 allocs/op
-BenchmarkResolveFloatValue/test:_staticFloatFlag-16      	 4382868	      1511 ns/op	    1024 B/op	      13 allocs/op
-BenchmarkResolveFloatValue/test:_targetingFloatFlag-16   	  867987	     10344 ns/op	    6081 B/op	      89 allocs/op
-BenchmarkResolveFloatValue/test:_staticObjectFlag-16     	 3913120	      1695 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveFloatValue/test:_missingFlag-16          	 3910468	      1349 ns/op	     784 B/op	      12 allocs/op
-BenchmarkResolveFloatValue/test:_disabledFlag-16         	 3642919	      1666 ns/op	    1072 B/op	      13 allocs/op
-BenchmarkResolveIntValue/test_staticIntFlag-16           	 4077288	      1349 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveIntValue/test_targetingNumberFlag-16     	  922383	      7601 ns/op	    6065 B/op	      87 allocs/op
-BenchmarkResolveIntValue/test_staticObjectFlag-16        	 4995128	      1229 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveIntValue/test_missingFlag-16             	 5574153	      1274 ns/op	     768 B/op	      12 allocs/op
-BenchmarkResolveIntValue/test_disabledFlag-16            	 3633708	      1734 ns/op	    1072 B/op	      13 allocs/op
-BenchmarkResolveObjectValue/test_staticObjectFlag-16     	 1624102	      4559 ns/op	    2243 B/op	      37 allocs/op
-BenchmarkResolveObjectValue/test_targetingObjectFlag-16  	  443880	     11995 ns/op	    7283 B/op	     109 allocs/op
-BenchmarkResolveObjectValue/test_staticBoolFlag-16       	 3462445	      1665 ns/op	    1008 B/op	      11 allocs/op
-BenchmarkResolveObjectValue/test_missingFlag-16          	 4207567	      1458 ns/op	     784 B/op	      12 allocs/op
-BenchmarkResolveObjectValue/test_disabledFlag-16         	 3407262	      1848 ns/op	    1072 B/op	      13 allocs/op
-PASS
-ok  	github.com/open-feature/flagd/core/pkg/evaluator	239.506s
+BenchmarkFractionalEvaluation/test_a@faas.com-16         	  466902	     17410 ns/op	    7683 B/op	     150 allocs/op
+BenchmarkFractionalEvaluation/test_b@faas.com-16         	  369529	     15836 ns/op	    7683 B/op	     150 allocs/op
+BenchmarkFractionalEvaluation/test_c@faas.com-16         	  384133	     16960 ns/op	    7683 B/op	     150 allocs/op
+BenchmarkFractionalEvaluation/test_d@faas.com-16         	  322873	     17541 ns/op	    7683 B/op	     150 allocs/op
+BenchmarkResolveBooleanValue/test_staticBoolFlag-16      	 3500460	      2044 ns/op	    1176 B/op	      25 allocs/op
+BenchmarkResolveBooleanValue/test_targetingBoolFlag-16   	  610275	     10674 ns/op	    6249 B/op	     101 allocs/op
+BenchmarkResolveBooleanValue/test_staticObjectFlag-16    	 2736013	      2451 ns/op	    1192 B/op	      25 allocs/op
+BenchmarkResolveBooleanValue/test_missingFlag-16         	 2921277	      2202 ns/op	    1016 B/op	      28 allocs/op
+BenchmarkResolveBooleanValue/test_disabledFlag-16        	 2441956	      2608 ns/op	    1240 B/op	      27 allocs/op
+BenchmarkResolveStringValue/test_staticStringFlag-16     	 2510768	      2509 ns/op	    1224 B/op	      27 allocs/op
+BenchmarkResolveStringValue/test_targetingStringFlag-16  	  419666	     12443 ns/op	    6282 B/op	     103 allocs/op
+BenchmarkResolveStringValue/test_staticObjectFlag-16     	 2544326	      2534 ns/op	    1192 B/op	      25 allocs/op
+BenchmarkResolveStringValue/test_missingFlag-16          	 2883064	      2302 ns/op	    1016 B/op	      28 allocs/op
+BenchmarkResolveStringValue/test_disabledFlag-16         	 2482075	      2679 ns/op	    1241 B/op	      27 allocs/op
+BenchmarkResolveFloatValue/test:_staticFloatFlag-16      	 2715316	      2462 ns/op	    1192 B/op	      27 allocs/op
+BenchmarkResolveFloatValue/test:_targetingFloatFlag-16   	  491826	     13936 ns/op	    6265 B/op	     103 allocs/op
+BenchmarkResolveFloatValue/test:_staticObjectFlag-16     	 2295111	      2364 ns/op	    1192 B/op	      25 allocs/op
+BenchmarkResolveFloatValue/test:_missingFlag-16          	 2779935	      2477 ns/op	    1016 B/op	      28 allocs/op
+BenchmarkResolveFloatValue/test:_disabledFlag-16         	 2432382	      2799 ns/op	    1241 B/op	      27 allocs/op
+BenchmarkResolveIntValue/test_staticIntFlag-16           	 2850421	      2564 ns/op	    1176 B/op	      25 allocs/op
+BenchmarkResolveIntValue/test_targetingNumberFlag-16     	  628972	     12417 ns/op	    6250 B/op	     101 allocs/op
+BenchmarkResolveIntValue/test_staticObjectFlag-16        	 2350480	      2459 ns/op	    1192 B/op	      25 allocs/op
+BenchmarkResolveIntValue/test_missingFlag-16             	 3103119	      2174 ns/op	    1000 B/op	      28 allocs/op
+BenchmarkResolveIntValue/test_disabledFlag-16            	 2485292	      2499 ns/op	    1241 B/op	      27 allocs/op
+BenchmarkResolveObjectValue/test_staticObjectFlag-16     	 1000000	      5445 ns/op	    2427 B/op	      51 allocs/op
+BenchmarkResolveObjectValue/test_targetingObjectFlag-16  	  326774	     16668 ns/op	    7467 B/op	     123 allocs/op
+BenchmarkResolveObjectValue/test_staticBoolFlag-16       	 2559854	      2671 ns/op	    1176 B/op	      25 allocs/op
+BenchmarkResolveObjectValue/test_missingFlag-16          	 2775954	      2186 ns/op	    1016 B/op	      28 allocs/op
+BenchmarkResolveObjectValue/test_disabledFlag-16         	 2447349	      2761 ns/op	    1241 B/op	      27 allocs/op
```

</details>

core/pkg/store/query.go
thisthat: Since we're getting this data as an HTTP Header, I would vote to follow the format specified in [RFC9651](https://www.rfc-editor.org/rfc/rfc9651) or a subset of it since I don't think we will support list 🤔
Hence, we should replace `,` with `;` and disallow `=` in the schema definition of `metadata.flagSetId`.

toddbaert: I'm ok with that. And we can enforce this in the json schema so that violation is in flagsetid will cause json validation failures and syntax errors in IDEs

@beeme1mr @tangenti WDYT?

tangenti: I would prefer to keep the simplicity for the moment until we have better understandings of more use cases of the selector.

core/pkg/store/store.go
```diff
@@ -131,146 +216,176 @@ func (f *Store) String() (string, error) {
...
+			it, err := s.selectOrAll(selector)
+			ws.Add(it.WatchCh())
+
+			flags := s.collect(it)
+			watcher <- FlagQueryResult{
+				Flags: flags,
+			}
+
+			if err = ws.WatchCtx(ctx); err != nil {
+				close(watcher)
```
tangenti, line 347: Would be helpful to log the error down at least for `debug`.

core/pkg/store/store.go
```diff
@@ -131,146 +216,176 @@ func (f *Store) String() (string, error) {
...
 }
 
-func (f *Store) GetMetadataForSource(source string) model.Metadata {
-	f.mx.RLock()
-	defer f.mx.RUnlock()
-	perSource, ok := f.MetadataPerSource[source]
-	if ok && perSource != nil {
-		return maps.Clone(perSource)
+// Watch the result-set of a selector for changes, sending updates to the watcher channel.
+func (s *Store) Watch(ctx context.Context, selector Selector, watcher chan FlagQueryResult) {
```
tangenti: nit: `chan<-`

core/pkg/store/query.go
```diff
@@ -0,0 +1,131 @@
...
+const keySourceCompoundIndex = keyIndex + "+" + sourceIndex
+const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" + sourceIndex
+
+// flagSetId defaults to a UUID generated at startup to make our queries consistent
+// any flag without a "flagSetId" is assigned this one; it's never exposed externally
+var nilFlagSetId = uuid.New().String()
+
+// A selector represents a set of constraints used to query the store.
+type Selector interface {
+	SelectorMapToQuery() (indexId string, constraints []interface{})
```
tangenti: nit: keep consistent for the method name

```
SelectorMapToQuery() -> ToQuery()
SElectorToMetadata() -> ToMetadata()
```

core/pkg/store/query.go
```diff
@@ -0,0 +1,131 @@
...
+var _ Selector = (*selector)(nil)
+
+type selector struct {
+	indexMap map[string]string
+}
+
+// NewSelector creates a new Selector from a selector expression string.
+// For example, to select flags from source "./mySource" and flagSetId "1234", use the expression:
+// "source=./mySource,flagSetId=1234"
+func NewSelector(selectorExpression string) Selector {
```
tangenti, line 38: I don't see a good reason to make `Selector` an interface for now - do we soon expect to have another implementation of selector?

core/pkg/store/query.go
```diff
@@ -0,0 +1,131 @@
...
+func (s selector) WithIndex(key string, value string) Selector {
+	m := maps.Clone(s.indexMap)
+	m[key] = value
+	return selector{
+		indexMap: m,
+	}
+}
+
+func (s selector) IsEmpty() bool {
+	return len(s.indexMap) == 0
```
tangenti: You could use `s == nli || len(s.indexMap) == 0` to avoid the nil check before calling `IsEmpty`.

core/pkg/store/query.go
toddbaert: It was just because it was always a pointer (to make it optional). I can revert, it's just a style choice I guess.

core/pkg/store/query.go
toddbaert: I agree with keeping it simple for now, but @thisthat 's proposal brings up the question of whether `,` or `;` might be a better separator token.

core/pkg/store/query.go
toddbaert: Agreed

core/pkg/store/query.go
toddbaert: I've reverted to make this a struct and have funcs take pointers to it, as before.

core/pkg/store/query.go
toddbaert: I misunderstood you at first, but good idea.

core/pkg/store/query.go
toddbaert: I added full testing here.

core/pkg/store/query.go
```diff
@@ -0,0 +1,121 @@
...
+// compound indices; maintain sub-indexes alphabetically; order matters; these must match what's generated in the SelectorMapToQuery func.
+const flagSetIdSourceCompoundIndex = flagSetIdIndex + "+" + sourceIndex
+const keySourceCompoundIndex = keyIndex + "+" + sourceIndex
+const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" + sourceIndex
+
+// flagSetId defaults to a UUID generated at startup to make our queries consistent
+// any flag without a "flagSetId" is assigned this one; it's never exposed externally
+var nilFlagSetId = uuid.New().String()
+
+type Selector struct {
```
toddbaert, line 31:
```suggestion
// A selector represents a set of constraints used to query the store.
type Selector struct {
```

core/pkg/store/query.go
thisthat: If I have to choose, I'd go with `;` since it splits parameters. However, I am happy with both options as long as it is removed from being a valid char in the schema. We can discuss this in a follow-up since this PR is already quite big :)

tangenti: In general I think it's not a good practice to mark some chars as special. 

I've considered ANTLR but it seems a bit overkill at the moment, but if we don't have a better option, I'd suggesting using ANTLR

core/pkg/notifications/notifications.go
tangenti: Thanks. Feel free to resolve - I'm unable to resolve comment (perhaps due to permission).

core/pkg/notifications/notifications.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/open-feature/flagd/core/pkg/model"
+)
+
+const typeField = "type"
+
+// Use to represent change notifications for  mode PROVIDER_CONFIGURATION_CHANGE events.
+type Notifications map[string]any
+
+// Generate notifications (deltas) from old and new flag sets for use in RPC mode PROVIDER_CONFIGURATION_CHANGE events.
+func NewFromFlags(oldFlags, newFlags map[string]model.Flag) map[string]interface{} {
```
tangenti: nit: 
```func NewFromFlags(oldFlags, newFlags map[string]model.Flag) Notifications```

File: core/pkg/store/store.go [hunk 54]
```diff
@@ -76,48 +131,77 @@ func NewStore(logger *logger.Logger) (*Store, error) {
...
+
+	// if present, use the selector to query the flags
+	if !selector.IsEmpty() {
+		selector := selector.WithIndex("key", key)
+		indexId, constraints := selector.ToQuery()
+		s.logger.Debug(fmt.Sprintf("getting flag with query: %s, %v", indexId, constraints))
+		raw, err := txn.First(flagsTable, indexId, constraints...)
+		flag, ok := raw.(model.Flag)
+		if (err != nil) || !ok {
+			return model.Flag{}, queryMeta, fmt.Errorf("flag %s not found: %w", key, err)
```
tangenti, line 167: Shall we distinguish these two cases?

Type casting failure suggests a bug somewhere, while a non-nil err can simply mean the flag is not found.

core/pkg/evaluator/json.go
```diff
@@ -301,19 +307,24 @@ func resolve[T constraints](ctx context.Context, reqID string, key string, conte
...
+	var selector store.Selector
+	s := ctx.Value(store.SelectorContextKey{})
+	if s != nil {
+		selector = s.(store.Selector)
+	} else {
+		selector = store.NewSelector("")
+	}
+	if s != nil {
+		selector = s.(store.Selector)
+	}
```
tangenti: L318 - L320 seems duplicate.

Also, this can be written as 
```
selector := store.NewSelector("")
if s != nil {
   selector = s.(store.Selector)
}
```

core/pkg/store/query.go
```diff
@@ -0,0 +1,133 @@
...
+const flagSetIdIndex = "flagSetId"
+
+// compound indices; maintain sub-indexes alphabetically; order matters; these must match what's generated in the SelectorMapToQuery func.
+const flagSetIdSourceCompoundIndex = flagSetIdIndex + "+" + sourceIndex
+const keySourceCompoundIndex = keyIndex + "+" + sourceIndex
+const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" + sourceIndex
+
+// flagSetId defaults to a UUID generated at startup to make our queries consistent
+// any flag without a "flagSetId" is assigned this one; it's never exposed externally
+var nilFlagSetId = uuid.New().String()
```
tangenti, line 28: What's the motivation behind this decision? 

IIUC, this will make it impossible for providers to select those flags without a flag set ID.

toddbaert: The motivation is mostly due to limitations in the query language and index support of the DB. To support the compound flagSetId+key primary index we want, we have to put something in the flagSetId, even when there isn't one defined by the config.

These flags can be selected by simply omitting a `flagSetId`. Logically, we are treating the `flagSetId: null` as it's own set, but the way the `Get` query with no constraints works [here](https://github.com/open-feature/flagd/pull/1702/files#diff-22143897620da6ff41e16b0a3b64441aad51ca0c91cb3bf18f19f4ea83f7776cR359-R360), everything is returned. In combination, we get the behavior that specifying a `flagSetId` will only return flags belonging to that set, but omitting one will return everything (with the previous "source priority" rules honored for duplicated flags). I have tests covering this in the `store_test.go`. See gif for a demo.

nil

toddbaert: Note that currently, there's no way to _explicitly_ select flags without a flagSetId - but I think that's acceptable, and perhaps even optimal - these flags don't belong to a particular set, so I would argue they shouldn't be queried that way, but instead only be returned in source-queries, or set-less queries as above. This can be changed though.

core/pkg/evaluator/json.go
toddbaert: I've changed this to:

```go
var selector store.Selector
s := ctx.Value(store.SelectorContextKey{})
if s != nil {
  selector = s.(store.Selector)
}
```

So we can avoid the allocation of a new selector un-necessarily (we can safely pass nil because `isEmpty` checks that).

core/pkg/store/store.go
toddbaert: fixed

core/pkg/store/query.go
toddbaert: I've added this.

core/pkg/store/query.go
toddbaert: I reverted this.

core/pkg/store/query.go
tangenti: Thanks. I'll take a look tomorrow to see if such query can be supported without a UUID hack.

I think we discussed the problem when introducing the flag lists. Imagine there are two flags {key: feature, set: A} and {key: feature, set: null}, when an in-process provider constructs the selector, there's no way for it to get a valid flag map with {key: feature, set: null} in the result. It will be a problem for adopting flag sets - basically without the ability to select on the null set, users have to create a "default" set for all existing flags if they need to add a flag with the same key under another set.

toddbaert: > Thanks. I'll take a look tomorrow to see if such query can be supported without a UUID hack.

I _think_ it can be done with a `memdb.ConditionalIndex`; but IMO, it adds some complexity to do it this way, without much additional value (UUIC, you'd need to do a query something like `txn.Get(flagsTable, flagSetPresentIndex, setISPresent)` where `flagSetPresentIndex` is a memdb.ConditionalIndex index returning true if flagSetId is set, and `setISPresent` is a bool parameter to match the memdb.ConditionalIndex predicate's return value against. Maybe you could branch of mine, and give it a shot. If it works I'm happy to merge it in.

You also may need to use a `memdb.CompoundMultiIndex`, because as the in-line doc says:

`if AllowMissing is set, not only is it valid to have empty index fields, but it will still create index values up to the first empty index. This means that if you have a value with an empty field, rather than using a prefix for lookup, you can simply pass in less arguments`

Otherwise, we can't even insert a a flag with a zero-value for `flagSetId`. There may be other ways as well, including some iteration.

> I think we discussed the problem when introducing the flag lists. Imagine there are two flags {key: feature, set: A} and {key: feature, set: null}, when an in-process provider constructs the selector, there's no way for it to get a valid flag map with {key: feature, set: null} in the result. It will be a problem for adopting flag sets - basically without the ability to select on the null set, users have to create a "default" set for all existing flags if they need to add a flag with the same key under another set.

This is true; I suppose I was being a but myopic here, assuming that if you're using flagSets you're using them comprehensively.

Even with the current implementation with the "UUID hack", it's possible to implement selection on null - I think the biggest question is how we'd encode that in the selector... would it be `"flagsSetId="` ? Another option could be to automatically put them in a "well-known" `default` flagSet... (ie: `flagSetId=default`) but I'm less enthusiastic about that. Take care that we also need to maintain the old behavior where all flags are returned and the source-based priority override is honored if no `flagSetId` is specified (even if we make it possible to select on the old set)

toddbaert: @tangenti would you be OK with merging this as is, and creating a new issue to discuss and implement:

- selection on null flagSetId
- possible refactors to avoid the nilFlagSetId/uuid hack?

tangenti: Yes, please. Sorry for not being able to check this earlier.

tangenti (approved):

PR #1704, opened 2025-08-05
chore!: removes the `fractionalEvaluation` operator since it has been replaced with `fractional`.

### LINKED ISSUE #1677: Remove deprecated `fractionalEvaluation` operator
The `fractionalEvaluation` operator has been replaced with `fractional` and can be removed.

[This Go file](https://github.com/open-feature/flagd/blob/2f330af516376238e889064500a285b13ce8bb9a/core/pkg/evaluator/legacy_fractional.go#L25) and its references can be deleted.

NeaguGeorgiana23:
## This PR

- adds this new feature

### Related Issues
- removes the `fractionalEvaluation` operator since it has been replaced with `fractional`.

Fixes #1677


beeme1mr (approved):
Thanks, @NeaguGeorgiana23! The docs reference the old `fractionalEvaluation` but only to explain how to migrate, so I think we can leave it as is.

PR #1705, opened 2025-08-07
refactor: store cleanup

andreyturkov:
## This PR
IStore interface usage instead of Store implementation
Separate locks for flag data and metadata in store

### DIFF
--- a/core/pkg/store/store.go
+++ b/core/pkg/store/store.go
@@ -31,7 +31,8 @@ type IStore interface {
 
 // store is the in-memory implementation of IStore. It should not be used directly by consumers.
 type store struct {
-	mx                sync.RWMutex
+	flagMx            sync.RWMutex
+	metaMx            sync.RWMutex
 	db                *memdb.MemDB
 	logger            *logger.Logger
 	FlagSources       []string
@@ -114,16 +115,15 @@ func (f *store) Get(_ context.Context, key string) (model.Flag, model.Metadata, 
 }
 
 func (f *store) SelectorForFlag(_ context.Context, flag model.Flag) string {
-	f.mx.RLock()
-	defer f.mx.RUnlock()
-
+	f.metaMx.RLock()
+	defer f.metaMx.RUnlock()
 	return f.SourceDetails[flag.Source].Selector
 }
 
 func (f *store) String() (string, error) {
 	f.logger.Debug("dumping flags to string")
-	f.mx.RLock()
-	defer f.mx.RUnlock()
+	f.flagMx.RLock()
+	defer f.flagMx.RUnlock()
 
 	state, _, err := f.GetAll(context.Background())
 	if err != nil {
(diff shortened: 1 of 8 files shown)


beeme1mr:
Hey @andreyturkov, thanks for the PR. There's a [fairly large refactor](https://github.com/open-feature/flagd/pull/1702) open right now that we're hoping to get in soon. I'd like to hold off on making this change until that one has been merged. FYI @toddbaert

andreyturkov:
> Hey @andreyturkov, thanks for the PR. There's a [fairly large refactor](https://github.com/open-feature/flagd/pull/1702) open right now that we're hoping to get in soon. I'd like to hold off on making this change until that one has been merged. FYI @toddbaert

I'have rebased my branch with the latest main. Please take a look.

beeme1mr:
The licensing issue was a false positive and I've added an exception in FOSSA.

core/pkg/store/store.go
```diff
@@ -25,11 +25,13 @@ type IStore interface {
...
 	Watch(ctx context.Context, selector *Selector, watcher chan<- FlagQueryResult)
+	Update(source string, flags map[string]model.Flag, metadata model.Metadata) (map[string]interface{}, bool)
+	String() (string, error)
 }
 
-var _ IStore = (*Store)(nil)
+var _ IStore = (*store)(nil)
 
-type Store struct {
+type store struct {
```
tangenti: The best practice is to keep the struct public with private properties, and return the pointer of the struct: https://go.dev/wiki/CodeReviewComments#interfaces

andreyturkov:
> Could you also check the usages of lock in `Store`?
> 
> Now the lock should only be used for metadata, as flag CRUD is done with transaction. In `String()` for example, the lock is still used which is not needed I think

You are absolutely right. The lock in the String() method is unnecessary as the underlying GetAll() call is transactional and doesn't rely on the mutex.
The same logic applies to the Update method.
I'll remove the mutex usage from both methods, and since it's no longer needed, I'll also remove the mutex field from the store struct itself.

core/pkg/store/store.go
andreyturkov: Done

core/pkg/store/store.go
```diff
@@ -45,7 +45,7 @@ type SourceDetails struct {
 
 // NewStore creates a new in-memory store with the given sources.
 // The order of sources in the slice determines their priority, when queries result in duplicate flags (queries without source or flagSetId), the higher priority source "wins".
-func NewStore(logger *logger.Logger, sources []string) (*Store, error) {
+func NewStore(logger *logger.Logger, sources []string) (IStore, error) {
```
tangenti: Here as well, let's use the struct pointer.

tangenti (approved):

toddbaert:
Great work @andreyturkov , thanks! merging once the CI completes...

Do you have any interest in implementing [this](https://github.com/open-feature/flagd/issues/1708)? It's something we believe is important; there's also room for an optional refactor if you're keen to that.

toddbaert (approved):

PR #1706, opened 2025-08-08
docs(ADR): update support for explicit code default values in flagd configuration

beeme1mr:
## This PR

- adds this new feature

### Related Issues

Fixes #1234523

### DIFF
--- a/docs/architecture-decisions/support-code-default.md
+++ b/docs/architecture-decisions/support-code-default.md
@@ -1,8 +1,8 @@
 ---
 status: accepted
-author: @beeme1mr
+author: Michael Beemer
 created: 2025-06-06
-updated: 2025-06-20
+updated: 2025-08-08
 ---
 
 # Support Explicit Code Default Values in flagd Configuration
@@ -49,14 +49,17 @@ Related discussions and context can be found in the [OpenFeature specification](
 
 We propose implementing **Option 1: Allow `null` as Default Variant**, potentially combined with **Option 2: Make Default Variant Optional** for maximum flexibility.
 
-The implementation leverages field presence in evaluation responses across all protocols (in-process, RPC, and OFREP). When a flag configuration has `defaultVariant: null`, the evaluation response omits the value field entirely, which serves as a programmatic signal to the client to use its code-defined default value.
+The implementation leverages field presence in evaluation responses across all protocols.
+When a flag configuration has `defaultVariant: null`, the evaluation response omits the value field entirely and uses the "DEFAULT" reason code, which serves as a programmatic signal to the client to use its code-defined default value.
 
 This approach offers several key advantages:
 
-1. **No Protocol Changes**: RPC and OFREP protocols remain unchanged
-2. **Clear Semantics**: Omitted value field = "use your code default"
-3. **Backward Compatible**: Existing clients and servers continue to work
-4. **Universal Pattern**: Works consistently across all evaluation modes
+1. **Semantically Correct**: Uses "DEFAULT" reason code which accurately represents the evaluation outcome
+2. **Success Responses**: Treats code default usage as successful evaluation, not an error
+3. **Clear Semantics**: Omitted value field = "use your code default"
+4. **Backward Compatible**: Existing clients and servers continue to work
+5. **Universal Pattern**: Works consistently across all evaluation modes
+6. **Accurate Telemetry**: Metrics correctly reflect successful evaluations rather than false errors
 
 The absence of a value field provides an unambiguous signal that distinguishes between "the server evaluated to null/false/empty" (value field present) and "the server delegates to your code default" (value field absent).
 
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/support-code-default.md
```diff
@@ -1,8 +1,8 @@
 ---
 status: accepted
-author: @beeme1mr
+author: Michael Beemer
```
beeme1mr: The @ symbol was breaking frontmatter parsing in GitHub 🤷‍♂️

docs/architecture-decisions/support-code-default.md
```diff
@@ -49,14 +49,17 @@ Related discussions and context can be found in the [OpenFeature specification](
...
 This approach offers several key advantages:
 
-1. **No Protocol Changes**: RPC and OFREP protocols remain unchanged
-2. **Clear Semantics**: Omitted value field = "use your code default"
-3. **Backward Compatible**: Existing clients and servers continue to work
-4. **Universal Pattern**: Works consistently across all evaluation modes
+1. **Semantically Correct**: Uses "DEFAULT" reason code which accurately represents the evaluation outcome
+2. **Success Responses**: Treats code default usage as successful evaluation, not an error
+3. **Clear Semantics**: Omitted value field = "use your code default"
+4. **Backward Compatible**: Existing clients and servers continue to work
```
guidobrei, line 60: Will hardly be backwards compatible.

docs/architecture-decisions/support-code-default.md
```diff
@@ -190,22 +236,32 @@ To ensure correct implementation across all components:
...
 - How do we ensure consistent behavior across all provider implementations?
     - Gherkin tests will be added to the flagd testbed to ensure all providers handle the new behavior consistently.
-- Should providers validate that the reason is "DEFAULT" when value is omitted, or accept any omitted value as delegation?
-    - Providers should accept any omitted value as delegation.
-- How do we handle edge cases where network protocols might strip empty fields?
-    - It would behaving as expected, as the absence of fields is the intended signal.
+- How do OFREP providers detect and handle responses with omitted value fields?
+    - Providers should check for the absence of the `value` field in successful responses and treat it as a signal to use code defaults.
+- Should we maintain backward compatibility for providers that don't yet support omitted value fields?
+    - Yes, older providers will continue to work but may not benefit from the code default deferral feature until updated.
```
guidobrei, line 258: If an old provider receives a config with reason `DEFAULT` in RPC mode, it will contain a datatype default value (`false`, `0`, ` ""`), which is not the same as code default. This would be a breaking change.

docs/architecture-decisions/support-code-default.md
```diff
@@ -190,22 +236,32 @@ To ensure correct implementation across all components:
     - Yes, we'll support both `null` and absent fields to maximize flexibility. An absent `defaultVariant` will be the equivalent of `null`.
 - What migration path should we recommend for users currently using workarounds?
     - Update the flag configurations to use `defaultVariant: null` and remove any misconfigured rulesets that force code defaults.
+- How should we handle the breaking change for older clients evaluating `defaultVariant: null` flags?
```
guidobrei, line 249: If flagd would be aware of supported provider features, it could fall back to FLAG_NOT_FOUND behavior for older providers.

docs/architecture-decisions/support-code-default.md
cupofcat: I think you can also use ``` `@beeme1mr` ``` or `(@beeme1mr)`

cupofcat:
@beeme1mr could you please also add some more details / examples of the behavior (and changes) in the in-process providers?

docs/architecture-decisions/support-code-default.md
guidobrei: > If flagd would be aware of supported provider features, it could fall back to FLAG_NOT_FOUND behavior for older providers.

We have the `provider_id` in the [sync.proto](https://github.com/open-feature/flagd-schemas/blob/main/protobuf/flagd/sync/v1/sync.proto#L27). Are providers using this to identify themselves?

cc @toddbaert, @aepfli

beeme1mr:
> could you please also add some more details / examples of the behavior (and changes) in the in-process providers?

I've added a short section explaining how in-process would. Does that address your concerns @cupofcat?

docs/architecture-decisions/support-code-default.md
toddbaert: No we're not using this field for version negotiation or anything like that, but I think before 1.0, nothing here worries me too much.

docs/architecture-decisions/support-code-default.md
toddbaert: Oh I think you're right about this. OFREP won't be a problem but I think RPC will due to the way protobuf works.

toddbaert: Ya...

> An optional field is in one of two possible states:

> the field is set, and contains a value that was explicitly set or parsed from the wire. It will be serialized to the wire.
the field is unset, and will return the default value. It will not be serialized to the wire.
You can check to see if the value was explicitly set.

I think we'll be forced to do another version of RPC and support both for a while. I don't think there's a clean way or else providers will get erroneous evaluations seeing the flag value as the zero value because they aren't checking the optionality of the field.

toddbaert: Cc @beeme1mr

docs/architecture-decisions/support-code-default.md
```diff
@@ -77,22 +80,57 @@ The absence of a value field provides an unambiguous signal that distinguishes b
...
-   - Client detects this reason value field and uses its code-defined default
-   - This same pattern works across all evaluation modes
+   - Server responds with reason "DEFAULT" and omits value and variant fields
+   - Client detects the omitted fields and uses its code-defined default
+   - This pattern works consistently across all evaluation modes
+
+3. **Protobuf Schema Changes**:
+
+   - Update response message definitions to use `optional` fields for `value` and `variant`
+   - This enables proper field presence detection for code default signaling
```
tangenti, line 92: Could the client solely rely on the reason field?

docs/architecture-decisions/support-code-default.md
toddbaert: @guidobrei 

After thinking about this some more, I think the best option is to add support for this with a new v2 proto with this now marked as an optional field, and add an option to opt-into that v2 proto in the providers... then after "a while" default to this option and remove the old implementation (before flagd v1)

tangenti: > If an old provider receives a config with reason `DEFAULT` in RPC mode, it will contain a datatype default value (`false`, `0`, ` ""`), which is not the same as code default. This would be a breaking change.

Does RPC mode mean the evaluation gRPC service? Is this an intended behavior for interpreting `DEFAULT`?

toddbaert: > Does RPC mode mean the evaluation gRPC service?

Yes.

> Is this an intended behavior for interpreting DEFAULT?

I'm not 100% sure I'm understanding your question, but maybe I can level-set:

Currently, getting `DEFAULT` (over the wire or from `in-process` evaluators) means that `targeting` rules exist, but returned no variant for this evaluation, so the value reverted "fell-through" to the `defaultVariant` (see https://flagd.dev/reference/flag-definitions/?h=reason#variants-returned-from-targeting-rules). This proposal extends this idea slightly, to allow for returning `DEFAULT` with _no variant_, conveying that targeting rules did not match, AND no `defaultVariant` was set in the config, meaning that the code default should be used.

The problem with evaluation.v1, is that in that proto, value/variant are not marked as optional, so our client's have no way to know the difference between an unset `value` and a zero-value `value`, meaning we would need a v2.

cc @marcozabel

docs/architecture-decisions/support-code-default.md
toddbaert: I think not, due to the reasons mentioned [here](https://github.com/open-feature/flagd/pull/1706#discussion_r2742484160).

toddbaert:
I'd just like to say that most of this is implemented already in a less-than-ideal way, with flags without defaults returning `FLAG_NOT_FOUND` as a hacky alternative to a real solution. This is already supported by the flag definition schema - the only breaking changes are in the .proto, and we can handle this gracefully by adding a v2 proto and supporting both in flagd for a duration.

I suggest we merge this next week unless there's any strong objections.

Issue #1481, opened 2024-12-18
[BUG] Inconsistent error behavior between Resolve<T> and ResolveAll during flagd runtime


cupofcat:
### Observed behavior

For targeting rules that can return different types based on the context (which does not conform to schema but is allowed by flagd with just a warning):
- `ResolveAll` checks whether the type of the evaluated variant matches the type of the `defaultVariant`
- `Resolve<T>` checks whether the type of the evaluated variant matches `<T>` (`defaultVariant` does not matter)

### Expected Behavior

I would expect the behavior to be consistent. If `defaultVariant` is used to infer the flag type, then `Resolve<T>` methods should also check whether `<T>` matches the type of `defaultVariant`  and not the type of the evaluated variant.

### Steps to reproduce

For the following flag config:
```
{
    "$schema": "https://flagd.dev/schema/v0/flags.json",
    "flags": {
      "is-enabled": {
        "defaultVariant": "off",
        "state": "ENABLED",
        "targeting": {
          "if": [
            {
              "<": [
                {
                  "%": [
                    {
                      "var": "request_id"
                    },
                    1000
                  ]
                },
                100
              ]
            },
            "on",
            "off"
          ]
        },
        "variants": {
          "on": 1,
          "off": false
        }
      }
    }
  }
```

flagd returns the following:
```
curl -X POST "http://localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" \
  -d '{"flagKey":"is-enabled","context":{"request_id": 42}}' -H "Content-Type: application/json"
{"code":"invalid_argument","message":"Type mismatch error"}%

curl -X POST "http://localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" \
  -d '{"flagKey":"is-enabled","context":{"request_id": 420}}' -H "Content-Type: application/json"
{"value":false, "reason":"TARGETING_MATCH", "variant":"off", "metadata":{}}%

curl -X POST "http://localhost:8013/flagd.evaluation.v1.Service/ResolveInt" \
  -d '{"flagKey":"is-enabled","context":{"request_id": 42}}' -H "Content-Type: application/json"
{"value":"1", "reason":"TARGETING_MATCH", "variant":"on", "metadata":{}}%

curl -X POST "http://localhost:8013/flagd.evaluation.v1.Service/ResolveInt" \
  -d '{"flagKey":"is-enabled","context":{"request_id": 420}}' -H "Content-Type: application/json"
[...]

PR #1746, opened 2025-08-18
chore(ADR): Proposal to add flag type

andreyturkov:
### Related Issues
https://github.com/open-feature/flagd/issues/1481

### DIFF
--- a/docs/architecture-decisions/flag-type.md
+++ b/docs/architecture-decisions/flag-type.md
@@ -38,7 +38,7 @@ This proposal is to extend the flag definition with an optional `flagType`proper
 
 By introducing an explicit `flagType`field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
 
-The new `flagdType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will fall back to infer the flag type from its variants. As the flag schema enforces that all variants are of the same type, the type of the first variant will be used. When the `flagdType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
+The new `flagType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will fall back to infer the flag type from its variants. As the flag schema enforces that all variants are of the same type, the type of the first variant will be used. When the `flagType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
 
 This change will make the behavior of `flagd` more predictable and reliable.
 


docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+## Background
+
+Currently, `flagd` has inconsistent behavior in type validation between its `Resolve<T>` and `ResolveAll` API methods. The `Resolve<T>` method validates the evaluated flag variant against the type `T` requested by the client, while `ResolveAll` validates it against the type of the `defaultVariant` specified in the flag's definition. This discrepancy can lead to situations where a flag evaluation succeeds with one method but fails with the other, depending on the evaluation context and the variant returned. This inconsistent behavior is further detailed in bug report #1481.
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself. Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`) , which is not always a reliable source of truth for all possible variants. This can lead to unexpected errors and make it difficult for developers to debug their feature flags.
+
+
+## Requirements
+
+* The new `type` field in the flag definition must be optional to ensure backward compatibility.
```
beeme1mr: I would prefer to use `flagType` so that it matches what we defined in the [flag manifest](https://github.com/open-feature/cli/blob/main/schema/v0/flag-manifest.json) used by the OpenFeature CLI.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+* If the `type` field is present, `flagd` must validate that all variants of the flag conform to this type during initialization.
```
beeme1mr:
```suggestion
* If the `type` field is present, `flagd` must validate that all variant values of the flag conform to this type during initialization.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+By introducing an explicit `type` field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
+
+The new `type` field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `type` field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
+
+This change will make the behavior of `flagd` more predictable and reliable.
+
+
+### API changes
+
+The `flagd` flag definition will be updated to include an optional `type` property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"number"`, and `"object"`.
```
beeme1mr: The OpenFeature CLI defines `float`, `integer`, `object`, `string`, and `boolean`. We should probably keep them in sync but I'd like to get others thoughts on this.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+
+# Extending Flag Definition with a Type Property
+
+
+## Background
+
+Currently, `flagd` has inconsistent behavior in type validation between its `Resolve<T>` and `ResolveAll` API methods. The `Resolve<T>` method validates the evaluated flag variant against the type `T` requested by the client, while `ResolveAll` validates it against the type of the `defaultVariant` specified in the flag's definition. This discrepancy can lead to situations where a flag evaluation succeeds with one method but fails with the other, depending on the evaluation context and the variant returned. This inconsistent behavior is further detailed in bug report #1481.
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself. Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`) , which is not always a reliable source of truth for all possible variants. This can lead to unexpected errors and make it difficult for developers to debug their feature flags.
+
```
tangenti, line 16: Could you mention the recent ADR for supporting code default? This essentially makes it impossible to always infer the type from `defaultVariant`.

docs/architecture-decisions/flag-type.md
tangenti: +1. This is also what's defined in the manifest schema.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+#### The good
+* It improves the reliability and predictability of flag evaluations.
+* It allows for early error detection of type mismatches.
+* It improves the developer experience by making the API more explicit.
+
+#### The bad
+* It adds a new field to the flag definition, which developers need to be aware of.
+* It requires updating all `flagd` SDKs to support the new field.
+* It requires updating flag manifest schema
```
tangenti: Is this required?

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Timeline
+
+* **Phase 1: Core Implementation**
+    * Update the `flagd` core to support the new `type` field.
+    * Implement the type validation logic.
+    * Update the JSON schema.
```
tangenti, line 133: I'd expect the schema change happening first.

docs/architecture-decisions/flag-type.md
andreyturkov: done

docs/architecture-decisions/flag-type.md
andreyturkov: done

docs/architecture-decisions/flag-type.md
andreyturkov: done

docs/architecture-decisions/flag-type.md
andreyturkov: done

docs/architecture-decisions/flag-type.md
andreyturkov: Nope, removed

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself. Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`) , which is not always a reliable source of truth for all possible variants. This problem is getting worse by the planned support for code-defined defaults (as detailed in the [Support Code Default ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/support-code-default.md)), which allows the `defaultVariant` to be `null`. This makes it impossible to resolve the flag's type from the `defaultVariant`, increasing the risk of runtime errors.
+
+
+## Requirements
+
+* The new `flagdType`field in the flag definition must be optional to ensure backward compatibility.
```
beeme1mr:
```suggestion
* The new `flagdType` field in the flag definition must be optional to ensure backward compatibility.
* If the `flagdType` field is present, `flagd` must validate that all variant values of the flag conform to this type during initialization.
* Type mismatches found during initialization must be reported as errors.
* The `Resolve<T>` and `ResolveAll` methods must use the `flagdType` field for validation when it is available.
* The implementation must be consistent with the OpenFeature specification and the flag manifest schema.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Considered Options
+
+* **Consistent `defaultVariant` Validation:** Align the behavior of `Resolve<T>` with `ResolveAll` by making `Resolve<T>` validate the evaluated variant against the type of the `defaultVariant`.
+* **API Extension with Explicit Flag Type:** Introduce an optional `flagdType`property to the flag definition to serve as the single source for type validation.
+
+
+## Proposal
+
+This proposal is to extend the flag definition with an optional `flagdType`property. This approach is chosen over simply aligning the `Resolve<T>` and `ResolveAll` validation because it addresses the root cause of the type inconsistency and provides a more robust, long-term solution.
```
beeme1mr:
```suggestion
This proposal is to extend the flag definition with an optional `flagType`property. This approach is chosen over simply aligning the `Resolve<T>` and `ResolveAll` validation because it addresses the root cause of the type inconsistency and provides a more robust, long-term solution.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+By introducing an explicit `flagdType`field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
```
beeme1mr:
```suggestion
By introducing an explicit `flagType` field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+The new `flagdType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `flagdType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
```
beeme1mr:
```suggestion
The new `flagType` field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `flagType` field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+This change will make the behavior of `flagd` more predictable and reliable.
+
+
+### API changes
+
+The `flagd` flag definition will be updated to include an optional `flagdType`property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"integer"`, `"float"`, and `"object"`. This aligns with the OpenFeature CLI and the flag manifest schema.
```
beeme1mr:
```suggestion
The `flagd` flag definition will be updated to include an optional `flagType`property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"integer"`, `"float"`, and `"object"`. This aligns with the OpenFeature CLI and the flag manifest schema.
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+The following changes will be made to the `schemas/json/flags.json` file:
+
+1.  A new `flagdType`property will be added to the `flag` definition:
+
+```json
+"flag": {
+  "type": "object",
+  "properties": {
+    "type": {
```
beeme1mr:
```suggestion
    "flagType": {
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+### API changes
+
+The `flagd` flag definition will be updated to include an optional `flagdType`property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"integer"`, `"float"`, and `"object"`. This aligns with the OpenFeature CLI and the flag manifest schema.
+
+#### JSON Schema
+
+The following changes will be made to the `schemas/json/flags.json` file:
+
+1.  A new `flagdType`property will be added to the `flag` definition:
```
beeme1mr:
```suggestion
1.  A new `flagType` property will be added to the `flag` definition:
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+### Consequences
+
+#### The good
+  * It improves the reliability and predictability of flag evaluations.
+  * It allows for early error detection of type mismatches.
+  * It improves the developer experience by making the API more explicit.
+
+#### The bad
+  * It adds a new field to the flag definition, which developers need to be aware of.
+  * It requires updating all `flagd` SDKs to support the new field.
```
beeme1mr: It may be worth exploring this impact a bit further. The only change I can think of would be that the SDKs could use this new property to validate that a configuration is valid. I don't think it would impact the evaluations since in-process would already know the requested type and RPC would be handled by the flagd changes.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+    },
+    "state": {
+      ...
+    },
+    ...
+  }
+}
+```
+
+2.  The `booleanFlag`, `stringFlag`, `integerFlag`, `floatFlag`, and `objectFlag` definitions will be updated to enforce the `flagdType`property:
```
beeme1mr:
```suggestion
2.  The `booleanFlag`, `stringFlag`, `integerFlag`, `floatFlag`, and `objectFlag` definitions will be updated to enforce the `flagType`property:
```

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself. Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`) , which is not always a reliable source of truth for all possible variants. This problem is getting worse by the planned support for code-defined defaults (as detailed in the [Support Code Default ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/support-code-default.md)), which allows the `defaultVariant` to be `null`. This makes it impossible to resolve the flag's type from the `defaultVariant`, increasing the risk of runtime errors.
```
guidobrei: > not always a reliable source of truth for all possible variants

I'm wondering if this is still the case. It was mentioned in #1481 by @cupofcat, where a flag contained mixed typed variants. However, if I read the current [flagd schema](https://github.com/open-feature/flagd-schemas/blob/main/json/flags.json#L112) correctly, this cannot happen anymore.
An exception might be a numeric flag that cannot distinguish between int and float.

guidobrei: > This makes it impossible to resolve the flag's type from the `defaultVariant`

True, but the `defaultVariant` is a reference to one of the named variants. They all have the same type so we can pick any of them.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## Requirements
+
+* The new `flagdType`field in the flag definition must be optional to ensure backward compatibility.
+* If the `flagdType`field is present, `flagd` must validate that all variant values of the flag conform to this type during initialization.
+* Type mismatches found during initialization must be reported as errors.
+* The `Resolve<T>` and `ResolveAll` methods must use the `flagdType`field for validation when it is available.
+* The implementation must be consistent with the OpenFeature specification and the flag manifest schema.
```
guidobrei, line 28: > and the flag manifest schema

Is  the OF CLI manifest schema meant?

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself. Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`) , which is not always a reliable source of truth for all possible variants. This problem is getting worse by the planned support for code-defined defaults (as detailed in the [Support Code Default ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/support-code-default.md)), which allows the `defaultVariant` to be `null`. This makes it impossible to resolve the flag's type from the `defaultVariant`, increasing the risk of runtime errors.
+
+
+## Requirements
+
+* The new `flagdType`field in the flag definition must be optional to ensure backward compatibility.
+* If the `flagdType`field is present, `flagd` must validate that all variant values of the flag conform to this type during initialization.
```
guidobrei: This is already validated with the schema.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,130 @@
...
+* Type mismatches found during initialization must be reported as errors.
+* The `Resolve<T>` and `ResolveAll` methods must use the `flagdType`field for validation when it is available.
```
guidobrei: What is the expected behavior? How does this behavior relate to flagd providers? 

Providers currently try to convert the returned value to the requested evaluation type. They return an evaluation error when conversion between resolved and requested type is not supported by the underlying type system.

You can for example, call `evaluateStringFlag(...)` on a boolean flag and if the boolean value can be converted to a String, it will do so.
Do we expect with this change, that the before mentioned example should now return an evaluation error in flagd? If yes, this behavior should be consistent in flagd providers, so there is no behavioral difference between RPC and inProcess evaluation modes.

What do we expect when calling `Resolve<Float>` on a flag with `flagType: "integer"`. Should this fail?

docs/architecture-decisions/flag-type.md
guidobrei: > If the field is omitted, `flagd` will treat the flag as having `object`

For backwards compatibility, I propose fall back to the current behavior of trying to infer the flag type. With one amendment: Infer it from one of the `variants` items instead of the `defaultVariant` which can now be optional.

guidobrei (commented):
All in all, I appreciate the proposal of being explicit in the flag definition.
However, after reviewing the flag schema, I am not sure if the original issue #1481 it tries to solve is still present.

One important thing to clarify is what validation of the introduced flag type in evaluation methods mean:
* We validate that requested type and flag type match exactly
* We validate that requested type and flag type can be converted

docs/architecture-decisions/flag-type.md
tangenti: This is only being validated on the schema level and flagd binary/provider does not know the exact type of the schema validation error.

docs/architecture-decisions/flag-type.md
tangenti: +1. Probably better to infer from `defaultVariant` and set the inferred type in the flag store.

andreyturkov: That's why this ADR appeared - there are no single source of true of what is the expected behavior of Resolve methods with different scenarios.

Based on current schema and implementation flag is being validated against schema and if variants don't conform with type, i.e. they have different types, the validation is just throws **warning** (**not error**) and flagd binary/provider does not know the exact type of the schema validation error.

It means current schema allows user to have flags with different types.

The main idea of this ADR is to set explicitly which type should be expected and used for validation and resolcing the variant. 
But I understand the concern regarding the case when the type is ommited. 

I envision 3 oprions:
1. Fall back to infer type from variants (proposed by  @guidobrei) - but what if variants have different types - it's not obviouse which of them needs to be taken into resolving
2. Fall back to infer from defaultVariant  [current implementation] (proposed by @tangenti ) - it has 2 issues: first, when it's null, second, inconsistency between ResolveAll and Resolve<T> in bug #1481.
3. Proposed option in this ADR is to treat such flags as `object` - it changes the existing behavior, but the behavior anyway needs to be fixed eigher by this ADR or by fixing the bug #1481. And IMO this approach is more transparent: if you have type defined for your flag - it will check variants against it, otherwise we allow any types.

cc @beeme1mr @toddbaert

docs/architecture-decisions/flag-type.md
andreyturkov: yes

docs/architecture-decisions/flag-type.md
andreyturkov: this is still the case. see comment below https://github.com/open-feature/flagd/pull/1746#discussion_r2291108766

docs/architecture-decisions/flag-type.md
toddbaert: I think `1.` should work... because unless I'm missing something, it's not possible for them to have different types if the schema is valid. See [here](https://www.jsonschemavalidator.net/s/jeI0IH0U).

If I'm not mistaken or missing something, we can just use whatever the type of the first variant is.

docs/architecture-decisions/flag-type.md
toddbaert: I hope I'm not missing something [here](https://github.com/open-feature/flagd/pull/1746#discussion_r2291996059).

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,127 @@
...
+By introducing an explicit `flagType`field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
+
+The new `flagType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `flagType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
+
+This change will make the behavior of `flagd` more predictable and reliable.
+
+
+### API changes
+
+The `flagd` flag definition will be updated to include an optional `flagType`property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"integer"`, `"float"`, and `"object"`. This aligns with the OpenFeature CLI and the flag manifest schema.
```
toddbaert: I'm not 100% sure it's a good idea to differentiate between float/int for this purpose. A couple reasons:

- I think we could easily write a JSON schema to enforce that flags with `"flagType":boolean` only have variants of type boolean, and `"flagType":string` flags only have variants of type string, etc, etc... but I don't think we can enforce that ints are ints and not floats (we might be able to do this with [multipleOf](https://json-schema.org/understanding-json-schema/reference/numeric#multiples) but I'd like to see a PoC first).
- Our providers don't care about float/int as it is; they convert instead of indicating a type error.
- JSON itself doesn't distinguish between numeric types, so attempting to do so might be fraught.

Very interested in other thoughts on this.

beeme1mr: The main reason to have both `float` and `int` is to match the flag manifest used in the OpenFeature cli. I think it's important to keep them in sync. However, we may need to support casting from at least int to float.

tangenti: > Our providers don't care about float/int as it is; they convert instead of indicating a type error.

I see the type system of JSONLogic as a limitation, as `Number` is essentially an IEEE 754 Double.

Conversions between Double and Integer64 may lose precision in both directions. I'd like to see a stronger typing is enforced if possible (depending on the language), to prevent the case that the client gets a different value than what's put in the evaluation rules, which can be quite dangerous for some situations of using `Number` flags, for example, when the flag represents a GCP project number

toddbaert: Do you think this is something we need to enhance the OpenFeature specification to support?

IMO, I think numeric precision is a fuzzy part of the spec, mostly because it's non-trivial to write language-neutral specification about that sort of thing.

beeme1mr (approved):

docs/architecture-decisions/flag-type.md
tangenti: I don't have enough contexts and knowledge to form a mature opinion for OpenFeature spec.... I guess we could explore the extension in flagd and OFREP first and we'll learn more about the feasibility. For now, I think drawing the boundary for each type in the flagd can be helpful.

docs/architecture-decisions/flag-type.md
andreyturkov: Added a section about float/integer types.

docs/architecture-decisions/flag-type.md
andreyturkov: Added a section about float/integer types.

andreyturkov:
@tangenti, @toddbaert, @guidobrei could you please take a look the latest version of this ADR?

docs/architecture-decisions/flag-type.md
```diff
@@ -12,45 +12,44 @@ updated: 2025-08-14
...
+By introducing an explicit `flagdType`field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
 
-The new `type` field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `type` field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
+The new `flagdType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will treat the flag as having `object`, and no type validation will be performed against the `defaultVariant`. When the `flagdType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
```
tangenti: Could you add more details about how the validation would work? 

I'm particularly interested in the cases that the literal values set to the variants do not fit with the flag type.  Currently flagd seems to just cast the value to the type, **with** precision lost (this is perhaps more related to the JSONLogic engine impl). 

CC @toddbaert

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,144 @@
...
+
+The root cause of this issue is the absence of a dedicated, authoritative type definition for the flag itself.
+Instead, the type is inferred from the `defaultVariant` or API itself (`T` from `Resolve<T>`), which is not always a reliable source of truth for all possible variants.
+This problem is getting worse by the planned support for code-defined defaults (as detailed in the [Support Code Default ADR](https://github.com/open-feature/flagd/blob/main/docs/architecture-decisions/support-code-default.md)), which allows the `defaultVariant` to be `null`.
+This makes it impossible to resolve the flag's type from the `defaultVariant`, increasing the risk of runtime errors.
+
+## Requirements
+
+* The new `flagType`field in the flag definition must be optional to ensure backward compatibility.
+* If the `flagType`field is present, `flagd` must validate that all variant values of the flag conform to this type during initialization.
```
toddbaert, line 25: I _believe_ we could even validate this in the JSON schema with some backwards compatible changes to the definition: https://flagd.dev/schema/v0/flags.json

toddbaert (commented):
This seems reasonable, I have a concern [here](https://github.com/open-feature/flagd/pull/1746#discussion_r2445878680) before I can approve.

docs/architecture-decisions/flag-type.md
```diff
@@ -0,0 +1,144 @@
...
+
+By introducing an explicit `flagType`field, it establishes a single source of truth for the flag's type, independent of its variants. This allows for early and consistent type validation during flag definition parsing, preventing type-related errors at runtime.
+
+The new `flagType`field will be optional to maintain backward compatibility with existing flag configurations. If the field is omitted, `flagd` will fall back to infer the flag type from its variants. As the flag schema enforces that all variants are of the same type, the type of the first variant will be used. When the `flagType`field is present, `flagd` will enforce that all variants of the flag conform to the specified type.
+
+This change will make the behavior of `flagd` more predictable and reliable.
+
+### API changes
+
+The `flagd` flag definition will be updated to include an optional `flagType`property. This property will be a string enum with the following possible values: `"boolean"`, `"string"`, `"integer"`, `"float"`, and `"object"`.
```
toddbaert, line 47: I'm not sure about the differentiation between `integer` and `float` here. I think that we might not want to discriminate this way. Justifications:

- we already convert float/int in providers with static numeric types: https://github.com/open-feature/java-sdk-contrib/blob/0b720f3d482a2a6871614fd85bab2002f2c9df66/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java#L221-L227
- we already just consider flags "numeric" generally (not float/int) in our schema: https://flagd.dev/schema/v0/flags.json

So think just considering flags numeric and converting automatically will work (with the expected precision loss) and that this will be fairly intuitive to users.

One thing to note, I don't think we have robust e2e testing around this in particular (conversion between numeric types), or I couldn't find it. @aepfli cc'ing you because you also know a lot about our test suite.

aepfli: we do not have tests currently for cross conversion, but generally i think we should just stick with number and provide the most generic type to the users. We maybe want to provide convenience methods to convert to float, int, ... but imho, i think just providing a Number would be the best option.

andreyturkov:
Closed PR due to lack of time and getting outdated. Will reopen once have updates on this.

PR #1747, opened 2025-08-20
refactor: removed unused Selector from Flag and Store.

andreyturkov:
## This PR

This pull request removes the unused `Selector` field from the `Flag` struct and deletes the `SourceDetails` struct, which is no longer in use.

These changes are part of a cleanup effort following the major refactoring in PR #1702, which made these components obsolete.

### Related Issues

Follow-up to #1702

### How to test

All existing tests should continue to pass.


toddbaert (approved):
This one is very small / safe so I'm going to merge with only my approval.

Once again I'll plug [this](https://github.com/open-feature/flagd/pull/1705#issuecomment-3211282080).

Issue #1737, opened 2025-08-14
[FEAT] Harden Hashing Consistency And Add Support For Non-string Attributes in Fractional Evaluation


cupofcat:
- [ ] Review the current Gherkin suite to ensure any potential inconsistencies are captured
- [ ] Support non-string randomisation units (e.g. bytes) in fractional without casting
- [ ] Ensure consistency of hashing and bucketing across all providers (languages), no matter the internal encoding choices made by compilers and platforms

Out of scope for this issue:
- Bucketing improvements
- Any new advanced features

cupofcat:
Can I assign this to myself?

Today, when evaluating the `fractional` targeting rule, if the first argument inside `fractional` evaluates to a string, it's used as the hashing input, and, subsequently as bucketing value. Otherwise, the evaluation context is searched for the "targetingKey" property, treated as string, string-concatenated with flagKey and that is used as hashing input instead.

In both cases, the input is assumed to be a string. As [mentioned](https://github.com/open-feature/flagd/issues/1622#issuecomment-2862427286) by @tangenti, this is problematic :

> Consistency of casting and encoding across all supported languages can be a painful problem.. Hashing the bytes directly don't have this problem.

I think we can solve it like this:
- no changes to the schema
- if someone wants to use bytes, they just use "var" directly, or, they can also use "targetingKey"
- providers, when evaluating, they check the type of var / targetingKey; this should work as, I think, all our providers today are strongly typed
- if the type is byte we call murmur's Sum32 method directly rather than StringSum32

I would love to hear opinions about the user experience. One challenge I see is this: Today, the suggested best practice, when not using targetingKey, is to use "cat" and join "flagKey" with "var". However, cat outputs a string (at least in some languages). If using "var" without "cat", the user would have to make sure that "flagKey" is included in "var" already.  An alternative could be using merge, which outputs a list, and treat that specially in the implementation.

beeme1mr:
> I would love to hear opinions about the user experience. One challenge I see is this: Today, the suggested best practice, when not using targetingKey, is to use "cat" and join "flagKey" with "var". However, cat outputs a string (at least in some languages). If using "var" without "cat", the user would have to make sure that "flagKey" is included in "var" already. An alternative could be using merge, which outputs a list, and treat that specially in the implementation.

[...]

PR #1783, opened 2025-08-21
chore(ADR): Create fractional-non-string-rand-units.md

cupofcat:
## This PR
- Proposes `Support Non-String Inputs for Fractional Bucketing` ADR

### Related Issues
https://github.com/open-feature/flagd/issues/1737


docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,86 @@
...
+## Proposal
+
+We will modify the evaluation logic for the `fractional` operator.
+
+When evaluating the first element of the `fractional` array:
+
+1. **If the resolved value is a string:** The existing logic will be preserved. Providers will continue to use their respective language's MurmurHash3 `StringSumN` (or equivalent) function. This guarantees backward compatibility.
+2. **If the resolved value is a non-string type (e.g., integer, float, boolean):** Providers will hash the standard byte-representation of the value using a MurmurHash3 `SumN` (or equivalent) function.
+
+This approach relies on the strong typing within our providers (e.g., the JS provider uses TypeScript, Go is strongly typed), allowing them to reliably distinguish between string and non-string types at evaluation time.
```
beeme1mr: I'd expect the type to be checked at runtime using an assertion.

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,86 @@
...
+* Good, because it avoids unnecessary casting
+* Bad, because strings will still rely on specific of string encoding for the provider language and use `"StringSumN"` methods
+
+### Timeline
+
+Pre flagd 1.0.
+
+### Open questions
+
+* Should we also treat strings in the same way as other types? (this might result in rebucketing after the change is launched)
```
beeme1mr: Perhaps we could schedule a breaking change at the same time we update the bucket size.

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,86 @@
...
+* Should we ensure all the providers use prescribed string encoding for all strings (e.g. UTF8)?
+* Should we also change the behavior when `targetingKey` is used?
```
beeme1mr: What do you have in mind?

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,86 @@
...
+
+## More Information
+
+Today, flagd recommends to salt the variable with flagKey directly in the `fractional` logic, using the `"cat"` operator. This will not be possible for non-string types. As a separate ADR we can consider an improvement where either a new custom operator or some existing operator can be used to salt variables directly in `fractional` logic, without casting to strings.
```
beeme1mr: This is an important callout and something we should definitely address. It would lead to very confusing behavior if the flag configuration expects context to be set, but it isn't for some reason. This would lead to every request being put into the same bucket.

beeme1mr (approved):

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,86 @@
...
+
+Option 1 was chosen for its ergonomics and zero-impact on existing schemas. Option 2 adds unnecessary complexity to the flag definition language, and Option 3 creates confusing and non-obvious semantics.
+
+## Proposal
+
+We will modify the evaluation logic for the `fractional` operator.
+
+When evaluating the first element of the `fractional` array:
+1.  **If the resolved value is a string:** The existing logic will be preserved. Providers will continue to use their respective language's MurmurHash3 `StringSumN` (or equivalent) function. This guarantees backward compatibility.
+2.  **If the resolved value is a non-string type (e.g., integer, float, boolean):** Providers will hash the standard byte-representation of the value using a MurmurHash3 `SumN` (or equivalent) function.
```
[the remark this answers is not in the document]
cupofcat: How about we use https://cbor.io/ for consistent serialization? Here's Gemini's discussion of how that would look like: https://g.co/gemini/share/3525b8964ada

cupofcat:
I've been thinking about this more and I wonder WDYT about this approach:

We take this pre-1.0 opportunity to introduce a breaking change for how hashing / bucketing works:
- If fractional arguments are all arrays, then we use `targetingKey` for hashing, similar to today
- Otherwise, the first argument needs to be an object (`{}`)

If the first argument is an object, it's evaluated and the evaluated value is passed to CBOR for marshalling into a consistent byte array. That byte array is passed to MurMur3 for hashing.

If `targetingKey` is used, a 2 element array is created with the first element being the flag key, and the second element being the `targetingKey` value. That array is passed to CBOR for marshalling and the resulting byte array is passed to MurMur3 for hashing.

CBOR libraries in each language ensure that the same values get the same byte encoding and murmur3 libraries will ensure that same byte arrays will get the same hash. This works across any type, even for strings.

That way we have langauge-agnostic, stable, and consistent bucketing.

CBOR is specified in an Internet Standard RFC by IETF, which should mean this stays stable for foreseeable future.

beeme1mr:
> We take this pre-1.0 opportunity to introduce a breaking change for how hashing / bucketing works:
> 
> * If fractional arguments are all arrays, then we use `targetingKey` for hashing, similar to today
> * Otherwise, the first argument needs to be an object (`{}`)

@cupofcat would you mind showing an example configuration of what you had in mind? When you say object, do you mean referencing a context attribute `{ "var": "useId" }` or something else?

cupofcat:
> > We take this pre-1.0 opportunity to introduce a breaking change for how hashing / bucketing works:
> > 
> > * If fractional arguments are all arrays, then we use `targetingKey` for hashing, similar to today
> > * Otherwise, the first argument needs to be an object (`{}`)
> 
> @cupofcat would you mind showing an example configuration of what you had in mind? When you say object, do you mean referencing a context attribute `{ "var": "useId" }` or something else?

<del>I meant that essentially the first item in fractional is {...} and not a list. It can be { var: ...} or something else. I'm open to other suggestions, we just need some way to detect if we should use targetingKey or evaluate the first argument for hashing input.</del>

<del>I would want to keep it pretty generic though. In theory, a list of elements could also be a valid input to hashing.</del>

Edit: The above did not make sense, sorry.

@beeme1mr I've been thinking about this more and I think we have the following options:

1. Similarly to today we have an optional 1st argument to fractional to provide value to hash for bucketing.
2. We make the first argument mandatory; if someone wants to use `targetingKey` they need to provide `null` as first argument
3. We introduce a new custom operator to be used inside `fractional`
A. something like `cbor-8949-bytes` that is guaranteed to return a consistent byte array in every language
B. something like `object` that preserves the keys and evaluates the values
C. we can brainstorm other custom operator with other semantics too

For (1), since the first argument is optional, we need a way to distinguish it from the subsequent ones. Today, this is done based on type (if it casts to a string, it's treated specially). That means the first argument, if the user wants to use it for bucketing, cannot evaluate to an array. If it is an array we cannot distinguish it from the rest of `fractional` values. We could go further, and look into types inside the array and say it cannot start with [string, number] but overall I think this is a really hacky semantic. We could say, that if it evaluates to a map/dictionary/object then we encode it to bytes and hash that. However, I cannot seem to figure out how to return an object in JSONLogic using existing operators...

(2) is much clearer, but the semantic of the current schema changes (the schema itself can stay the same). We make the first argument mandatory as the bucketing input. If it's `null` we use the `targetingKey`. This is a pretty big breaking change, as all the flag configs that rely on `targetingKey` would break. But, perhaps, this is OK when moving to 1.0.

(3A) Similarly to today, if the first value in fractional array is bytes we hash that. If not, we use the current behavior as is.

```json
"fractional": [
  {
    "cbor-8949-bytes": [{"var" : "$flagd.flagKey"}, {"var" : "some-var"}]
  },
  ["a", 50], ...
]
```

(3B) Similarly to today, if the first value in fractional array is a map/dict we hash that. If not, we use the current behavior as is.

```json
"fractional": [
  {
    "object": {
      "comment" : "Any keys are valid, we will just translate all of this to bytes (JSONLogic would still evaluate the values)",
      "key1" : {"var" : "$flagd.flagKey"},
      "key2" : {"var" : "some-var"},
      "extra-salt" : "himalayan"
    }
  },
  ["a", 50], ...
]
```

If none of the above works, perhaps there is also:

4. We introduce a new custom operator replacing `fractional`. We can call it `fractional1` or `fractionalWithKey` or `fractionalBytes` or something. We make it officially recommended for flagd 1.0 and we deprecate `fractional`. We could keep evaluating `fractional` using current logic with warnings. We could also remove it completely from 1.0, given that this will be a pretty big breaking change anyway for the users.

WDYT?

P.S. What I find complicates this exercise is that today we send the entire contents of `targeting` to JSONLogic for evaluation. So we cannot really do any parsing on the structure of the object. Our `fractional` evaluation implementation already receives evaluated values so we don't know what the arguments looked like.

So I guess there is also:

5. Parse either `targeting` or `fractional` looking for specific fields to control the behavior of what gets passed to JSONLogic and how.

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: Please see my comment below. I think this is becoming pretty complex to solve without a big breaking change.

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: Not treat it as string, but also allow all types and use consistent bytes encodings as hashing input.

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: Yes. Or we just launch this in 1.0

cupofcat:
Based on the discussions on Slack I updated this ADR to focus on hashing improvements in the existing fractional only.

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+
+* **Inconsistent Hashing:** Different providers (Go, PHP, Java) may encode the same string into bytes differently (e.g., UTF-8 vs UTF-16). Since hashing functions like MurmurHash3 operate on bytes, this leads to different hash results and thus different bucket assignments for the same logical input across platforms.
+* **Unnecessary Coercion:** If a user wishes to bucket based on a numeric ID (e.g., `userId: 12345`), they must first explicitly cast it to a string (`"12345"`) within the flag definition using an operator like `"cat"`.
+
+This proposal seeks to resolve these issues by allowing `fractional` to operate directly on the byte representation of non-string inputs and to explicitly encode values to bytes with deterministic encoders.
+
+## Requirements
+
+* Users must be able to use both string and non-string variables (e.g., integers, booleans) as the primary input for `fractional` evaluation.
+* Same "value" (e.g. 57.2, "some text", true, etc) should result in the same bucket assignment no matter the language of the provider and platform used.
```
tangenti: It's probably more accurate to use "same input" here, as literals can be encoded into different bytes depending on the type. For example, 42 as an int32 and 42 as a double.

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Non-requirements
+
+* This change does not need to be backward-compatible. Nevertheless, care should be taken to minimize disrputions - strings should be encoded using UTF-8 (already a default on most platforms and providers).
+* Support advnaced features like salting non-string types in JSON directly (that will be a separate ADR)
```
tangenti: typo:advanced

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+Option 1 was chosen for its ergonomics and zero-impact on existing schemas. Option 2 adds unnecessary complexity to the flag definition language, and Option 3 creates confusing and non-obvious semantics.
+
+## Proposal
+
+We will modify the evaluation logic for the `fractional` operator.
+
+When inspecting the first element of the `fractional` array:
+
+1. If the first element in `fractional` evaluates to a non-array type then deterministically encode it to a well defined byte array (using UTF-8 for strings) and hash the bytes.
+2. Otherwise, if `targetingKey` is a string, concatenate `flagKey` and `targetingKey`, encode to UTF-8 and hash that (current behavior).
```
tangenti: What if `targetKey` is not set in the context?

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+3. Otherwise, if `targetingKey` is non-string, create a 2 element array of [`flagKey`, `targetingKey`] and hash that.
```
tangenti: This is conflict with the OF spec: https://openfeature.dev/specification/glossary#targeting-key

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Considered Options
+
+1. **Proposed:** *Type-Aware Hashing:* Extend the current behavior to support non-string types as first arguments to `fractional` and in `targetingKey`.
+2. *New Operator:* Introduce a new operator, such as `"bytesVar"`, to explicitly signal that the variable's raw bytes should be hashed.
+3. *Operator Overloading:* Reuse an existing operator (e.g., `"merge"`) or structure (e.g., providing a list) to imply byte-based hashing.
+
+Option 1 was chosen for its ergonomics and zero-impact on existing schemas. Option 2 adds unnecessary complexity to the flag definition language, and Option 3 creates confusing and non-obvious semantics.
+
+## Proposal
```
tangenti, line 83: Could you explicitly list the types of context properties that will be supported? https://openfeature.dev/specification/sections/evaluation-context#requirement-312

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: We are getting into the realm of implementation details here but I think both approaches are valid. E.g., I think, CBOR first packs the number into smallest amount of bits and then encodes that. So 42 would always be treated as int32.

This might be important for languages like Python which, I think, does not necessarily have standard types like int32, etc.

Generally, I would say the outcome should be that we differentiate between integers (42) and float (42.0) but not between int32/int64, etc.

With CBOR there are further standards that could be picked (e.g. 42 and 42.0 would also get the same encoding).

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: I think we keep the current behavior, which is an error and the default gets returned.

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: Oh, thanks. So if `targetingKey` can only be a string then I guess I will update the ADR such that if it's not a string, we keep the current behavior (error + return default).

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,129 @@
...
+
+We will modify the evaluation logic for the `fractional` operator.
+
+When inspecting the first element of the `fractional` array:
+
+1. If the first element in `fractional` evaluates to a non-array type then deterministically encode it to a well defined byte array (using UTF-8 for strings) and hash the bytes.
```
toddbaert: > non-array

Should we also say "non-object"? As in, it has to be a string or "primitive" type? Or will we try to serialize objects somehow?

cupofcat: Today, if the first argument is an object with 1 key, JSONLogic evaluates it first (that's how "var" or "cat" work). If the object has more than 1 key, I think we should treat it as a map-like type in the given language and encode it in the consistent way.

toddbaert: > then deterministically encode it to a well defined byte array

Does this present some challenges specifically with numbers? Different implementations will store numbers differently, so while we can easily say "UTF-8" for strings, don't we have to mention some byte layout for numbers as well? We would want `-127` to behave the same in Java/JS/C#/Go/etc...

We could specify a particular representation (ie: https://en.wikipedia.org/wiki/IEEE_754), etc, but we'd have to do that for both floats and ints, and I'm worried about how complex this might be, and more importantly, how possible it might be in higher level languages with less strength around number precision, like JS.

**_It might be easier to simply stringify them first, and then follow the string encoding?_** If you agree, I think we should specifically mention this.

cupofcat: Yes. Do we want to specify the bytes layouts for numbers explicitly in this ADR though? Or leave it as implementation detail, as long as it's consistent?

I feel strongly that CBOR is the right choice as it deals with all of the complexities of consistent hashing and is an Internet Standard.

Perhaps we should just prescribe that as part of this ADR?

https://www.rfc-editor.org/rfc/rfc8949.html

cupofcat:
I updated the ADR to explicitly require CBOR encodings + some other minor improvements based on feedback.

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,167 @@
...
+* Good, because it avoids unnecessary casting.
+* Bad, because all of the users will experience rebucketing.
+
+### Timeline
+
+Prior to flagd 1.0 launch.
+
+## More Information
+
+Today, flagd recommends salting the variable with flagKey directly in the `fractional` logic, using the `"cat"` operator. This will not be possible for non-string types. Advanced features like that will be considered in a separate ADR.
```
cupofcat, line 163: This is the thing I dislike the most about this ADR. I've been coming back to this over and over but I cannot figure out a good solution that is both compatible with the current semantics of the schema and allows for salting non-strings.

I think one "hack" we could do is to allow arrays of size 1, and 3+ in the first position but not size 2. Or, if the array evaluates to 2-elements array of [string, integer] then we treat is as bucket and, in all other cases, we treat it as input to hashing. But I don't like that either.

Another way would be to introduce a new custom operator called `merge-wrap` (or similar), which behaves similarly to `merge` but returns an object and not an array:

```json
"fractional": [
  {
    "merge-wrap" : [
      {"var" : "$flagd.flagKey"},
      {"var" : "userId"}
    ]
  },
  ["a", 50], ...
]

// Would evaluate to something like:
"fractional": [
  {
    "targetingKey" : ["some-flag-key", 12345]
  },
  ["a", 50], ...
]
```

and we could then unwrap that inside fractional. That is also hacky.

docs/architecture-decisions/fractional-non-string-rand-units.md
toddbaert: > Yes. Do we want to specify the bytes layouts for numbers explicitly in this ADR though? Or leave it as implementation detail, as long as it's consistent?
> 
> I feel strongly that CBOR is the right choice as it deals with all of the complexities of consistent hashing and is an Internet Standard.
> 
> Perhaps we should just prescribe that as part of this ADR?
> 
> https://www.rfc-editor.org/rfc/rfc8949.html

I think this is probably the place to decide. If we don't decide here, we'll just be having a similar discussion with less formality in some PR or issue.

I have some concerns about the feasibility of hashing non-strings, especially maps/objects. CBOR sounds promising, but I'm worried about things like key ordering and null/undefined keys and such minutia really giving us a hard time and making hashing unpredictable. Even if CBOR prescribes such things, I'm worried about the input to the CBOR functions being inconsistent for people in ways they won't consider.

toddbaert: Yes, sorry, I meant if the value retrieved, not the JSONLogic expression. See: https://github.com/open-feature/flagd/pull/1783#discussion_r2318212308

toddbaert: It looks like the most rigorous CBOR variants support deterministic ordering: https://datatracker.ietf.org/doc/html/rfc8949#name-length-first-map-key-orderi

cupofcat: In the updated ADR I suggested we use [4.2.1. Core Deterministic Encoding Requirements](https://www.rfc-editor.org/rfc/rfc8949.html#section-4.2.1) rather than [4.2.3](https://www.rfc-editor.org/rfc/rfc8949.html#name-length-first-map-key-orderi), but yeah, there are specifications.

PTAL at the current ADR text.

cupofcat: Also, I think this is precisely why we should expand and harden the test suite. If we do they work of considering the edge cases of inputs/outputs then the implementers won't have to consider those, hopefully :)

docs/architecture-decisions/fractional-non-string-rand-units.md
toddbaert: Another option is to shift this more toward the user, and remove the current salting. With the additional flexibility of non-strings, our previous assumptions simply are not valid. We could just say it's up to users now to salt flag keys themselves.

To be honest, I think this is worth figuring out in this PR as well.

toddbaert:
I am in favor of this proposal, but I can't help but consider [this](https://github.com/open-feature/flagd/pull/1783#discussion_r2316738882) a bit of a blocker.

@beeme1mr WDYT about this one? If we cant append the flag key, should we remove this in the string case? I'd prefer not to have string as a "special case".

cupofcat:
> I am in favor of this proposal, but I can't help but consider [this](https://github.com/open-feature/flagd/pull/1783#discussion_r2316738882) a bit of a blocker.
> 
> @beeme1mr WDYT about this one? If we cant append the flag key, should we remove this in the string case? I'd prefer not to have string as a "special case".

I would not necessarily consider this a special case for strings. It's a result of JSONLogic rather than flagd approach (JSON Logic has "cat" operator for strings but does not have anything like that for non-string types). We can fix that by implementing an additional custom operator.

That way we keep the new approach backwards compatible with old configs (and, tbh, there is no way around this - "cat" is a feature of JSONLogic, not flagd, so we cannot really block it unless we do some hacks).

docs/architecture-decisions/fractional-non-string-rand-units.md
```diff
@@ -0,0 +1,167 @@
...
+## Proposal
+
+We will modify the evaluation logic for the `fractional` operator.
+
+When inspecting the first element of the `fractional` array:
+
+1. If the first element in `fractional` evaluates to a non-array type then deterministically encode it to a well defined byte array and hash the bytes.
+2. Otherwise, if `targetingKey` is a string, build a 2-elements array of `flagKey` and `targetingKey`, deterministically encode that and hash (**NOTE:** This is different than string concatenation used today).
+3. Otherwise, if `targetingKey` is non-string, report an error and return a default value (as this breaks the [OpenFeature spec](https://openfeature.dev/specification/glossary/#targeting-key)).
+4. Otherwise, if `targetingKey` is missing, report an error and return a default value.
```
beeme1mr: > return a default value.

I believe the fractional operator should evaluate to a falsy value like `null` if targetingKey is missing.

toddbaert: This is a good point - currently - the rule returns null in this case - I don't think we should change it, though we could log a warning.

toddbaert:
> > I am in favor of this proposal, but I can't help but consider [this](https://github.com/open-feature/flagd/pull/1783#discussion_r2316738882) a bit of a blocker.
> > @beeme1mr WDYT about this one? If we cant append the flag key, should we remove this in the string case? I'd prefer not to have string as a "special case".
> 
> I would not necessarily consider this a special case for strings. It's a result of JSONLogic rather than flagd approach (JSON Logic has "cat" operator for strings but does not have anything like that for non-string types). We can fix that by implementing an additional custom operator.
> 
> That way we keep the new approach backwards compatible with old configs (and, tbh, there is no way around this - "cat" is a feature of JSONLogic, not flagd, so we cannot really block it unless we do some hacks).

I can accept that the "salting" only occurs in the string case, as long as we document it clearly. The main advantage of the "salting" is that it means that users don't end up in the same bucket for every fractional flag... but the nonstring situation is a corner case of a corner case here, so I don't want to block this ADR just for this.

docs/architecture-decisions/fractional-non-string-rand-units.md
cupofcat: Changed the ADR to specify returning nil.

Issue #1785, opened 2025-08-27
[FEATURE] Add OAuth2 client credentials flow authorization scheme


guidobrei:
# Goal
Support OAuth2 Client Credentials flow as authorization scheme for HTTP syncs.

# Context
Flagd currently supports configuring a (static) value passed in the `Authorization` header that is passed along sync requests to access the flag source. This mechanism does not work well when using short lived access tokens that need to be refreshed from time to time.

# Open questions
* Would other sync sources, like the blob storage sources also benefit from OAuth support in flagd?

toddbaert:
@thisthat

beeme1mr:
> * Would other sync sources, like the blob storage sources also benefit from OAuth support in flagd?

We use the [Go Cloud Kit](https://gocloud.dev/howto/blob/#s3) for interacting with blog storage. Auth is handled by the blob-specific driver.

PR #1791, opened 2025-09-02
feat: Add OAuth support for HTTP Sync

thisthat:
## This PR

- adds support for OAuth in the HTTP Sync

### Related Issues

Addresses #1785

### Notes

Properly testing this via unit tests is not possible. The HTTP Sync is built by the `syncbuilder` pkg. This entails that testing any behavior of the HTTP Sync can only be possible as part of the `syncbuilder` module, which IMHO doesn't make much sense.

### Follow-up Tasks

Add documentation for the new properties used in the HTTP Sync. 
Available here: https://github.com/open-feature/flagd/pull/1805

### How to test

Manually create a small mockup server that handles OAuth requests

```go

func (h *OAuthHandler) Authorize(w http.ResponseWriter, r *http.Request) {

	reqDump, _ := httputil.DumpRequest(r, true)
	fmt.Printf("GOT OAUTH REQUEST:\n%s", string(reqDump))

	w.WriteHeader(http.StatusOK)
	w.Header().Set("Content-Type", "application/x-www-form-urlencoded")
	w.Write([]byte("access_token=mocktoken&scope=mockscope&token_type=bearer"))
}
```

Afterwards, start flagd pointing to the mockup server:

```
go run main.go start --sources='[{ "uri": "http://localhost:8180/flags", "provider": "http", "interval": 1, "bearerToken": "it-will-be-overridden", "oauthConfig": { "clientID": "test", "clientSecret": "test", "tokenUrl": "http://localhost:8180/oauth" }}]'
```

The HTTP request to `http://localhost:8180/flags` will contain the HTTP Header `Authorization: Bearer mocktoken`.

### DIFF
--- a/core/pkg/sync/http/http_sync_test.go
+++ b/core/pkg/sync/http/http_sync_test.go
@@ -603,7 +603,10 @@ func TestHTTPSync_OAuth(t *testing.T) {
 					oauthMock.count++
 					w.WriteHeader(tt.oauthResponse)
 					w.Header().Set("Content-Type", "application/x-www-form-urlencoded")
-					w.Write([]byte(fmt.Sprintf("access_token=%s&scope=mockscope&token_type=bearer", bearerToken)))
+					_, err := w.Write([]byte(fmt.Sprintf("access_token=%s&scope=mockscope&token_type=bearer", bearerToken)))
+					if err != nil {
+						t.Fatalf("cannot write response: %v", err)
+					}
 					return
 				}
 				httpMock.lastHeader = r.Header
@@ -657,8 +660,11 @@ func TestHTTPSync_OAuthFolderSecrets(t *testing.T) {
 	)
 	// given an oauth server
 	oauthMock := &oauthHttpMock{}
+	flagsPath := "/flags.json"
+
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if strings.HasSuffix(r.URL.Path, oauthPath) {
+			// mock token response
 			body, _ := io.ReadAll(r.Body)
 			params, _ := url.ParseQuery(string(body))
 			oauthMock.lastParams = params
@@ -666,7 +672,21 @@ func TestHTTPSync_OAuthFolderSecrets(t *testing.T) {
 			oauthMock.count++
 			w.WriteHeader(http.StatusOK)
 			w.Header().Set("Content-Type", "application/x-www-form-urlencoded")
-			w.Write([]byte(fmt.Sprintf("access_token=%s&scope=mockscope&token_type=bearer", bearerToken)))
+			_, err := w.Write([]byte(fmt.Sprintf("access_token=%s&scope=mockscope&token_type=bearer", bearerToken)))
+			if err != nil {
+				t.Fatalf("cannot write response: %v", err)
+			}
+			return
+		} else if strings.HasSuffix(r.URL.Path, flagsPath) {
+			// mock flags response
+			io.ReadAll(r.Body)
+
+			w.WriteHeader(http.StatusOK)
+			w.Header().Set("Content-Type", "application/json")
+			_, err := w.Write([]byte(fmt.Sprintf(`{"flagKey": {"default": true}}`)))
+			if err != nil {
+				t.Fatalf("cannot write response: %v", err)
+			}
 			return
 		}
 		w.WriteHeader(http.StatusInternalServerError)
@@ -689,7 +709,7 @@ func TestHTTPSync_OAuthFolderSecrets(t *testing.T) {
 
 	l := logger.NewLogger(nil, false)
 	s := NewHTTP(sync.SourceConfig{
-		URI:         ts.URL,
+		URI:         ts.URL + flagsPath,
 		BearerToken: "it_should_be_replaced_by_oauth",
 		OAuth: &sync.OAuthCredentialHandler{
 			ClientID:     clientID,
@@ -699,9 +719,13 @@ func TestHTTPSync_OAuthFolderSecrets(t *testing.T) {
 			ReloadDelayS: 0, // we force loading the secret at each req
 		},
 	}, l)
-	d := make(chan sync.DataSync, 1)
+	d := make(chan sync.DataSync, 2)
 	// when we fire the HTTP call
-	s.ReSync(context.Background(), d)
+	err = s.ReSync(context.Background(), d)
+	if err != nil {
+		t.Fatalf("resync failed: %v", err)
+	}
+
 	// then the right secrets are used
 	require.Equal(t, secretClientID, oauthMock.lastParams.Get("client_id"))
 	require.Equal(t, secretClientSecret, oauthMock.lastParams.Get("client_secret"))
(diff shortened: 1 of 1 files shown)


core/pkg/sync/http/http_sync.go
```diff
@@ -192,3 +197,39 @@ func (hs *Sync) Fetch(ctx context.Context) (string, error) {
...
+		oauth := clientcredentials.Config{
+			ClientID:     config.OAuthConfig.ClientId,
+			ClientSecret: config.OAuthConfig.ClientSecret,
+			TokenURL:     config.OAuthConfig.TokenUrl,
+			AuthStyle:    oauth2.AuthStyleInParams,
+		}
+		client = oauth.Client(context.Background())
+	} else {
+		client = &http.Client{
+			Timeout: time.Second * 10,
```
thisthat: should we move this magic number into the config? 🤔 
@toddbaert

core/pkg/sync/builder/syncbuilder.go
```diff
@@ -186,29 +184,6 @@ func (sb *SyncBuilder) newK8s(uri string, logger *logger.Logger) (*kubernetes.Sy
 	), nil
 }
 
-func (sb *SyncBuilder) newHTTP(config sync.SourceConfig, logger *logger.Logger) *httpSync.Sync {
```
thisthat, line 189: move the logic of creating the HTTPSync inside the HTTPSync

core/pkg/sync/http/http_sync.go
```diff
@@ -9,25 +9,79 @@ import (
...
 type Sync struct {
-	URI         string
-	Client      Client
-	Cron        Cron
-	LastBodySHA string
-	Logger      *logger.Logger
-	BearerToken string
-	AuthHeader  string
-	Interval    uint32
+	uri         string
```
thisthat, line 27: we don't need to expose these details anymore since we moved its construction here

core/pkg/sync/http/http_sync.go
```diff
@@ -9,25 +9,79 @@ import (
...
+		oauth := &clientcredentials.Config{
+			ClientID:     och.clientId,
+			ClientSecret: och.clientSecret,
+			TokenURL:     och.tokenUrl,
+			AuthStyle:    oauth2.AuthStyleInParams,
+		}
+		return oauth, nil
+	}
+	// we load from files
+	id, err := os.ReadFile(och.folderSource + "/" + och.clientId)
```
thisthat: to avoid too many params, I am reusing the `clientId` as filename

aepfli: could we actually use just the following
```suggestion
	id, err := os.ReadFile(och.folderSource + "/clientId")
```
we can decide via config anyways where the secrets are stored, and this way we do not multipurpose the clientId field (sidenote same for ClientSecret).

// pair-review with @alexandraoberaigner @guidobrei

core/pkg/sync/http/http_sync.go
```diff
@@ -9,25 +9,79 @@ import (
...
+	clientSecret       string
+	tokenUrl           string
+	folderSource       string
+	reloadDelaySeconds int
+
+	lastUpdate time.Time
+}
+
+func (och *oauthCredentialHandler) loadOAuthConfiguration() (*clientcredentials.Config, error) {
+	if och.folderSource == "" {
```
aepfli, line 53: we should properly document, that if a foldersource is specified, we prioritize this, and ignore the definied clientId and ClientSecret - based on https://github.com/open-feature/flagd/pull/1791/files#r2332321164 - anyways we should properly document it, based on the outcome of the other comment

core/pkg/sync/http/http_sync.go
```diff
@@ -192,3 +246,61 @@ func (hs *Sync) Fetch(ctx context.Context) (string, error) {
 	body, _, err := hs.fetchBody(ctx, false)
 	return body, err
 }
+
+func (hs *Sync) getClient() Client {
+	if hs.oauthCredential != nil {
+		delta := time.Since(hs.oauthCredential.lastUpdate).Seconds()
+		if delta <= float64(hs.oauthCredential.reloadDelaySeconds) && hs.client != nil {
```
aepfli, line 254: [question] Isn't this suppose to be handled automagically by the oauthCredentials - at least that would be my expectations, sorry i did not investigate further, but i assume you have already more insights than me :) Or what is the purpose for this reloadDelaySeconds?

core/pkg/sync/isync.go
```diff
@@ -47,4 +47,15 @@ type SourceConfig struct {
 	Selector    string `json:"selector,omitempty"`
 	Interval    uint32 `json:"interval,omitempty"`
 	MaxMsgSize  int    `json:"maxMsgSize,omitempty"`
+
+	OAuthConfig *OAuthCredentialHandler `json:"oauthConfig,omitempty"`
```
aepfli:
```suggestion
	OAuthConfig *OAuthCredentialHandler `json:"oauth,omitempty"`
```

[nit] Naming it `oauthConfig` is somewhat redundant, as this is a configuration anyways, i thing we can simply call it `oauth`

// pair-review: @alexandraoberaigner @guidobrei

aepfli (commented):
Generally those changes look great, we do have some minor improvements/questions - but generally the implementation looks fine.

The biggest part missing tough is documentation, and currently this looks like a hidden feature. Especially the fact that filesource for oauth takes priority, and the current double purpose of clientId and clientSecret.

toddbaert:
I'll review this tomorrow, first thing. Thanks @thisthat

core/pkg/sync/http/http_sync.go
toddbaert: Yes I think I like @aepfli and @alexandraoberaigner 's suggestion better here - though I wonder if we should use `client-id` instead.

core/pkg/sync/isync.go
toddbaert: agree

core/pkg/sync/http/http_sync.go
toddbaert: Yes, I think so.

core/pkg/sync/builder/syncbuilder.go
```diff
@@ -193,11 +196,24 @@ func (sb *SyncBuilder) newHTTP(config sync.SourceConfig, logger *logger.Logger)
...
+			TokenURL:     config.OAuthConfig.TokenUrl,
+			AuthStyle:    oauth2.AuthStyleInParams,
+		}
+		client = oauth.Client(context.Background())
+	} else {
+		client = &http.Client{
 			Timeout: time.Second * 10,
-		},
+		}
+	}
```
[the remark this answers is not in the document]
toddbaert: @thisthat 

I think both are a good idea: https://github.com/golang/oauth2/blob/014cf778b444f29c82ececa4f3ec1f6fe3db3eaf/oauth2.go#L93-L95
WDYT?

core/pkg/sync/http/http_sync.go
thisthat: Ok, I will hardcode these under `client-id` and `client-secret`

core/pkg/sync/http/http_sync.go
thisthat: yes - doc will be done in a follow-up PR. I have currently issues with the build of the docs locally :/ something is not supported by ARM64

core/pkg/sync/http/http_sync.go
thisthat: The OAuth implementation handles automatically the refresh of tokens when they expire. This logic is to reload the client and secret after `reloadDelaySeconds`. I get custom to this behavior in K8s since secrets are usually mounted via ConfigMaps or secret injection on the file system. This reload of the auth params is handy-dandy to support rotation of secrets w/o restarting the container and dealing with mounting delays.

aepfli (approved):

toddbaert (approved):
I pushed one small change to fix some static checks on a unit test and improve some mocking

PR #1795, opened 2025-09-04
test: reduce complexity in tests

aepfli:


core/pkg/evaluator/fractional_test.go
```diff
@@ -427,24 +373,22 @@ func TestFractionalEvaluation(t *testing.T) {
...
-						Targeting: []byte(
-							`{
+			flags: map[string]model.Flag{
+				"headerColor": {
+					State:          "ENABLED",
+					DefaultVariant: "red",
+					Variants: map[string]any{
+						"red":  "#FF0000",
+						"blue": "#0000FF",
+					},
```
chrfwow:
```suggestion
					Variants: colorVariants,
```

core/pkg/evaluator/fractional_test.go
```diff
@@ -392,18 +345,12 @@ func TestFractionalEvaluation(t *testing.T) {
...
-							"green":  "#00FF00",
-							"yellow": "#FFFF00",
-						},
-						Targeting: []byte(`{
+			flags: map[string]model.Flag{
+				"headerColor": {
+					State:          "ENABLED",
+					DefaultVariant: "red",
+					Variants:       colorVariants,
+					Targeting: []byte(`{
```
chrfwow, line 353: Can we extract this whole flag declaration? It seems to be duplicated a few times

aepfli: I did not want to go so deep in comparing all the definitions. For now this should only reduce some of the things. Without digging deeper.

toddbaert (approved):

core/pkg/evaluator/fractional_test.go
toddbaert: @aepfli consider this.

Issue #1708, opened 2025-08-13
[ENHANCEMENT] Support ability to select on null flagSetId


toddbaert:
With our recent changes in https://github.com/open-feature/flagd/pull/1702, we can now use the `selector` to select flags based on their `flagSetId` - however, we can't explicitly select on flags WITHOUT a `flagSetId`. We need to:

- decide on a way to express this in the selector/header (ie: `flagSetId=` or `flagSetId=null` or `flagSetId=default`) or something like this. If we use `null` or `default` here, we must consider these specialk cases, so I recommend `flagSetId=` to express selecting the null set.
- implement selection on the null `flagSetId` (as currently implemented, this simply mean querying with the `nilFlagSetId` constant which is internally assigned to all flags without a `flagSetId`)
- optionally, find an alternative way (besides using the `nilFlagSetId` mentioned above) to index flags without `flagSetId`s.

> I think we discussed the problem when introducing the flag lists. Imagine there are two flags {key: feature, set: A} and {key: feature, set: null}, when an in-process provider constructs the selector, there's no way for it to get a valid flag map with {key: feature, set: null} in the result. It will be a problem for adopting flag sets - basically without the ability to select on the null set, users have to create a "default" set for all existing flags if they need to add a flag with the same key under another set.

_Originally posted by @tangenti in https://github.com/open-feature/flagd/pull/1702#discussion_r2267152692_

tangenti:
I'm leaning towards using `flagSetId=""` (easier for implementation) or simply `flagSetId=` (most consistent with the current syntax).

One question we may want to consider is the semantic of `flagSetID` explicitly set to `""` - do we consider it as the same as a flag with `flagSetID` unset?

guidobrei:
> One question we may want to consider is the semantic of `flagSetID` explicitly set to `""` - do we consider it as the same as a flag with `flagSetID` unset?

For simplicity I'd suggest to consider `nil` and `""` the same.

aepfli:
should we maybe stick to [CEL](http://cel.dev) and use a limited set of the common expression language in general?

beeme1mr:
> should we maybe stick to [CEL](http://cel.dev) and use a limited set of the common expression language in general?

I'd prefer to keep it simpler than that unless there are requirements that justify the additional complexity.

toddbaert:
[...]

Issue #1673, opened 2025-07-10
[FEATURE] Parse  Array of `flags` Objects From All Sync Sources


toddbaert:
# [FEATURE] Parse Array of flags Objects From All Sync Sources

## Background

flagd needs to support flags defined in an array, as well as the previous map format.

## Requirements

- [x] Flags must be parsed, and stored in memory, annotated with both their `flagSetId` (`nil` is a possible flagSetId) as well as their `source`
- [x] Refactors to the underlying storage layers are encouraged (this was mostly done with (https://github.com/open-feature/flagd/pull/1697).
- [ ] Particular care should be taken to make retreival of sets of flags based on their flag set efficient and easy
- [ ] Associated tests

[See parent issue for more background](https://github.com/open-feature/flagd/issues/1671).

toddbaert:
I'm assigning this to myself; probably won't complete it but at least doing a PoC.

PR #1797, opened 2025-09-08
feat!: enable parsing of array flag configurations for flagd

aepfli:
This pull request introduces a significant internal refactoring of the flag storage mechanism, transitioning from a map-based structure to a slice-based one. This change aims to enhance the flexibility and robustness of flag management, particularly in scenarios involving multiple flag sources and flag sets. The update also simplifies the evaluator's state management interface, contributing to a cleaner and more focused API for flag updates.

### Highlights

* **Store Interface Refactoring**: The core store interface (IStore) and its implementations have been refactored to manage flags as a slice of `model.Flag` instead of a map. This change impacts how flags are stored, retrieved, and updated internally, allowing for more flexible handling of flags, especially those associated with different flag sets.
* **IEvaluator Interface Simplification**: The `IEvaluator` interface has been simplified by removing the `GetState()` method and modifying the `SetState()` method signature. The `SetState()` method no longer returns detailed event information or a resync boolean, streamlining its contract to simply indicate success or failure via an error.
* **Flag Key Handling in Store**: To accommodate flags from different sources and flag sets, the internal storage now uses a compound key (`FlagSetId|Key`) to uniquely identify flags. This ensures correct handling of flags that might have the same `Key` but originate from different flag sets or sources.
* **Test Suite Updates**: Extensive updates have been made across the test suite, including `fractional_test.go`, `json_test.go`, `semver_test.go`, `string_comparison_test.go`, and `store_test.go`, to align with the new `[]model.Flag` structure and the simplified `IEvaluator.SetState()` signature. This includes changes to test data structures and assertion logic.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,4 +7,3 @@
 [submodule "schemas"]
 	path = schemas
 	url = https://github.com/open-feature/flagd-schemas.git
-	branch = jsonschema-v0.2.13
--- a/core/pkg/store/store_test.go
+++ b/core/pkg/store/store_test.go
@@ -458,7 +458,8 @@ func TestGetAllNoWatcher(t *testing.T) {
 
 	for _, tt := range tests {
 		for _, s := range sourceOrder {
-			tt := tt
+			wantFlags := make([]model.Flag, len(tt.wantFlags))
+			copy(wantFlags, tt.wantFlags)
 			t.Run(tt.name+" - "+s.name, func(t *testing.T) {
 				t.Parallel()
 
@@ -472,12 +473,12 @@ func TestGetAllNoWatcher(t *testing.T) {
 				}
 				gotFlags, _, _ := store.GetAll(context.Background(), tt.selector)
 
-				require.Equal(t, len(tt.wantFlags), len(gotFlags))
-				sort.Slice(tt.wantFlags, func(i, j int) bool {
-					if tt.wantFlags[i].FlagSetId != tt.wantFlags[j].FlagSetId {
-						return tt.wantFlags[i].FlagSetId < tt.wantFlags[j].FlagSetId
+				require.Equal(t, len(wantFlags), len(gotFlags))
+				sort.Slice(wantFlags, func(i, j int) bool {
+					if wantFlags[i].FlagSetId != wantFlags[j].FlagSetId {
+						return wantFlags[i].FlagSetId < wantFlags[j].FlagSetId
 					}
-					return tt.wantFlags[i].Key < tt.wantFlags[j].Key
+					return wantFlags[i].Key < wantFlags[j].Key
 				})
 				sort.Slice(gotFlags, func(i, j int) bool {
 					if gotFlags[i].FlagSetId != gotFlags[j].FlagSetId {
@@ -485,7 +486,7 @@ func TestGetAllNoWatcher(t *testing.T) {
 					}
 					return gotFlags[i].Key < gotFlags[j].Key
 				})
-				require.Equal(t, tt.wantFlags, gotFlags)
+				require.Equal(t, wantFlags, gotFlags)
 			})
 		}
 	}
--- a/flagd/pkg/service/flag-sync/handler.go
+++ b/flagd/pkg/service/flag-sync/handler.go
@@ -5,9 +5,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/open-feature/flagd/core/pkg/model"
 	"maps"
 	"time"
+
+	"github.com/open-feature/flagd/core/pkg/model"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -82,15 +83,6 @@ func (s syncHandler) SyncFlags(req *syncv1.SyncFlagsRequest, server syncv1grpc.F
 			return nil
 		}
 	}
-}
-
-type exportFlag struct {
-	State          string          `json:"state"`
-	DefaultVariant string          `json:"defaultVariant"`
-	Variants       map[string]any  `json:"variants"`
-	Targeting      json.RawMessage `json:"targeting,omitempty"`
-	Source         string          `json:"source"`
-	Metadata       model.Metadata  `json:"metadata,omitempty"`
 }
 
 func (s syncHandler) convertMap(flags []model.Flag) map[string]model.Flag {


flagd/pkg/service/flag-evaluation/eventing.go
```diff
@@ -39,7 +39,10 @@ func (eventing *eventingConfiguration) Subscribe(ctx context.Context, id any, se
 		// store the previous flags to compare against new notifications, to compute proper diffs for RPC mode
 		var oldFlags map[string]model.Flag
 		for result := range watcher {
-			newFlags := result.Flags
+			newFlags := make(map[string]model.Flag)
+			for _, flag := range result.Flags {
+				newFlags[flag.Key] = flag
+			}
```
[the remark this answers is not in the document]
aepfli: we are not expecting multiple flags from a different set here - @toddbaert can you confirm? else i would need to refactor the whole notification approach

core/pkg/store/store.go
```diff
@@ -275,9 +290,9 @@ func (s *Store) Update(
 			if oldFlag.FlagSetId != newFlag.FlagSetId {
```
aepfli, line 290: this needs to be delete, this was forgot to add this

flagd/pkg/service/flag-evaluation/eventing.go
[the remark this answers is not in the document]
toddbaert: Yes this is correct - we should already be filtering by set here (or if not, we should be using priorities, in which case we know how the overrides will work).

flagd/pkg/service/flag-evaluation/eventing.go
```diff
@@ -39,7 +39,10 @@ func (eventing *eventingConfiguration) Subscribe(ctx context.Context, id any, se
 		// store the previous flags to compare against new notifications, to compute proper diffs for RPC mode
 		var oldFlags map[string]model.Flag
 		for result := range watcher {
-			newFlags := result.Flags
+			newFlags := make(map[string]model.Flag)
+			for _, flag := range result.Flags {
+				newFlags[flag.Key] = flag
```
toddbaert, line 45:
```suggestion
                // we should be either selecting on a flag set here, or using the source-priority - duplicates are already handled, so we don't have to worry about overwrites
				newFlags[flag.Key] = flag
```

flagd/pkg/service/flag-evaluation/eventing.go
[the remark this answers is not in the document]
toddbaert: I've added a comment about this in-line.

toddbaert (approved):
LGTM:

- tests make sense
- indexes make sense
- I did some manual testing and selection/notification seem to work as expected

core/pkg/evaluator/json.go
```diff
@@ -454,30 +457,62 @@ func (je *JSON) configToFlagDefinition(config string, definition *Definition) er
...
+	trimmed := strings.TrimSpace(string(rawDef.Flags))
+	if strings.HasPrefix(trimmed, "[") {
+		// It's an array
+		var flags []model.Flag
+		if err := json.Unmarshal(rawDef.Flags, &flags); err != nil {
+			return fmt.Errorf("unmarshalling flags array: %w", err)
+		}
+		definition.Flags = flags
+	} else {
+		// It's an object
```
chrfwow, line 489: Should we also check if it is a (illegal) primitive?

core/pkg/store/store.go
```diff
@@ -75,6 +77,17 @@ func NewStore(logger *logger.Logger, sources []string) (*Store, error) {
 						Unique:  false,
 						Indexer: &memdb.StringFieldIndex{Field: model.FlagSetId, Lowercase: false},
 					},
+					// for looking up by flagSetId and key
+					flagSetIdKeyIndex: {
+						Name:   flagSetIdKeyIndex,
+						Unique: false,
```
chrfwow: I'm not familiar with `go-memdb`, but what does the `Unique` mean and why do we set it to `false`?

aepfli: we do have now multiple possible entries in our database where the flagsetId and the key are the same, because we can get those from multiple sources - setting this to true, would limit the possibilities, and might produce wrong information

core/pkg/evaluator/json.go
aepfli: we will error on the unmarshalling anyways - i do not think it is worth the extra check

core/pkg/store/store.go
```diff
@@ -54,6 +55,7 @@ func NewStore(logger *logger.Logger, sources []string) (*Store, error) {
 							Indexes: []memdb.Indexer{
 								&memdb.StringFieldIndex{Field: model.FlagSetId, Lowercase: false},
 								&memdb.StringFieldIndex{Field: model.Key, Lowercase: false},
+								&memdb.StringFieldIndex{Field: model.Source, Lowercase: false},
```
tangenti: Why we need to index the source?

core/pkg/store/store.go
tangenti: I'm confused by this change. 

I remember that we were talking about using (flagSetID, flagKey) as the composite key, and IIRC @toddbaert  confirmed that if there are same composite keys from different sources, the latest one will override as the semantic is always a full set, not update.

tangenti (changes requested):
Could you also update benchmark.txt to reflect the impact of this change?

core/pkg/store/store.go
aepfli: so here comes my assumption, lets say we do have two sources, with the same flagset "awesomeFlagSet" and the same flag, call them sourceA and sourceB. if i select based on flagset i want to get the flag from sourceB. but if i select based on source and request sourceA, i want to get the flag from sourceA - but without adding this, to the primary key, we ensure that both versions stay in the database

toddbaert: I think this is a valid-enough use case that we can justify this.

We can of course NOT implement this if we want to be strict about the composite key being SOLELY the flagSet/Key.

I think before release, we should document this more thoroughly though, along with deciding on [this](https://github.com/open-feature/flagd/issues/1708) and documenting that as well. We need documentation around how our new selector works in general. I'm happy to write these docs once we are all in agreement and have a release candidate.

toddbaert:
@aepfli @tangenti 

I re-ran the benchmark: https://github.com/open-feature/flagd/pull/1797/commits/3e99ad4c6e16253e37f7c8f0c8a4612eba829dfd (no significant deviation, looks like 15% faster/slower across various tests).

I hate that we don't have automation for this; I may have mentioned before that we once did, but github actions are far too vulnerable to noisy-neighbor effects to be useful in this sort of benchmarking.

core/pkg/store/store.go
tangenti: It sounds a new feature instead of a refactor? Before having the selectors for flag set IDs and memdb change, flags with the same key from different source are overriding based on the priority IIUC.

core/pkg/store/store.go
```diff
@@ -15,14 +16,14 @@ var noValidatedSources = []string{}
 type SelectorContextKey struct{}
 
 type FlagQueryResult struct {
-	Flags map[string]model.Flag
+	Flags []model.Flag
```
aepfli, line 19: this change breaks the sync endpoint - we need to revert this

core/pkg/store/store.go
aepfli: i am not sure if this is a new feature or a undefined side-effect of the maps configuration. the maps configuration and the selector allow for different possible configuration and retrieval of flags. We never specified what happens if a flag is defined in multiple sources for the same flagset, and what is the impact for the user. -> my interpretation, we are using a database and a selector, i allow selection per flagset and per source, if i select by source, it is unexpected for me, that the flag is missing, although it is defined in the source. Maybe we should clarify what we expect in this case. Do we need to improve the ADR or create a new one? Is it worth it?

toddbaert: I can go either way on this one, TBH.

I agree with @tangenti that this IS really a new feature. The only thing to consider is which behavior is more in accordance to the _principle of least astonishment_. We could remove this new index and tests and fundamentally the PR would have accomplished it's goal.

toddbaert: @beeme1mr do you have any opinion on this one?

core/pkg/store/store.go
toddbaert: This discussion continues [here](https://github.com/open-feature/flagd/pull/1797#discussion_r2336972714).

core/pkg/store/store.go
aepfli: The more I am thinking about it, i agree that the simplicity of flagset + key is desireable. but I think that we should keep this simplicity to the outside, and decouple this from our store. I already mentioned the selector edge case, but there is eg. also this one, say we have:

Source A (higher priority) has flag "new-checkout"
Source B (lower priority) also has flag "new-checkout" but with completely different targeting
Source A removes the flag
Source B wasnt updated in hours

do we want to fallback in this case to B? should be with the next update of source B suddenly reappear? 

Another case, how do we handle updates when the source is asynchronous, eg. the grpc streams, and those are not happening in order, how do we define if we should overwrite the flag. The priority/source is hard to utilize, because again, maybe the priortized source removes this flag.

I do agree that this adds complexity on the implementation level, and makes the logic less clear, but i think it is easier to keep the data in memdb and handle it via logic, rather than deciding already on import how to handle this. Currently this datamodel represents the source model, it is the same as the sources, and as a user I would expect this to happen, when querying for data. As it is in sync with the data i provide. The simplicity for flagset and key is still valid to the outside, as providers only select currently based on selector (old deprecated) or flagsetid (new stylish fancy awesome) ;).

We also need to be aware that although we do not encourage people to such a setup, we should still somehow be aware of potential occurences, especially as this duplication might also be a migration scenario from one source to another.

I am not stricly opposing with just flagsetid and key as a primary key, but i think this might add more complexity or unpredicted behaviour in the future. - so how should we proceed? (if we continue with just flagsetId and key, we should definitely define our expected behaviour clearly)

toddbaert: I think these are good questions, but maybe they are questions for a different PR so that we can move forward with the array support for now.

toddbaert: @aepfli can we revert just the part? I think we can just remove a couple tests and the new index. 

I think for now it's just easier to say the PK is flag set + key, and keep it there, with any behavior around duplicates unspecified and not a supported use case

aepfli: yes i will, sorry i was occupied with other topics and sick

i am still curious to here the opinion about all the strange cases outlined from me by @tangenti - currently it feels like we are allowing a lot of options, but only focus on some specific usecases/implementations and ignoring all other possibilities which can happen.

aepfli: i improved the tests for store_test.go, by adding another dimension in which i am also changing the order updates are happening (especially interesting for asynchronous data sources like grpc) (maybe my knowledge here is wrong, and this is not a problem) - but as soon as i am removing the source from the primary key, i am getting a lot inconsistencies. 

Therefor, i just want to check with everyone, should i really revert to just flagsetid and key? - remember this  is only our internal data representation.

// edit: food for thought - part of this is related with our multisources approach, removing this would be the other solution to this problem ;)

tangenti: Sorry, I was sick last week. Feel free to ping me on Slack if I won't respond in time.

Regarding the cases you mentioned, yes it's a problem of today and there are more cases we should handle more gracefully, such as the granularity of the override, whether a source could revert a version etc. It's an important improvement that I'd like to have, however I don't think this PR is a good place to solve that problem. 

I'd appreciate if you or someone could start an ADR to discuss the problems we aim to solve more thoroughly, and check if there's API changes required for the changes. Once people agree on the scope and the solution, we could then go back to the implementation.

aepfli: fyi: @open-feature/flagd-approvers - any other input?

toddbaert: I think this is not something we should add now, and at the moment I'm not convinced it should be added at all.

My reasoning:

- though this helps in the hypothetical situation that a `flagdSetId`+`key` are duplicated in multiple sources, it doesn't help with the case that it's duplicated in the same source (so we aren't really fully removing the edge case)
- it's much easier to explain to someone "the PK for flags in flagd is a compound of `flagSetId` and `flagKey`" than to also explain `sources`; without sources it's a simpler model
- the problem this is solving is not a use-case I think will really be encountered in often in real usage; simply don't create source with flags with duplicate PKs (I want to document this once we're done all the associated work).
- there's no ADR discussing this, it's a scope creep on top of the [original goal of supporting arrays of flags](https://github.com/open-feature/flagd/issues/1673)

If somebody requests this feature, I'm happy to consider adding it, but I truly don't think it's worth the complexity before that point, and, so far, this functionality hasn't been agreed on.

beeme1mr: I agree that this doesn't seem like an edge case we need to support now. When we document this feature, we can explicitly mention that duplicate flag key + flag set ID combos are not supported. That was the uniqueness constraint we agreed on in the ADR.

aepfli: i still think that as a library/tool implementor, it is my responsibility to provide a solution for all the possible usecases which the configuration allows, although it is not within my usecase.  - i reverted the change, i commented out the tests, in which i am loading the sources in different kind of orders (might still be the case that his is happening due to latency of different sources). so they could be easier activated if needed

toddbaert (approved):
@aepfli Thanks!

I made a few small changes in this commit: https://github.com/open-feature/flagd/pull/1797/commits/a47ddd6158dacd958cca3e3128852728b1f8f0f4

- removed branch from submodule
- fixed a race condition in a test (only the test itself had the race, not the prod code)
- removed an unused type

toddbaert:
> Note this is only a change in the core lib, not flagd itself, so I will modify the flagd release accordingly when we release this.

Tested this manually as well, merging now.

PR #1798, opened 2025-09-09
feat: add configuration property to define which key is used if a string without equals is used as a selector

aepfli:


core/pkg/store/store.go
```diff
@@ -28,16 +28,22 @@ type IStore interface {
 var _ IStore = (*Store)(nil)
 
 type Store struct {
-	db      *memdb.MemDB
-	logger  *logger.Logger
-	sources []string
+	db                    *memdb.MemDB
+	logger                *logger.Logger
+	sources               []string
+	expressionFallbackKey string
```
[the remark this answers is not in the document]
aepfli: this is not an issue, we are actually updating this where we need it.

aepfli:
i need to rework this... to much mixed approaches

aepfli:
/gemini review

last review from gemini, rest is for the community

core/pkg/store/query.go
```diff
@@ -29,50 +30,99 @@ var nilFlagSetId = uuid.New().String()
...
-func expressionToMap(sExp string) map[string]string {
+func expressionToMap(sExp string, fallbackExpressionKey string) (map[string]string, bool) {
 	selectorMap := make(map[string]string)
 	if sExp == "" {
-		return selectorMap
+		return selectorMap, false
 	}
 
+	// Handle pure fallback case (no equals anywhere)
 	if strings.Index(sExp, "=") == -1 {
```
toddbaert, line 55: @aepfli staticcheck says: S1003: should use !strings.Contains(sExp, "=") instead (staticcheck), which makes sense to me.

core/pkg/store/query.go
```diff
@@ -29,50 +30,99 @@ var nilFlagSetId = uuid.New().String()
 
 // A selector represents a set of constraints used to query the store.
 type Selector struct {
-	indexMap map[string]string
+	indexMap      map[string]string
+	usingFallback bool
```
toddbaert, line 34: I don't think we need this member (per Selector), and we don't need a separate "`With`" for it either, since this is an immutable global setting (all Selectors will share it).

Instead, I would recommend taking advantage of module initialization and `sync.Once` to do this, by adding _something_ like this at the top of the file (just below the imports):

```go
var fallbackKey string // our fallback key global
var fallbackKeyOnce sync.Once // sync.Once instance

// public function
func SetFallbackKey(key string) {
	fallbackKeyOnce.Do(func() {
		fallbackKey = key
	})
}
```

Then we just call `SetFallbackKey `at some point during init, which sets the fallback key ONLY ONCE for the duration of the program; then we can get rid of this per instance and just always refer to this module global.

core/pkg/store/query_test.go
```diff
@@ -156,38 +320,342 @@ func TestSelector_ToMetadata(t *testing.T) {
...
+		indexMap: map[string]string{
+			"source":    "benchmark-source",
+			"flagSetId": "benchmark-flagset",
+		},
+	}
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_ = selector.WithIndex("key", "benchmark-key")
+	}
```
toddbaert, line 660: I don't think we need specific benchmark tests for this.

core/pkg/store/query_test.go
```diff
@@ -156,38 +320,342 @@ func TestSelector_ToMetadata(t *testing.T) {
...
+				t.Errorf("expressionToMap() map = %v, want %v", gotMap, tt.wantMap)
+			}
+			if gotUsingFallback != tt.wantUsingFallback {
+				t.Errorf("expressionToMap() usingFallback = %v, want %v", gotUsingFallback, tt.wantUsingFallback)
+			}
+		})
+	}
+}
+
+func TestSelector_WithFallback(t *testing.T) {
```
toddbaert, line 502: We can remove all these tests (perhaps moving 1 or 2 to the other functions) if we do [this](https://github.com/open-feature/flagd/pull/1798/files#r2408215431).

core/pkg/store/query.go
toddbaert: BTW for testing, you can manually reset the module private `sync.Once`.

toddbaert:
This was mostly to facilitate an internal migration issue, but we've come up with an alternative that will avoid adding this to flagd.

aepfli:
I am not sure I agree here, yes the motivation stems from an internal problem, and we might be able to fix this differently internally, but the problem still exists for other users of flagd, which also do need some kind of migration path.

How should they handle this?

toddbaert:
> I am not sure I agree here, yes the motivation stems from an internal problem, and we might be able to fix this differently internally, but the problem still exists for other users of flagd, which also do need some kind of migration path.
> 
> How should they handle this?

This is not a problem for external users. A selector without an expression will continue to represent a source, there's no plan to change this, and no need to change it, since up to now is been the only possibility.

aepfli:
i am not sure i am agreeing with this comment, because flagset will be in the future the default approach, and as a library/service we should provide a migration path also for other projects. But it is not he right approach to this via this pr implicitly, we should do this explicitly with also defining the migration and deprecation path. In the end this is also something like the experimental flags for the jvm, till a feature reaches GA and afterwards they get removed again. We should make this migration easy for users of flagd to also rely on the flagset as a default, without forcing their sdk users, to update their dependencies.

toddbaert:
> i am not sure i am agreeing with this comment, because flagset will be in the future the default approach, and as a library/service we should provide a migration path also for other projects. But it is not he right approach to this via this pr implicitly, we should do this explicitly with also defining the migration and deprecation path. In the end this is also something like the experimental flags for the jvm, till a feature reaches GA and afterwards they get removed again. We should make this migration easy for users of flagd to also rely on the flagset as a default, without forcing their sdk users, to update their dependencies.

Nobody who is using flagd right now is using the selector for a flagSetId because it's not possible, so there's no need for a "migration path". Anyone who wants to select on a flagSetId can use an expression to do that in the future, so I'm not seeing the point.

PR #1800, opened 2025-09-10
docs(ADR): extends the fractional operator to support up to .001% distributions

beeme1mr:
## This PR

- extends the fractional operator to support up to .001% distributions.
- defines the expected behavior of several edge cases.

### Notes

Addresses a limitation of the current fractional operator that prevents sub-percent traffic allocations. In high-throughput services, 1% of traffic may represent a significant number of requests.

### DIFF
--- a/docs/architecture-decisions/high-precision-fractional-bucketing.md
+++ b/docs/architecture-decisions/high-precision-fractional-bucketing.md
@@ -1,0 +1,276 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: Michael Beemer
+created: 2025-09-10
+updated: 2025-09-10
+---
+
+# High-Precision Fractional Bucketing for Sub-Percent Traffic Allocation
+
+This ADR proposes enhancing the fractional operation to support high-precision traffic allocation down to 0.001% granularity by increasing the internal bucket count from 100 to 100,000 while maintaining the existing weight-based API.
+
+## Background
+
+The current fractional operation in flagd uses a 100-bucket system that maps hash values to percentages in the range [0, 100].
+This approach works well for most use cases but has significant limitations in high-throughput environments where precise sub-percent traffic allocation is required.
+
+Currently, the smallest allocation possible is 1%, which is insufficient for:
+
+- Gradual rollouts in ultra-high-traffic systems where 1% could represent millions of users
+- A/B testing scenarios requiring precise control over small experimental groups
+- Canary deployments where operators need to start with very small traffic percentages (e.g., 0.1% or 0.01%)
+
+The current implementation in `fractional.go` calculates bucket assignment using:
+
+```go
+bucket := hashRatio * 100 // in range [0, 100]
+```
+
+This limits granularity to 1% increments, making it impossible to achieve the precision required for sophisticated traffic management strategies.
+
+## Requirements
+
+- Support traffic allocation precision down to 0.001% (3 decimal places)
+- Maintain backwards compatibility with existing weight-based API
+- Preserve deterministic bucketing behavior (same hash input always produces same bucket)
+- Ensure consistent bucket assignment across different programming languages
+- Support weight values up to a reasonable maximum that works across multiple languages
+- Maintain current performance characteristics
+- Prevent users from being moved between buckets when only distribution percentages change
+- Guarantee that any variant with weight > 0 receives some traffic allocation
+- Handle edge cases gracefully without silent failures
+- Validate weight configurations and provide clear error messages for invalid inputs
+
+## Considered Options
+
+- **Option 1: 10,000 buckets (0.01% precision)** - 1 in every 10,000 users, better but still not sufficient for many high-throughput use cases
+- **Option 2: 100,000 buckets (0.001% precision)** - 1 in every 100,000 users, meets most high-precision needs
+- **Option 3: 1,000,000 buckets (0.0001% precision)** - 1 in every 1,000,000 users, likely overkill and could impact performance
+
+## Proposal
+
+Implement a 100,000-bucket system that provides 0.001% precision while maintaining the existing integer weight-based API.
+
+### API changes
+
+No API changes are required. The existing fractional operation syntax remains unchanged:
+
+```json
+"fractional": [
+  { "cat": [{ "var": "$flagd.flagKey" }, { "var": "email" }] },
+  ["red", 50],
+  ["blue", 30], 
+  ["green", 20]
+]
+```
+
+### Implementation Changes
+
+1. **Bucket Count**: Change from 100 to 100,000 buckets by modifying bucket calculation from `hashRatio * 100` to `hashRatio * 100000`
+2. **Minimum Allocation Guarantee**: Any variant with weight > 0 receives at least 1 bucket (0.001%)
+3. **Excess Bucket Handling**: Remove excess buckets from the largest variant to maintain exactly 100,000 total buckets
+4. **Weight Sum Validation**: Reject configurations where total weight exceeds maximum safe integer value
+5. **Maximum Weight Sum**: Use language-specific maximum 32-bit signed integer constants for cross-platform compatibility
+
+### Minimum Allocation Guarantee
+
+To prevent silent configuration failures, any variant with a positive weight will receive at least 0.001% allocation (1 bucket), even if the calculated percentage would round to zero. This ensures predictable behavior where positive weights always result in some traffic allocation.
+
+**Example**: Configuration `["variant-a", 1], ["variant-b", 1000000]`
+
+- Without guarantee: variant-a gets 0% (never selected)
+- With guarantee: variant-a gets 0.001%, variant-b gets 99.999%
+
+### Excess Bucket Management
+
+When minimum allocations cause the total to exceed 100,000 buckets, excess buckets are removed from the variant with the largest allocation.
+This approach:
+
+- Maintains the minimum guarantee for small variants
+- Has minimal impact on large variants (small relative reduction)
+- Preserves deterministic behavior
+- Prevents bucket count overflow
+
+### Weight Sum Validation
+
+When the total weight sum exceeds the maximum safe integer value, the fractional evaluation will return a validation error with a clear message.
+This prevents integer overflow issues and provides immediate feedback to users about invalid configurations.
+
+```go
+import "math"
+
+func validateWeightSum(variants []fractionalEvaluationVariant) error {
+    var totalWeight int64 = 0
+    for _, variant := range variants {
+        totalWeight += int64(variant.weight)
+        if totalWeight > math.MaxInt32 {
+            return fmt.Errorf("total weight sum %d exceeds maximum allowed value %d", 
+                totalWeight, math.MaxInt32)
+        }
+    }
+    return nil
+}
+```
+
+Implementations should prefer built-in language constants (e.g., `math.MaxInt32` in Go, `Integer.MAX_VALUE` in Java, `int.MaxValue` in C#) rather than hardcoded values to ensure maintainability and clarity.
+
+### Edge Case Handling
+
+The implementation addresses several edge cases:
+
+1. **All weights are 0**: Returns empty string (maintains current behavior)
+2. **Negative weights**: Treated as 0 (maintains current validation behavior)
+3. **Single variant**: Receives all 100,000 buckets regardless of weight value
+4. **Empty variants**: Returns error (maintains current validation behavior)
+5. **Weight sum overflow**: Returns validation error with clear message
+6. **Multiple variants with minimum allocation**: Excess distributed fairly among largest variants
+
+### Maximum Weight Considerations
+
+To ensure cross-language compatibility, we establish a maximum total weight sum equal to the maximum 32-bit signed integer value (2,147,483,647). This limit:
+
+- Works reliably across all target languages (Go, Java, .NET, JavaScript, Python)
+- Provides more than sufficient range for any practical use case
+- Prevents integer overflow issues in 32-bit signed integer systems
+- Allows for extremely fine-grained control (individual weights can be 1 out of 2+ billion)
+- Uses language-native constants for better maintainability
+
+### Code Changes
+
+The following shows how the core logic in `fractional.go` would be modified.
+
+```go
+const bucketCount = 100000
+
+// bucketAllocation represents the number of buckets allocated to a variant
+type bucketAllocation struct {
+    variant string
+    buckets int
+}
+
+func (fe *Fractional) Evaluate(values, data any) any {
+    valueToDistribute, feDistributions, err := parseFractionalEvaluationData(values, data)
+    if err != nil {
+        fe.Logger.Warn(fmt.Sprintf("parse fractional evaluation data: %v", err))
+        return nil
+    }
+
+    if err := validateWeightSum(feDistributions.weightedVariants); err != nil {
+        fe.Logger.Warn(fmt.Sprintf("weight validation failed: %v", err))
+        return nil
+    }
+
+    return distributeValue(valueToDistribute, feDistributions)
+}
+
+func validateWeightSum(variants []fractionalEvaluationVariant) error {
+    var totalWeight int64 = 0
+    for _, variant := range variants {
+        totalWeight += int64(variant.weight)
+        if totalWeight > math.MaxInt32 {
+            return fmt.Errorf("total weight sum %d exceeds maximum allowed value %d", 
+                totalWeight, math.MaxInt32)
+        }
+    }
+    return nil
+}
+
+func calculateBucketAllocations(variants []fractionalEvaluationVariant, totalWeight int) []bucketAllocation {
+    allocations := make([]bucketAllocation, len(variants))
+    totalAllocated := 0
+    
+    // Calculate initial allocations
+    for i, variant := range variants {
+        if variant.weight == 0 {
+            allocations[i] = bucketAllocation{variant: variant.variant, buckets: 0}
+        } else {
+            // Calculate proportional allocation
+            proportional := int(float64(variant.weight) / float64(totalWeight) * bucketCount)
+            // Ensure minimum allocation of 1 bucket for any positive weight
+            buckets := max(1, proportional)
+            allocations[i] = bucketAllocation{variant: variant.variant, buckets: buckets}
+        }
+        totalAllocated += allocations[i].buckets
+    }
+    
+    // Handle excess buckets by removing from largest allocation
+    excess := totalAllocated - bucketCount
+    if excess > 0 {
+        // Sort indices by bucket count (descending) to find largest allocation
+        indices := make([]int, len(allocations))
+        for i := range indices {
+            indices[i] = i
+        }
+        sort.Slice(indices, func(i, j int) bool {
+            return allocations[indices[i]].buckets > allocations[indices[j]].buckets
+        })
+        
+        // Remove excess from largest allocation, respecting minimum guarantee
+        for _, idx := range indices {
+            if excess <= 0 {
+                break
+            }
+            
+            // Don't reduce below 1 bucket if original weight > 0
+            minAllowed := 0
+            if variants[idx].weight > 0 {
+                minAllowed = 1
+            }
+            
+            canRemove := allocations[idx].buckets - minAllowed
+            toRemove := min(excess, canRemove)
+            allocations[idx].buckets -= toRemove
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,276 @@
...
+## Requirements
+
+- Support traffic allocation precision down to 0.001% (3 decimal places)
+- Maintain backwards compatibility with existing weight-based API
+- Preserve deterministic bucketing behavior (same hash input always produces same bucket)
+- Ensure consistent bucket assignment across different programming languages
+- Support weight values up to a reasonable maximum that works across multiple languages
+- Maintain current performance characteristics
+- Prevent users from being moved between buckets when only distribution percentages change
+- Guarantee that any variant with weight > 0 receives some traffic allocation
```
beeme1mr, line 41: I went back and forth on this. It isn't necessary if the flag is configured properly but I'm afraid that it wouldn't be that obvious that there's a misconfiguration. This basically prevents 0% distribution if a weight is defined.

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,276 @@
...
+        totalWeight += int64(variant.weight)
+        if totalWeight > math.MaxInt32 {
+            return fmt.Errorf("total weight sum %d exceeds maximum allowed value %d", 
+                totalWeight, math.MaxInt32)
+        }
+    }
+    return nil
+}
+
+func calculateBucketAllocations(variants []fractionalEvaluationVariant, totalWeight int) []bucketAllocation {
```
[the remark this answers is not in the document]
beeme1mr: Properly supporting guaranteed bucketing may add more complexity than I'd like. I'm sure I can address this issue but I'd like feedback on if it's worth supporting this at all.

The reason I added this is to avoid configurations like `["variant-a", 1], ["variant-b", 1000000]`, resulting in zero evaluations returning `variant-a`. You could argue it's a misconfiguration, but it may not be obvious the the user.

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,279 @@
...
+- Prevent users from being moved between buckets when only distribution percentages change
+- Guarantee that any variant with weight > 0 receives some traffic allocation
+- Handle edge cases gracefully without silent failures
+- Validate weight configurations and provide clear error messages for invalid inputs
+
+## Considered Options
+
+- **Option 1: 10,000 buckets (0.01% precision)** - 1 in every 10,000 users, better but still not sufficient for many high-throughput use cases
+- **Option 2: 100,000 buckets (0.001% precision)** - 1 in every 100,000 users, meets most high-precision needs
+- **Option 3: 1,000,000 buckets (0.0001% precision)** - 1 in every 1,000,000 users, likely overkill and could impact performance
```
toddbaert, line 49: Is there any obvious reason that we don't want the max bucket amount to be the sum of all bucket ratios?

ie: `["variant-a", 1], ["variant-b", 1000000]` results in `1000001` buckets?

This is:

- backwards compatible
- supports infinite precision (though we'd still have to do the max integer check you mentioned)
- allows us to sidestep the "Minimum Allocation Guarantee" (what you are describing [here](https://github.com/open-feature/flagd/pull/1800#discussion_r2337943658)) which I think is somewhat surprising behavior to end users and feels somewhat arbitrary

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,279 @@
...
+1. **All weights are 0**: Returns empty string (maintains current behavior)
+2. **Negative weights**: Treated as 0 (maintains current validation behavior)
+3. **Single variant**: Receives all 100,000 buckets regardless of weight value
+4. **Empty variants**: Returns error (maintains current validation behavior)
+5. **Weight sum overflow**: Returns validation error with clear message
+6. **Multiple variants with minimum allocation**: Excess distributed fairly among largest variants
+
+### Maximum Weight Considerations
+
+To ensure cross-language compatibility, we establish a maximum total weight sum equal to the maximum 32-bit signed integer value (2,147,483,647). This limit:
```
toddbaert: I think this is a good limit, regardless of [this](https://github.com/open-feature/flagd/pull/1800/files#r2445829048).

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: I also feel conflicted about this, if we were to go forward with a strictly defined max bucket size.

TBH I'm not sure the special handling is worth the possible user confusion in this extreme case.

The other obvious solution is to add a warning at evaluation time (we do similar things for other rules, like invalid semver params)

docs/architecture-decisions/high-precision-fractional-bucketing.md
beeme1mr: We must have a static bucket size, or we lose determinism.

Here's an example showing a static bucket:
image


And here's one showing a dynamic bucket size:
image

The important part is that the bucket used must always remain the same, regardless of the configuration.

beeme1mr: I need to think about this a bit more. I don't think my second image is accurate because, despite the bucket sizes changing, the location across the distribution should be consistent. I'll need to run a few tests to see which approach is better.

cupofcat: If I understand the challenge here properly (and I am not 100% sure so please correct me), we want to ensure that, for a given value "X", it's always distributed into the same bucket no matter how we express the buckets proportion. So, [0.3, 0.5, 0, 0.2], [3, 5, 2], [9, 15, 6], etc should all work the same, across all platforms, all languages.

Additionally, we would like to ensure that very skewed distributions (e.g. [0.1, 1000000]) don't end up simplifying some buckets to 0.

Lastly, we need to be cautious about floating point arithmetic.

Overall, this is a pretty complex problem :D

To meet all these requirements, I think, we need to implement some sort of integer-based bucket normalization to get the canonical representation of the buckets proportion (e.g. the examples above would all normalize to [3, 5, 2]). To achieve that we might need the following:
1. Use exact precision libraries (e.g. Decimal) when parsing the buckets rather than built-in float types (not sure if that exists for all our languages).
2. If non-integer buckets are present, scale everything to integers
3. Calculate the Greatest Common Divisor and normalize using that
4. Use some advanced algorithm to translate the hash into the correct range (using simple mod operation might skew the distribution).
5. We probably need to do scaling and normalization and subsequent bucket calculations using Bigints and not standard int32/int64 types (cause of overflows).

The challenge with that approach though is that now we can have the sum of all the buckets be larger than the maximum hash... There are ways around this (that would require using a different approach to "hashing" or downscaling the buckets to fit in int) or we can just say that such cases constitute invalid inputs (I think that saying that the minimum valid resolution is 0.001% essentially guarantees that?).

toddbaert: Ya I think you understand the challenge.

I spoke with @beeme1mr a bit about this, and we suspect my proposal [here](https://github.com/open-feature/flagd/pull/1800#discussion_r2445829048) might solve most of our concerns. I don't think we "lose determinism" in any substantially different way than alternatives.

We can lean on JSON schema to mark any non-integer inputs as invalid - we don't really need to support decimals here... as long as we give users the ability to describe relative weights (this can be done with ints, obviously) we can sidestep that mess, IMO.

Overflows might still be a concern for high numbers, but we can also specify and document limits to the total weight and error if we exceed that - that might be better than calculating a GCD, which adds a performance cost that's not going to be necessary in most cases (I'm quite confident most people will just use configs with weights like `1/1/1` or `50/50` etc).

toddbaert: I'm going to do a little PoC for this.

cupofcat: Even if we get rid of floats, in your proposal, how would we deal with same proportions but different values, e.g. [3, 5, 2], [9, 15, 6]?

toddbaert: > Even if we get rid of floats, in your proposal, how would we deal with same proportions but different values, e.g. [3, 5, 2], [9, 15, 6]?

I've re-read the conversation above, and now I'm less sure we are understanding each other. :sweat_smile: 

Let me try to level-set:

@beeme1mr 's concern with my proposal (I think) is the amount of "reassignment" arbitrary users experience when the configuration is changed (we want to minimize that if possible). He believes my proposal makes it worse, but I'm not sure that's the case. The benefit of my proposal is that we sidestep completely the things he's outlined in the `Minimum Allocation Guarantee` and `Excess Bucket Management`. These are artifacts of a constant bucket size. I'm proposing a dynamic bucket size, which I think satisfies the requirements (though I haven't mathematically or experimentally proven it).

Maybe there's some edge case where `[2,3]` would behave differently than `[4,6]`, with what I've proposed, but I don't see it, and I do not even think we need to make assurances that `[2,3]` / `[4,6]` would behave _exactly_ the same.

toddbaert: @chrfwow developed a rough test for this. I played with it a bit and at least to me, in realistic usage scenarios, @beeme1mr 's algorithm has ~25-35% less thrashing than mine.

@chrfwow and I spoke for a while about this topic, and although we initially preferred my proposal, we realized it brushes up against some limitations as well - namely that there's a practical limitation to the size of the max bucket count when calculated dynamically anyway; specifically, the maximum size of the numeric type used to represent it. With this in mind, we are inclined to suggest a modified version of @beeme1mr 's proposal, with the max bucket count simply being the max 32-bit signed int size (2,147,483,647). This is likely to satisfy anyone's precision needs. Additionally, by using this, we also sidestep the `Minimum Allocation Guarantee` - the proposal [already considers a total of weights greater than this value invalid](https://github.com/open-feature/flagd/pull/1800/changes#diff-844f86b50491b140561b7e6db409ea1b0e06e34f6f40049c4b0027ed3849e1d5R108) so we cannot have the "round to zero" problem.

As far as performance, I don't think it's worth considering cost here. Dividing with a large 32bit divisor can in fact actually be faster than by a small 32bit divisor in x86, since the latency is mostly determined by the number of bits in the quotient... but the difference in either case is negligible, and the memory cost obviously the same.

toddbaert:
I've significantly amended this proposal after some discussions with @chrfwow . I think this new proposed solution gives us the power we want while keeping things very simple and sidestepping a lot of complexity. Please have a look.

Here is the [diff](https://github.com/open-feature/flagd/pull/1800/changes/8bf04e9db4bd69f10e15fd24fbb9e6e942baf384#diff-844f86b50491b140561b7e6db409ea1b0e06e34f6f40049c4b0027ed3849e1d5R41), if you're interested in the change from the last proposal.

toddbaert:
> Looks good to me, but let's keep using percentiles mainly in the examples / docs, as it can be a bit hard to understand the max int at first.

Fully agree. I think if people want more granularity they will ask or dig deeper, but let's not confuse basic use-cases by exposing these details unnecessarily.

toddbaert:
@tangenti @cupofcat any points against this proposal? Please see my most recent updates [here](https://github.com/open-feature/flagd/pull/1800#issuecomment-3819808639), I think this is a fairly elegant solution.

tangenti:
@cupofcat  Could you take another look of the proposal and ensure it's compatible with the changes that will be introduced by the non-string fractional?

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,126 @@
...
+const bucketCount = math.MaxInt32 // 2,147,483,647
+
+func distributeValue(value string, feDistribution *fractionalEvaluationDistribution) string {
+    if feDistribution.totalWeight == 0 {
+        return ""
+    }
+
+    hashValue := int32(murmur3.StringSum32(value))
+    hashRatio := math.Abs(float64(hashValue)) / math.MaxInt32
+    bucket := int64(hashRatio * float64(feDistribution.totalWeight))
```
cupofcat: Hmm, would this not fail?
1. Input: A string hashes to 0x80000000 (2,147,483,648).
2. Cast: int32(0x80000000) wraps to -2,147,483,648 (MinInt32).
3. Abs: math.Abs(float64(MinInt32)) returns 2,147,483,648.0.
4. Division: 2,147,483,648.0 / 2,147,483,647.0 (MaxInt32) results in ~1.00000000046.
5. Result: bucket is calculated as > totalWeight.

Outcome: The loop if bucket < rangeEnd never evaluates to true because bucket exceeds the sum of all weights. The function returns "".

Also, In Java, Math.abs(Integer.MIN_VALUE) returns Integer.MIN_VALUE (negative) due to overflow.

However, I think we can completely sidestep this with my main comment.

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,126 @@
...
+- **Option 3: 1,000,000 buckets (0.0001% precision)** - 1 in every 1,000,000 users, likely overkill and could impact performance
+- **Option 4 (Favored): Max 32-bit signed integer buckets** - Use `math.MaxInt32` (2,147,483,647) as the bucket count, equal to the maximum allowed weight sum. This naturally sidesteps minimum allocation guarantees and excess bucket handling
+
+## Proposal: Max Int32 Bucket Count (Favored)
+
+> **Amendment (2026-01-29):** This alternative is now favored over the original 100,000-bucket proposal.
+
+### Rationale
+
+After experimentation comparing static vs dynamic bucket sizes, and considering implementation complexity, a simpler approach emerged: use the maximum 32-bit signed integer value (`math.MaxInt32` = 2,147,483,647) as the bucket count.
```
cupofcat: Hmm, I think the wording could be improved. The buckets count is dynamic, based on sum of weights, no? Or am I misunderstanding?

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Requirements
+
+- Support traffic allocation precision down to 0.001% (3 decimal places)
+- Maintain backwards compatibility with existing weight-based API
+- Preserve deterministic bucketing behavior (same hash input always produces same bucket)
+- Ensure consistent bucket assignment across different programming languages
+- Support weight values up to a reasonable maximum that works across multiple languages
+- Maintain current performance characteristics
+- Prevent users from being moved between buckets when only distribution percentages change
```
cupofcat, line 40: Can you expand on that? If currently the buckets are [10, 20, 30] and my user_id hashed to equivalent of 10 and the new buckets are [10, 15, 35], I should stay int the same bucket? But if the new buckets are [10, 20, 40], then I will be moved, right?

docs/architecture-decisions/high-precision-fractional-bucketing.md
```diff
@@ -0,0 +1,126 @@
...
+
+```json
+"fractional": [
+  { "cat": [{ "var": "$flagd.flagKey" }, { "var": "email" }] },
+  ["red", 50],
+  ["blue", 30], 
+  ["green", 20]
+]
+```
+
```
cupofcat, line 168: * Constraint: The sum of all variant weights must not exceed Math.MaxUint32 (4,294,967,295).
* Constraint: Weights must be defined as Integers.

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: This requirement is a bit poorly stated. It should just say something like "prefer avoiding un-neccessary re-assignments". But with our current proposal, I don't think there's any impact on this compared to what we have now.

I will update it.

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: Very astute observation.

toddbaert:
> Hi, I have two main suggestions:
> 
> 1. To make this proposal compatible with the non-string fractional and consistent hashing, we should explicitly mention it and the examples should not rely on string anymore, and string hashing.
> 2. I think we can completely sidestep any floating point shenanigans by working in uInt64 space and using multiplication?
> 
> So, the main example from the ADR, assuming the above, would look like this:
> 
> ```
> // distributeValue accepts the hash calculated by the "Harden Hashing" ADR logic.
> // It relies purely on integer math.
> func distributeValue(hashValue uint32, feDistribution *fractionalEvaluationDistribution) string {
>     // 0. Validation: Handle empty distribution
>     if feDistribution.totalWeight == 0 {
>         return ""
>     }
> 
>     // 1. Logic Integration: Use the hash provided by the non-fractional ADR.
>     //    Do NOT call StringSum32(value).
> 
>     // 2. Projection: Map 32-bit hash to [0, totalWeight)
>     //    We cast to uint64 to ensure the multiplication does not overflow.
>     //    Shifting right by 32 bits is mathematically equivalent to dividing by 2^32.
>     //    This logic is safe across major languages because it relies on fundamental binary operations.
>     bucket := (uint64(hashValue) * uint64(feDistribution.totalWeight)) >> 32
> 
>     // 3. Selection: Find which variant range the bucket falls into
>     var rangeEnd int64 = 0
>     for _, variant := range feDistribution.weightedVariants {
>         rangeEnd += int64(variant.weight)
>         
>         if int64(bucket) < rangeEnd {
>             return variant.variant
>         }
>     }
> 
>     // Should be unreachable given strict validation of weights (no float, sum < MaxInt32)
>     panic(...)
> }
> ```

@cupofcat wow what a nice idea. I've updated the ADR to use this!

I think, though, that we should keep the "max weight sum" to MaxInt32 (2,147,483,647), not MaxUint32 (4,294,967,295). My reasoning for this is I believe it means we can use Java's `long` type to handle the intermediate product, avoiding any non-standard numeric types there (please check my math)

In either case, we will need a `BigInt` in JS, but they are de-factor standard now and I think we can all agree JS' numbers are lacking...

I am going to add this table to the ADR as a survey:

| Language | 64-bit Type | Max Value | Max Product Fits? |
|----------|-------------|-----------|-------------------|
| Go | `uint64` | 1.8 × 10¹⁹ | ✅ Yes |
| Java | `long` (signed) | 9.22 × 10¹⁸ | ✅ Yes |
| JavaScript | `Number` | 9.0 × 10¹⁵ (safe) | ❌ No — must use `BigInt` |

**Max product:** `MaxUint32 × MaxInt32` = 9.22 × 10¹⁸ (fits within Java's `long` with ~6 billion headroom)

@cupofcat I've added all your suggestions in [this commit](https://github.com/open-feature/flagd/pull/1800/changes/7e4570a50cd881f4344db06ad6f63c276a3f5246), with the exception of keeping the max weight as `math.MaxInt32` so we can use Java `long` for the intermediate product.

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: See my other comment. I love your algorithm but I think we need to keep this input and max weight to MaxInt32 not MaxUint32, or else it becomes an issue with Java. I don't think people will care much about losing this range... we are already approaching a max-ratio close to 1vs(total-population-of-Earth) :joy:

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: Changed to:

> After experimentation comparing static vs dynamic bucket sizes, and considering implementation complexity, a simpler approach emerged: use the maximum 32-bit signed integer value (`math.MaxInt32` = 2,147,483,647) as the maximum allowed weight sum.

docs/architecture-decisions/high-precision-fractional-bucketing.md
toddbaert: Added

```yaml
# Constraint: The sum of all variant weights must not exceed math.MaxInt32 (2,147,483,647).
# Constraint: Weights must be defined as Integers (can be enforced by JSON schema).
"fractional": [
  { "cat": [{ "var": "$flagd.flagKey" }, { "var": "email" }] },
  ["red", 50],
  ["blue", 30], 
  ["green", 20]
]
```

cupofcat: Haha, SGTM :)

PR #1805, opened 2025-09-29
feat(docs): add documentation for HTTP OAuth

### LINKED ISSUE #1785: [FEATURE] Add OAuth2 client credentials flow authorization scheme

thisthat:
## This PR

- adds docs for https://github.com/open-feature/flagd/pull/1791

### Related Issues

Fixes #1785

### DIFF
--- a/docs/reference/sync-configuration.md
+++ b/docs/reference/sync-configuration.md
@@ -78,7 +78,7 @@ Sync providers:
 - `file` - config/samples/example_flags.json
 - `fsnotify` - config/samples/example_flags.json
 - `fileinfo` - config/samples/example_flags.json
-- `http` - <http://my-flag-source.com/flags.json>
+- [`http`](#http-configuration) - <http://my-flag-source.com/flags.json>
 - `https` - <https://my-secure-flag-source.com/flags.json>
 - `kubernetes` - default/my-flag-config
 - `grpc`(insecure) - grpc-source:8080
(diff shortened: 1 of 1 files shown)


docs/reference/sync-configuration.md
```diff
@@ -139,3 +139,43 @@ sources:
...
+  "timeoutS": 10,
+  "oauth": { 
+    "clientID": "test", 
+    "clientSecret": "test", 
+    "tokenURL": "http://localhost:8180/sso/oauth2/token" 
+  }}]'
+```
+
+When deploying Flagd in Kubernetes, you can securely manage the secrets from the file system. In this case, the client id and secret
+will be read from the files `client-id` and `client-secret`, respectively.
```
alexandraoberaigner:
```suggestion
will be read from the files `client-id` and `client-secret`, respectively. If the `folder` attribute is set, client id and secret on top level will be ignored.
```
Might make sense to point this out similar to my suggestion as Mike commented [here](https://github.com/open-feature/flagd/pull/1791/files#r2332325563).

docs/reference/sync-configuration.md
```diff
@@ -139,3 +139,43 @@ sources:
...
+  "interval": 1,
+  "timeoutS": 10,
+  "oauth": { 
+    "clientID": "test", 
+    "clientSecret": "test", 
+    "tokenURL": "http://localhost:8180/sso/oauth2/token" 
+  }}]'
+```
+
+When deploying Flagd in Kubernetes, you can securely manage the secrets from the file system. In this case, the client id and secret
```
alexandraoberaigner: [Q] Couldn't one also use the file system for storage of the secrets if flagd is not deployed on K8s?

docs/reference/sync-configuration.md
```diff
@@ -139,3 +139,43 @@ sources:
   - uri: azblob://my-container/my-flags.json
     provider: azblob
 ```
+
+### HTTP Configuration
+
+The HTTP Configuration also supports OAuth that allows to securely fetch feature flag configurations from an HTTP endpoint
+that requires OAuth-based authentication.
+To enable OAuth, you need to update your Flagd configuration setting the `oauth` object which contains parameters to configure
```
alexandraoberaigner, line 150:
```suggestion
#### CLI-based OAuth Configuration
To enable OAuth, you need to update your Flagd configuration setting the `oauth` object which contains parameters to configure

....

#### File-based OAuth Configuration
```
I don't know how to call the not-file-based section best, maybe you have a better naming idea, but I'd suggest having 2 section headings to separate the different variants of configurations 

just a nit, not a blocker for me :)

docs/reference/sync-configuration.md
thisthat: yes, you can 👍

alexandraoberaigner: then we should phrase maybe as an example:

```suggestion
Secrets can also be managed from the file system. This can be handy when, for example, deploying Flagd in Kubernetes. In this case, the client id and secret
```

docs/reference/sync-configuration.md
toddbaert: @alexandraoberaigner I changed this from "normal" to "CLI-based".

toddbaert (approved):

PR #1811, opened 2025-10-09
feat: allow null flagSetId Selector, restrict Selector to single key-value-pairs (#1708)

alexandraoberaigner:
## This PR
 
This pull request updates the `Selector` logic in the flag store to allow selecting flags without a flagSetId (`flagSetId=`).

It also restricts selector expressions to a single key-value pair, and updates / skips related tests. This is a temporary limitation pending a decision on the selector syntax (#1708). 

### Related Issues

#1708

### Follow-up Tasks
TODO: we should decide on the selector syntax

### How to test

* run `make run-flagd-selector-demo`
* experiment with queries & flag configurations with/without flagSetIds (be aware that the top level flagSetId is setting the flagSetId for all flags without already exisiting flagSetId):
```
grpcurl -d '{"selector":"flagSetId=example"}' -import-path schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags | jq
grpcurl -d '{"selector":"flagSetId="}' -import-path schemas/protobuf/flagd/sync/v1/ -proto sync.proto -plaintext localhost:8015 flagd.sync.v1.FlagSyncService/SyncFlags | jq
```

### DIFF
--- a/core/pkg/store/store_test.go
+++ b/core/pkg/store/store_test.go
@@ -379,7 +379,8 @@ func TestGetAllNoWatcher(t *testing.T) {
 
 	sourceASelector := NewSelector("source=" + sourceA.Name)
 	flagSetIdCSelector := NewSelector("flagSetId=" + flagSetIdC)
-	flagSetIdAndCSelector := NewSelector("flagSetId=" + flagSetIdC + ",source=" + sourceC.Name)
+	// #1708 Until we decide on the Selector syntax, only a single key=value pair is supported
+	//flagSetIdAndCSelector := NewSelector("flagSetId=" + flagSetIdC + ",source=" + sourceC.Name)
 
 	t.Parallel()
 	tests := []struct {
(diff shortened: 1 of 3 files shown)


core/pkg/store/store_test.go
```diff
@@ -633,6 +637,9 @@ func TestQueryMetadata(t *testing.T) {
 	// setup initial flags
 	store.Update(sourceA, sourceAFlags, model.Metadata{})
 
+	// #1708 Until we decide on the Selector syntax, only a single key=value pair is supported
+	t.Skip()
```
guidobrei: I think we do not want to skip the QueryMetadata tests, but adjust the selector.

toddbaert: Yes, I agree with @guidobrei here - lets just update the test to only have a single k/v pair (maybe one each for `flagSetId` and `source`) and add a different TODO to add back another test with multiple k/v pairs once we support that.

toddbaert:
This looks like the minimum workable implementation until we decide on a more robust Selector syntax. I'm not sure we even want to document this for now, but IMO it's releasable and will allow us to release something now without regretting it.

I'm ready to approve except I agree with @guidobrei 's comment [here](https://github.com/open-feature/flagd/pull/1811#discussion_r2417087765).

core/pkg/store/store_test.go
alexandraoberaigner: done with [49f573b](https://github.com/open-feature/flagd/pull/1811/commits/49f573b3b4ac6cb663177c06d7589787b2429833)

toddbaert:
I'll merge this tomorrow unless I hear objections.

PR #1818, opened 2025-10-30
docs: fatal codes, re-init, and retry policy

toddbaert:
This PR specifies some provider behavior, specifically around stream health, gRPC retry policy, and FATAL codes.

Specifically, it:

- publishes a retry policy that is shall be used by all flagd providers
- specifies a new option for marking some gRPC status codes as FATAL, which will cause the provider to stop attempting to reconnect (generally useful and requested in https://github.com/open-feature/go-sdk-contrib/issues/756)
- makes clear via state diagram that flagd provider should support re-initialization (if not in FATAL state)


docs/reference/specifications/providers.md
```diff
@@ -64,18 +64,21 @@ stateDiagram-v2
...
+    READY --> NOT_READY: shutdown
     STALE --> ERROR: disconnect period >= retry grace period
+    STALE --> NOT_READY: shutdown
     ERROR --> READY: reconnected
-    ERROR --> [*]: shutdown
+    ERROR --> NOT_READY: shutdown
+    ERROR --> [*]: Error code == PROVIDER_FATAL
 
-    note right of STALE
+    note left of STALE
```
toddbaert, line 74: old:

Image

new: 

Image

The main different is we make it clear transitions are possible from non-fatal `ERROR`, back to `NOT_READY`... many implementations already support this, but not all.
I think it makes sense to specify this so we can be consistent.

docs/reference/specifications/providers.md
```diff
@@ -101,25 +104,47 @@ stateDiagram-v2
...
+flagd leverages gRPC built-in retry mechanism for all RPCs.
+In short, the retry policy attempts to retry all RPCs which return `UNAVAILABLE` or `UNKNOWN` status codes 3 times, with a 1s, 2s, 4s, backoff respectively.
+No other status codes are retried.
+The flagd gRPC retry policy is specified below:
 
-When disconnected, if the time since disconnection is less than `retryGracePeriod`, the provider emits `STALE` when it disconnects.
-While the provider is in state `STALE` the provider resolves values from its cache or stored flag set rules, depending on its resolver mode.
-When the time since the last disconnect first exceeds `retryGracePeriod`, the provider emits `ERROR`.
-The provider attempts to reconnect indefinitely, with a maximum interval of `retryBackoffMaxMs`.
+```json
```
toddbaert, line 120: This is standard retryPolicy, accepted in this JSON format by most gRPC implementations.

docs/reference/specifications/providers.md
```diff
@@ -262,28 +287,29 @@ precedence.
...
+| offlineFlagSourcePath | FLAGD_OFFLINE_FLAG_SOURCE_PATH | offline, file-based flag definitions, overrides host/port/targetUri                                             | string                       | null                          | file                    |
+| offlinePollIntervalMs | FLAGD_OFFLINE_POLL_MS          | poll interval for reading offlineFlagSourcePath                                                                 | int                          | 5000                          | file                    |
+| contextEnricher       | -                              | sync-metadata to evaluation context mapping function                                                            | function                     | identity function             | in-process              |
+| fatalStatusCodes      | -                              | a list of gRPC status codes, which will cause streams to give up and put the provider in a PROVIDER_FATAL state | array                        | []                            | rpc & in-process        |
```
toddbaert: This is the only new option - the other changes are just whitespace.

docs/reference/specifications/providers.md
```diff
@@ -101,25 +104,47 @@ stateDiagram-v2
...
-The provider attempts to reconnect indefinitely, with a maximum interval of `retryBackoffMaxMs`.
+```json
+{
+    "methodConfig": [
+        {
+            "name": [
+            {
+                "service": "flagd.sync.v1.FlagSyncService",
+                "service": "flagd.evaluation.v1.Service",
+            }
```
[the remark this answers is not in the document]
toddbaert: oops, fixed.

docs/reference/specifications/providers.md
```diff
@@ -101,25 +104,49 @@ stateDiagram-v2
 
 ### Stream Reconnection
 
-When either stream (sync or event) disconnects, whether due to the associated deadline being exceeded, network error or any other cause, the provider attempts to re-establish the stream immediately, and then retries with an exponential back-off.
-We always rely on the [integrated functionality of GRPC for reconnection](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md) and utilize [Wait-for-Ready](https://grpc.io/docs/guides/wait-for-ready/) to re-establish the stream.
-We are configuring the underlying reconnection mechanism whenever we can, based on our configuration. (not all GRPC implementations support this)
+When either stream (sync or event) disconnects, whether due to the associated deadline being exceeded, network error or any other cause, the provider attempts to re-establish the stream immediately.
```
aepfli: Take this with a salt of grain (as i am not sure i understood it correctly), but there are two different things, reconnection and retry. But my knowledge might be off here, so the reconnect is happening on the channel versus the retry is for the stream. So i do think that this table might be interesting for people to see how our reconnection attempt on a lost channel looks like.

docs/reference/specifications/providers.md
```diff
@@ -101,25 +104,49 @@ stateDiagram-v2
...
-| java              | ❌                                 | ❌                        | ❌                        | 0.2    | 1.6        |
+## gRPC Retry Policy
 
-[^1] : C++, Python, Ruby, Objective-C, PHP, C#, js(deprecated)
+flagd leverages gRPC built-in retry mechanism for all RPCs.
+In short, the retry policy attempts to retry all RPCs which return `UNAVAILABLE` or `UNKNOWN` status codes 3 times, with a 1s, 2s, 4s, backoff respectively.
+No other status codes are retried.
+The flagd gRPC retry policy is specified below:
 
-When disconnected, if the time since disconnection is less than `retryGracePeriod`, the provider emits `STALE` when it disconnects.
```
aepfli, line 119: I am not sure if just adding this to the overview in the mermaid chart is sufficient enough, i think this should be also explicitly mentioned

docs/reference/specifications/providers.md
```diff
@@ -262,28 +289,29 @@ precedence.
...
+| offlineFlagSourcePath | FLAGD_OFFLINE_FLAG_SOURCE_PATH | offline, file-based flag definitions, overrides host/port/targetUri                                             | string                       | null                          | file                    |
+| offlinePollIntervalMs | FLAGD_OFFLINE_POLL_MS          | poll interval for reading offlineFlagSourcePath                                                                 | int                          | 5000                          | file                    |
+| contextEnricher       | -                              | sync-metadata to evaluation context mapping function                                                            | function                     | identity function             | in-process              |
+| fatalStatusCodes      | -                              | a list of gRPC status codes, which will cause streams to give up and put the provider in a PROVIDER_FATAL state | array                        | []                            | rpc & in-process        |
```
alexandraoberaigner:
```suggestion
| fatalStatusCodes      | FLAGD_FATAL_STATUS_CODES                             | a list of gRPC status codes, which will cause streams to give up and put the provider in a PROVIDER_FATAL state | array                        | []                            | rpc & in-process        |
```
@toddbaert @aepfli We should probably make sure the change is consistent with:
https://github.com/open-feature/flagd-testbed/pull/311/files#diff-2f3b6fc7d0eec288e7349f23f8f56b197eecf05fef9320930ee266dda60fd6e7R25

docs/reference/specifications/providers.md
toddbaert: I've added a bit about the layer 4 re-connection, but that's pretty much automatic and can't be adjusted, so I don't say that much about it: https://github.com/open-feature/flagd/pull/1818/commits/aee9e311347adb562854dd15572fc1505948e8d9

docs/reference/specifications/providers.md
toddbaert: Agreed, adding something specific about this: https://github.com/open-feature/flagd/pull/1818/commits/aee9e311347adb562854dd15572fc1505948e8d9

toddbaert:
@aepfli @alexandraoberaigner made changes from your feedback, plz re-review.

docs/reference/specifications/providers.md
tangenti: I had the impression that `PROVIDER_FATAL` can only happen during initialization, where the error can be surfaced and handled by the caller.

With the current proposal,  `PROVIDER_FATAL` can be a result of a failing sync. As a user, it seems that I'll get the default value and an error. Am I supposed to handle this error and exit the program?

toddbaert: @tangenti I need to make some updates to reflect the discussion [here](https://github.com/open-feature/go-sdk-contrib/issues/756).

We decided the best path forward is to provide an option to enumerate the status codes that a user considers FATAL. In the case those are received, whether it's the initial connection or not, the program can exit (or rebuild a new provider). We believed this was the best trade-off between usability and complexity, and it's easy to understand: select what you want to consider FATAL, and take the action you want when those codes are received; by marking a code is FATAL you are telling the provider that this code represents a non-transient error state.

I will make the related updates.

toddbaert: I've included this.

chrfwow (approved):

docs/reference/specifications/providers.md
toddbaert: Another reason to go this way is it saves us from keeping track of even more state.

PR #1820, opened 2025-10-31
docs: update flags.json

chrfwow:
## This PR
Updates the flags.json file on the website to match the current content in the schemas repo


aepfli:
~~does the yaml also need an update?~~ nvm there is no yaml in the docs

guidobrei:
I think we should also add the [flagd schema](https://github.com/open-feature/flagd-schemas/blob/main/json/flagd.json) and update the [docs](https://github.com/open-feature/flagd/blob/main/docs/reference/schema.md).

toddbaert:
FYI to all just in case, this is done in the make-target `make update-public-schema` (including the new flagd schema @guidobrei [mentioned](https://github.com/open-feature/flagd/pull/1820#issuecomment-3489975664))

I've run it again and it's added some whitespace changes, so I'm committing them as well just so we have no deltas.

toddbaert (approved):

PR #1822, opened 2025-11-04
feat: Add custom regex_match string comparison

leth:
## This PR
- Adds a new `regex_match` jsonlogic string comparison

### Notes
I don't code in golang regularly so any feedback appreciated!
Also, completely new to the project, so please let me know if there are any more areas I need to update!

### How to test
Use `regex_match` in a flag targeting rule, similar to how you would `starts_with` etc


core/pkg/evaluator/string_comparison.go
```diff
@@ -123,3 +128,130 @@ func parseStringComparisonEvaluationData(values interface{}) (string, string, er
...
+			// On config change, rotate the regex caches
+			// If the current cache is empty, do nothing, to keep the previous cache intact in case it is still helpful
+			if len(*self.RegexCache) == 0 {
+				continue
+			}
+			self.PrevRegexCache = self.RegexCache
+			self.RegexCache = &map[string]*regexp.Regexp{}
+		}
+	}()
+	selector := store.NewSelector("")
```
leth, line 159: I couldn't figure out what the correct selector value should be here!

aepfli:
Hey thanks for the contribution. As this is adding this logic to flagd itself, a lot of providers also support in-process evaluation. Hence i think we should create an issue, to track all those changes needed for the providers and mostly to have one place for discussions etc. Can you maybe create an issue, or if there is already an issue link it please?

leth:
> Hey thanks for the contribution. As this is adding this logic to flagd itself, a lot of providers also support in-process evaluation. Hence i think we should create an issue, to track all those changes needed for the providers and mostly to have one place for discussions etc. Can you maybe create an issue, or if there is already an issue link it please?

Thanks @aepfli, I had a quick look before but didn't find one; I'll raise one later!

Do providers have the ability to identify when they don't support a particular targeting rule and need to fall-back to the server for evaluation?

leth:
Can anyone advise on the SQ code duplication problem: Would you prefer the string comparison tests refactored for less duplication, or would it be better to keep it as-is to keep it simple?

If it's the latter, do we need to put that file on the sonar cpd exclusion list?

aepfli:
> > Hey thanks for the contribution. As this is adding this logic to flagd itself, a lot of providers also support in-process evaluation. Hence i think we should create an issue, to track all those changes needed for the providers and mostly to have one place for discussions etc. Can you maybe create an issue, or if there is already an issue link it please?
> 
> Thanks @aepfli, I had a quick look before but didn't find one; I'll raise one later!
> 
> Do providers have the ability to identify when they don't support a particular targeting rule and need to fall-back to the server for evaluation?

We do one or the other, and we do not fallback to server communication if we do in-process. Hence i recommended to create a ticket. As this is a bigger change throughout the ecosystem, which we need to discuss and evaluate.

toddbaert:
I strongly recommend we don't attempt this. Regex is notoriously inconsistent, and I don't think we can reliably implemented this across all our supported runtimes without _substantial_ effort. Even "simple" patterns can differ in Unicode handling, anchoring defaults, multiline semantics, etc. Then there's also ReDoS to consider... If we could centralize on a single implementation that ran everywhere, the proposal might be more reasonable; I would consider that almost a pre-requisite.

toddbaert:
I'm closing this for now. If anyone strongly disagrees, feel free to reopen. We can perhaps revisit this later if/when we unify our evaluation engine impl.

PR #1821, opened 2025-10-31
test: using e2e framework to test flagd via go-sdk-contrib flagd provider

aepfli:
depends on:
- https://github.com/open-feature/flagd-testbed/pull/308
- https://github.com/open-feature/go-sdk-contrib/pull/744

PR #1827, opened 2025-11-27
fix: fixing sync return format missing flag layer, adding full e2e suite

aepfli:
with https://github.com/open-feature/flagd/pull/1797 we introduced this bug, that the format of the response is not correct.

current state:
```javacript
{ 
    flagConfiguration: {/* flag object */}
}
```
should be:
```javacript
{ 
    flagConfiguration: {
       flags: {/* flag object */}
    }
}
```

Clarification from @toddbaert - this is an UNRELEASED bug so far.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -124,5 +124,5 @@ jobs:
         with:
           go-version-file: 'flagd/go.mod'
 
-      - name: Run evaluation test suite
+      - name: Run test suite
         run: make workspace-clean && cd test/integration && go clean -testcache && go test -cover


aepfli:
we definitely need to get everything going for https://github.com/open-feature/flagd/pull/1821 to prevent this issue in the future

toddbaert:
Yep, :100: I found this late last week and was intending to open a bug as well. We need to enhance our testing to run in-process tests to find this - there's much less surface area to test with in-process mode on flagd itself (since evaluation happens in the workload) but it's required to find things like this.

Thanks!

test/integration/integration_test.go
```diff
@@ -1,13 +1,183 @@
...
+	})
+	defer runner.Cleanup()
+
+	// Define feature paths - using flagd-testbed gherkin files
+	featurePaths := []string{
+		"./",
+	}
+
+	// Run tests with InProcess-specific tags
+	tags := "@in-process && ~@unixsocket&& ~@metadata && ~@contextEnrichment && ~@customCert && ~@forbidden && ~@sync-port && ~@sync-payload && ~@grace && ~@targetURI"
```
aepfli: i am not sure why targetURI is failing here, in the provider it is working fine, did we maybe add another regression somehow?

aepfli: the test runs fine with this changes as e2e test in the flagd-golang-provider. so i assume that this might be some kind of strange execution issue. but not worth further exploration

aepfli: okay updated the go moduels and now it seems to work as intended

test/integration/integration_test.go
alexandraoberaigner, line 1: [Suggestion] maybe we could use a similar naming/file structure as in the go flagd provider to make it easier to find things? eg. have put rpc, file, and inprocess test functions in separate files. wdyt?

test/integration/go.mod
alexandraoberaigner, line 1: [Q] are these new imports all required for the testcontainers? 👀

.github/workflows/build.yaml
```diff
@@ -124,32 +124,5 @@ jobs:
...
-          -f file:${{ github.workspace }}/test-harness/flags/custom-ops.json \
-          -f file:${{ github.workspace }}/test-harness/flags/evaluator-refs.json \
-          -f file:${{ github.workspace }}/test-harness/flags/zero-flags.json \
-          -f file:${{ github.workspace }}/test-harness/flags/edge-case-flags.json &
-
-      - name: Run envoy proxy in background
-        run: |
-          envoy -c ./test/integration/config/envoy.yaml &
-
       - name: Run evaluation test suite
```
alexandraoberaigner:
```suggestion
      - name: Run test suite
```
[Q] Its not only the evaluation test suite anymore, is it? Should we rename this? Open for other names too

toddbaert (approved):
Thanks for the fix, but thanks way more for integrating the entire e2e suite. Very very good to have.

toddbaert:
@aepfli ... @alexandraoberaigner has some small questions/nits, but not blockers for me. I've rebased, so I'll merge this _soon_ unless you want to address those small nits.

PR #1838, opened 2025-12-19
chore: Fix oauth docs issue

guidobrei:
## This PR
fixes an issue in documentation about using OAuth with HTTP Syncs


toddbaert:
Small docs issue so merging with override.

PR #1854, opened 2026-01-27
feat: add ETag/caching support in OFREP

toddbaert:
Implement caching in bulk evaluations as per OFREP spec.

See: https://openfeature.dev/docs/reference/other-technologies/ofrep/openapi

This is a requirement for deprecating the flagd-web-provider, as without it, polling would be substantially more expensive.

Relates to: https://github.com/open-feature/js-sdk-contrib/pull/1459

### DIFF
--- a/docs/reference/flagd-ofrep.md
+++ b/docs/reference/flagd-ofrep.md
@@ -25,7 +25,7 @@ curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags'
 
 ## HTTP Caching
 
-The bulk evaluation endpoint supports HTTP caching via ETags (SHA1-based). Clients can use the `If-None-Match` header with a previously received `ETag` to validate cached responses. When the response hasn't changed, the server returns `304 Not Modified` without a body, reducing bandwidth.
+The bulk evaluation endpoint supports HTTP caching via ETags (MD5-based). Clients can use the `If-None-Match` header with a previously received `ETag` to validate cached responses. When the response hasn't changed, the server returns `304 Not Modified` without a body, reducing bandwidth.
 
 ```shell
 curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags' \
--- a/flagd/pkg/service/flag-evaluation/ofrep/handler.go
+++ b/flagd/pkg/service/flag-evaluation/ofrep/handler.go
@@ -2,7 +2,7 @@ package ofrep
 
 import (
 	"context"
-	"crypto/sha1"
+	"crypto/md5"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -194,7 +194,7 @@ func calculateETag(response ofrep.BulkEvaluationResponse) (string, []byte, error
 		return "", nil, fmt.Errorf("failed to marshal response for ETag calculation: %w", err)
 	}
 
-	hash := sha1.Sum(data)
+	hash := md5.Sum(data)
 	return fmt.Sprintf("\"%x\"", hash), data, nil
 }
 


flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -137,17 +174,30 @@ func (h *handler) writeJSONToResponse(status int, payload interface{}, w http.Re
...
+func calculateETag(response ofrep.BulkEvaluationResponse) (string, []byte, error) {
+	// marshal the response to JSON
+	data, err := json.Marshal(response)
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to marshal response for ETag calculation: %w", err)
 	}
+
+	// Calculate SHA256 hash of the JSON response
+	hash := sha256.Sum256(data)
+	return fmt.Sprintf("\"%s\"", hex.EncodeToString(hash[:])), data, nil
```
erka:
```suggestion
	return fmt.Sprintf("\"%x\"", hash), data, nil
```

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -137,17 +174,30 @@ func (h *handler) writeJSONToResponse(status int, payload interface{}, w http.Re
...
 	w.WriteHeader(status)
 	_, err = w.Write(marshal)
 	if err != nil {
-		h.Logger.Warn(fmt.Sprintf("error while writing response: %v", err))
+		h.Logger.Warn("error while writing response", zap.Error(err))
+	}
+}
+
+// calculateETag generates an ETag from the bulk evaluation response
+func calculateETag(response ofrep.BulkEvaluationResponse) (string, []byte, error) {
```
leakonvalinka, line 190: Q: For the ETag calculation the flags must be in the same order every time, is this guaranteed?

toddbaert: Great point!

Ordering in maps in Go is intentionally non-deterministic so that devs cannot rely on it, but the JSON marshalling in Go does sort them:

https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L183-L189

So we can rely on that.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
toddbaert: Nice.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
leakonvalinka: Okay good to know, thank you!

toddbaert: TIL as well (the json marshall ordering bit). I know maps were purposely random in Go, so I was quite scared for a second.

toddbaert:
> Overall it looks great. I’m not sure whether SHA-256 is really required here, and it could add extra CPU load. SHA-1 or [xxHash](https://pkg.go.dev/github.com/cespare/xxhash/v2) might be just as good for this use case.

I went back and forth on this as well. SHA-1 was my first thought. It's not spec'd or anything so we could change it later... I will go back to sha1 and save some CPU cycles.

toddbaert:
> > Overall it looks great. I’m not sure whether SHA-256 is really required here, and it could add extra CPU load. SHA-1 or [xxHash](https://pkg.go.dev/github.com/cespare/xxhash/v2) might be just as good for this use case.
> 
> I went back and forth on this as well. SHA-1 was my first thought. It's not spec'd or anything so we could change it later... I will go back to sha1 and save some CPU cycles.

Did this and also added some minor docs.

docs/reference/flagd-ofrep.md
```diff
@@ -23,4 +23,13 @@ To evaluate all flags currently configured at flagd, use OFREP bulk evaluation r
 curl -X POST 'http://localhost:8016/ofrep/v1/evaluate/flags'
 ```
 
+## HTTP Caching
+
+The bulk evaluation endpoint supports HTTP caching via ETags (SHA1-based). Clients can use the `If-None-Match` header with a previously received `ETag` to validate cached responses. When the response hasn't changed, the server returns `304 Not Modified` without a body, reducing bandwidth.
```
jonathannorris: FYI seems like MD5 is the most common for etags: https://www.fastly.com/blog/etags-what-they-are-and-how-to-use-them

They are a bit faster to generate and smaller than SHA1, but just as broken as well :)

jonathannorris: But yea probably should just use `SHA‑256` for correctness and pay the performance penalty, or let it be configurable with `SHA‑256` as the default.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -128,7 +134,37 @@ func (h *handler) HandleBulkEvaluation(w http.ResponseWriter, r *http.Request) {
...
-		h.writeJSONToResponse(http.StatusOK, ofrep.BulkEvaluationResponseFrom(evaluations, metadata), w)
+		response := ofrep.BulkEvaluationResponseFrom(evaluations, metadata)
+		h.writeBulkEvaluationResponse(w, r, response)
+	}
+}
+
+// writes the bulk evaluation response with ETag support
+func (h *handler) writeBulkEvaluationResponse(w http.ResponseWriter, r *http.Request, response ofrep.BulkEvaluationResponse) {
+	// calculate ETag and marshal response in one operation
+	eTag, body, err := calculateETag(response)
```
jonathannorris, line 145: Doesn't the ETag need to be generated based on the request + ruleset, not the response, so that an evaluation can be avoided on further requests?

toddbaert: This is client side Etag hashing as a bandwidth saver. I don't think OFREP has the robustness or specification yet to support anything that would allow us to avoid evaluations.

docs/reference/flagd-ofrep.md
toddbaert: I liked your article. I feel like for this purpose, md5 is fine, as it says. I'm open to making in configurable later, but I would do that only if requested.

toddbaert:
Hey @erka @jonathannorris @jonathannorris @leakonvalinka If you have time, take a look at this alternative, which does something more like @jonathannorris suggested: https://github.com/open-feature/flagd/pull/1858

PR #1855, opened 2026-01-28
fix: case sensitivity in header context mapping

toddbaert:
Fixes an issue that made header -> context mapping case-sensitive, because we were using direct access instead of `headers.Get`.

Also reduces some duplication by extracting this logic, and adds test coverage.

### DIFF
--- a/docs/reference/flag-definitions.md
+++ b/docs/reference/flag-definitions.md
@@ -299,9 +299,11 @@ The evaluation context can be accessed in targeting rules using the `var` operat
 | Retrieve property from the evaluation context or use a default | `#!json { "var": ["email", "noreply@example.com"] }` |
 | Retrieve a nested property from the evaluation context         | `#!json { "var": "user.email" }`                     |
 
-> For more information, see the `var` section in the [JsonLogic documentation](https://jsonlogic.com/operations.html#var).
+See the [cheat sheet](./cheat-sheet.md#context-aware-evaluation) for practical examples of context-sensitive evaluations. For details on extracting values from context, see the `var` section in the [JsonLogic documentation](https://jsonlogic.com/operations.html#var).
 
-See the [cheat sheet](./cheat-sheet.md#context-aware-evaluation) for practical examples of context-sensitive evaluation.
+!!! tip
+
+    flagd's evaluation engine is case-sensitive, including attribute extraction using the `var` operator.
 
 #### Conditions
 


flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -160,29 +161,15 @@ func extractOfrepRequest(req *http.Request) (ofrep.Request, error) {
...
 	} else {
 		log.WarnWithID(requestID, "provided context does not comply with flagd, continuing ignoring the context")
 	}
 
-	for k, v := range staticContextValues {
-		context[k] = v
-	}
-
-	for header, contextKey := range headerToContextKeyMappings {
-		if values, ok := headers[header]; ok {
```
toddbaert, line 182: This was the root of the problem. When net/http stores headers, it automatically converts the keys to `textproto.CanonicalMIMEHeaderKey` (`user-email` becomes `User-Email`). We were then accessing those header values directly with an UN-normalized key.

Using `headers.Get(key)` instead of `headers[key]` is the simple fix for that, and I've done so in the new extracted common function.

flagd/pkg/service/flag-evaluation/context_utils.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	return merged
+}
+
+// normalizeHeaderName converts a header name to canonical HTTP header format.
+// This is primarily useful for display/logging purposes.
+func normalizeHeaderName(name string) string {
+	return strings.ToLower(name)
+}
```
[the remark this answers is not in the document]
toddbaert: Yep, I ended up not using this.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
guidobrei: Oh, that's interesting.

That means, now header value extraction is case insensitive, but context attribute matching against the ruleset stays case sensitive. Is this behavior documented in the spec?

toddbaert:
> Great improvement. We should document the behavior.

Good point. Added a note in docs.

docs/reference/flag-definitions.md
```diff
@@ -299,9 +299,11 @@ The evaluation context can be accessed in targeting rules using the `var` operat
 | Retrieve property from the evaluation context or use a default | `#!json { "var": ["email", "noreply@example.com"] }` |
 | Retrieve a nested property from the evaluation context         | `#!json { "var": "user.email" }`                     |
 
-> For more information, see the `var` section in the [JsonLogic documentation](https://jsonlogic.com/operations.html#var).
+See the [cheat sheet](./cheat-sheet.md#context-aware-evaluation) for practical examples of context-sensitive evaluations. For details on extracting values from context, see the `var` section in the [JsonLogic documentation](https://jsonlogic.com/operations.html#var).
```
toddbaert, line 302: Just rephrased this.

Issue #1856, opened 2026-01-28
[FEATURE] Graceful defaulting to provided code default


toddbaert:
### Requirements

Previously, we worked on https://github.com/open-feature/flagd/issues/1646, which was a step in the right direction. In summary, we changed the minimum flag definition to allow for the omission of the `defaultVariant`:

```jsonc
    "simple-string": {
      "state": "ENABLED",
      "variants": {
        "greeting": "Hello, World!",
        "farewell": "Goodbye, World!"
      },
      "defaultVariant": null // this can now be set to null, or omitted completely instead of a variant; indicating that the code default should be used.
    },
```

**However, this was implemented in a stop-gap manner.** In order to facilitate the code-defaulting behavior as simply as possible, our in-process providers and flagd's remote evaluation protocols behave as if such flags do not exist (an error, which triggers the code default). This is not ideal since it pollutes telemetry with errors which are not actually representative of unexpected behavior, and hurts performance by causing stack traces to be produced internally in many cases. Moreover it simply confuses developers.

We need to improve this. This will involve:

- changes to the [evaluation.v1](https://github.com/open-feature/flagd-schemas/blob/main/protobuf/flagd/evaluation/v1/evaluation.proto) to support omitting a variant/value or otherwise communicating that the code default should be used
  - may require an evaluation.v2 if impossible or impractical to accomplish in a non-breaking way
- similar changes to flagd's OFREP handler as above (OFREP already supports this, see https://github.com/open-feature/protocol/blob/main/service/adrs/0006-optional-value-for-code-defaults.md and https://github.com/open-feature/protocol/pull/49/changes
- changes to every in-process evaluator
  - [ ] Go
  - [ ] Java
  - [ ] JS
  - [ ] Python
  - [ ] .NET
- updates to the test harness [here](https://github.com/open-feature/flagd-testbed/blob/ecb3d096c82c52424eeb545cf481f2f1d4794d21/gherkin/evaluation.feature#L76-L86) (we will probably need to have a step/annotation for this to progressively migrate it)
- updates to documentation (specifically see [this section](https://flagd.dev/reference/flag-definitions/?h=defaultva#default-variant), and the note contained therein)

toddbaert:
Done in all tier 1 providers.

cc @aepfli  - not sure where the Rust co-implementation is at the moment.

PR #1862, opened 2026-02-05
feat!: no `defaultVariant` -> code default (previosuly FLAG_NOT_FOUND) 

marcozabel:
## This PR
Adds support for v2 proto schemas, which enable RPC mode to signal that code-default should be used if no `defaultVariant` is configued.
Equivalent changes added to OFREP.
In-process resolvers will be updated to have the same behavior.

Previously, this situation resulted in `FLAG_NOT_FOUND` - now it just gracefully results in a code default (no error).

### Related Issues
https://github.com/open-feature/flagd/issues/1856

### DIFF
--- a/core/pkg/evaluator/json.go
+++ b/core/pkg/evaluator/json.go
@@ -395,10 +395,10 @@ func (je *Resolver) evaluateVariant(ctx context.Context, reqID string, flagKey s
 			if flag.DefaultVariant == "" {
 				if ctx.Value(ProtoVersionKey) != nil {
 					// old proto version behavior
-					return flag.DefaultVariant, flag.Variants, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
+					return "", flag.Variants, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
 				}
 
-				return flag.DefaultVariant, flag.Variants, model.FallbackReason, metadata, nil
+				return "", flag.Variants, model.FallbackReason, metadata, nil
 			}
 
 			return flag.DefaultVariant, flag.Variants, model.DefaultReason, metadata, nil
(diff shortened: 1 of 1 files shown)


core/pkg/evaluator/json.go
```diff
@@ -380,7 +385,11 @@ func (je *Resolver) evaluateVariant(ctx context.Context, reqID string, flagKey s
 
 		if trimmed == "null" {
 			if flag.DefaultVariant == "" {
-				return "", flag.Variants, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
+				if ctx.Value("protoVersion") != nil {
```
toddbaert: Let's extract this to a constant with an explanation of what it's for (as a sentinel for the defaulting behavior).

core/pkg/evaluator/json.go
```diff
@@ -399,7 +408,10 @@ func (je *Resolver) evaluateVariant(ctx context.Context, reqID string, flagKey s
 	}
 
 	if flag.DefaultVariant == "" {
-		return "", flag.Variants, model.ErrorReason, metadata, errors.New(model.FlagNotFoundErrorCode)
+		if ctx.Value("protoVersion") != nil {
```
toddbaert: Add a comment here, about why we are doing this.

toddbaert:
@marcozabel I squashed your changes, rebased on main, and fixed conflicts. I think there are some failing unit tests but they make sense - I think you just need to update the assertions. The e2e tests are working which means you had no breaking changes (perfect). The duplication warning from sonar is expected based on keeping v1. I will not block it for this, since we will eventually remove it.

The change looks exactly right, overall. Ready to approve once the tests are fixed and the nits are taken care of. 

UPDATE:

@marcozabel @leakonvalinka I've made some small additional changes:

- updated OFREP implementation to completely omit value/variant (`omitempty` marshall option) instead of using `null`
- also omit variant/value in OFREP bulk responses if `defaultVariant` is missing
- extracted a constant
- fix a test.


see: https://github.com/open-feature/flagd/pull/1862/changes/d203d7dfb5a2b0dd915d4f637c611ab3b0f01ce0

core/pkg/evaluator/json.go
toddbaert: done

core/pkg/evaluator/json.go
toddbaert: done

leakonvalinka:
It looks good to me too, I just changed `defaultVariant` to `""` in cases where the default variant is an empty string anyways, for easier readability.

Issue #1863, opened 2026-02-05
[BUG] Improve error context in store Watch: log selector and relevant details


aepfli:
### Observed behavior

When the store Watch function fails (e.g., due to context canceled), the error log message is not helpful: it only states 'error watching flags: context canceled.' This makes troubleshooting very hard, as we miss context about which flag, selector, or parameters were involved. Example:

```
2026-02-05T09:04:51.793Z	error	store/store.go:329	error watching flags: context canceled

github.com/open-feature/flagd/core/pkg/store.(*Store).Watch.func1
/src/core/pkg/store/store.go:329
```
Relevant code paths (for reference):
- [store.go: Watch implementation](https://github.com/open-feature/flagd/blob/main/core/pkg/store/store.go#L309-L335)
- [selector usage at selectOrAll](https://github.com/open-feature/flagd/blob/main/core/pkg/store/store.go#L337-L348)
- [Selector structure](https://github.com/open-feature/flagd/blob/main/core/pkg/store/query.go#L35-L43)

Currently, when the context is canceled (or another error occurs), the log does not include selector details or other relevant contextual data.

### Expected Behavior

The error message should include additional context such as the selector (if any), flag identifiers, or other properties available at the point of failure. This would enable maintainers and users to better debug watch failures and understand what caused them. Consider logging (safely):
- The selector expression/content (if not nil)
- Any flag set ID, key, or constraints visible from the selector or iteration
- If available, metadata/key/value involved at the time of error

### Steps to reproduce

1. Run flagd and create a setup where the store Watch may fail (e.g., timeout, context cancellation).
2. Observe current logs: messages provide little context.
3. Expected: logs should indicate not only the error, but also details like selector, flag or configuration segment being watched, to facilitate root cause analysis.

---

Relevant code for context and potential improvement locations:
- [core/pkg/store/store.go#L309-L335: Watch function](https://github.com/open-feature/flagd/blob/main/core/pkg/store/store.go#L309-L335)
- [Selector available at core/pkg/store/store.go#L337-L348 (selectOrAll) and core/pkg/store/query.go#L35-L43](https://github.com/open-feature/flagd/blob/main/core/pkg/store/query.go#L35-L43)

The selector and possibly related properties exist as arguments (see `selector *Selector`). Exposing this in error logs would make debugging much easier.

leakonvalinka:
I will take a look at this 👍🏼

PR #1865, opened 2026-02-05
fix: Enhance error logs in store's Watch func

leakonvalinka:
## This PR

- improves errror logs in the store's Watch function

### Related Issues

https://github.com/open-feature/flagd/issues/1863

### DIFF
--- a/docs/architecture-decisions/high-precision-fractional-bucketing.md
+++ b/docs/architecture-decisions/high-precision-fractional-bucketing.md
@@ -1,0 +1,187 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: Michael Beemer
+created: 2025-09-10
+updated: 2026-01-29
+---
+
+# High-Precision Fractional Bucketing for Sub-Percent Traffic Allocation
+
+This ADR proposes enhancing the fractional operation to support high-precision traffic allocation down to 0.001% granularity by increasing the internal bucket count from 100 to 100,000 while maintaining the existing weight-based API.
+
+## Background
+
+The current fractional operation in flagd uses a 100-bucket system that maps hash values to percentages in the range [0, 100].
+This approach works well for most use cases but has significant limitations in high-throughput environments where precise sub-percent traffic allocation is required.
+
+Currently, the smallest allocation possible is 1%, which is insufficient for:
+
+- Gradual rollouts in ultra-high-traffic systems where 1% could represent millions of users
+- A/B testing scenarios requiring precise control over small experimental groups
+- Canary deployments where operators need to start with very small traffic percentages (e.g., 0.1% or 0.01%)
+
+The current implementation in `fractional.go` calculates bucket assignment using:
+
+```go
+bucket := hashRatio * 100 // in range [0, 100]
+```
+
+This limits granularity to 1% increments, making it impossible to achieve the precision required for sophisticated traffic management strategies.
+
+## Requirements
+
+- Support traffic allocation precision down to 0.001% (3 decimal places)
+- Maintain backwards compatibility with existing weight-based API
+- Preserve deterministic bucketing behavior (same hash input always produces same bucket)
+- Ensure consistent bucket assignment across different programming languages
+- Support weight values up to a reasonable maximum that works across multiple languages
+- Maintain current performance characteristics
+- Prevent users from being moved between buckets when only distribution percentages change
+- Guarantee that any variant with weight > 0 receives some traffic allocation
+- Handle edge cases gracefully without silent failures
+- Validate weight configurations and provide clear error messages for invalid inputs
+
+## Considered Options
+
+- **Option 1: 10,000 buckets (0.01% precision)** - 1 in every 10,000 users, better but still not sufficient for many high-throughput use cases
+- **Option 2: 100,000 buckets (0.001% precision)** - 1 in every 100,000 users, meets most high-precision needs
+- **Option 3: 1,000,000 buckets (0.0001% precision)** - 1 in every 1,000,000 users, likely overkill and could impact performance
+- **Option 4 (Favored): Max 32-bit signed integer buckets** - Use `math.MaxInt32` (2,147,483,647) as the maximum allowed weight sum. This naturally sidesteps minimum allocation guarantees and excess bucket handling
+
+## Proposal: Max Int32 Weight Sum (Favored)
+
+> **Amendment (2026-01-29):** This alternative is now favored over the original 100,000-bucket proposal.
+
+### Rationale
+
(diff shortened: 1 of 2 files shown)


core/pkg/store/store.go
```diff
@@ -326,7 +332,7 @@ func (s *Store) Watch(ctx context.Context, selector *Selector, watcher chan<- Fl
 			}
 
 			if err = ws.WatchCtx(ctx); err != nil {
-				s.logger.Error(fmt.Sprintf("error watching flags: %v", err))
+				s.logger.Error(fmt.Sprintf("context error watching flags for selector %s: %v", selectorString, err))
```
leakonvalinka: I could also add flag keys (and/or other flag information like flagSetIds or sources or metadata), but I'm not sure since it could make the log message really long. On the other hand, if there is no selector specified, there is basically no improvement to the log message, so it might make sense to still add it.

core/pkg/store/store.go
```diff
@@ -311,9 +311,15 @@ func (s *Store) Watch(ctx context.Context, selector *Selector, watcher chan<- Fl
 	go func() {
 		for {
 			ws := memdb.NewWatchSet()
+			selectorString := "<none>"
```
aepfli: I think it is worth extracting this into a method, we do not need to compute this all the time, only in an error case. hence a separate method makes sense.

core/pkg/store/store.go
aepfli: what do you think make sense, flag information is not interesting, but the metadata could offer good insights. how hard is it to add those?

aepfli (commented):
The changes are simple and nice, some questions and suggestions from my side added. especially i think we should extract the generation of the string into an own method.

core/pkg/store/store.go
leakonvalinka: Ok sounds good!

core/pkg/store/store.go
leakonvalinka: I would extract the metadata from the flags as I don't see another source for it. I'm not exactly sure what information metadata would contain and if filtering out certain attributes would be a good idea?

toddbaert:
I think the reason this came up is because some client's like us internally use stream deadlines. Stream deadlines are sent from the client, but enforced both on the client and on the server. This error will fire when the context deadline is exceeded, which is perfectly expected in our case. You can test this by connecting a client with a 10s stream deadline to flagd, and you will see this basically every 10s, I think.

We may want to log this as debug if the error is a deadline exceeded.

EDIT:

You can do this with `grpcurl`

```
grpcurl -max-time 10   -plaintext   -import-path "$PROTO_DIR" -proto flagd/sync/v1/sync.proto   -d '{"selector": "flagSetId=app-flags"}'   localhost:8015   flagd.sync.v1.FlagSyncService/SyncFlags
```

10s later:

```
2026-02-05T07:46:04.414-0500    error   store/store.go:329      error watching flags: context canceled
github.com/open-feature/flagd/core/pkg/store.(*Store).Watch.func1
        /home/todd/git/flagd/core/pkg/store/store.go:329
```

leakonvalinka:
If we go from log level error to debug, it might take users longer to notice the logs. If this is a transitive error then that's fine I think, but if it requires user/client action it would be good to have some error log. I'm not 100% grasping what the error is about so if that doesn't make sense forget about it :D

toddbaert:
> If we go from log level error to debug, it might take users longer to notice the logs. If this is a transitive error then that's fine I think, but if it requires user/client action it would be good to have some error log. I'm not 100% grasping what the error is about so if that doesn't make sense forget about it :D

Basically, if the error is one of these types, it's not an error: `ctx.Err() == context.DeadlineExceeded || ctx.Err() == context.Canceled`, it's totally expected behavior because either the client has disconnected cleanly (maybe the client is shutting down) or the client's deadline is up.

Other errors are really unexpected, but these are not (not every ergonomic, I know).

core/pkg/store/store.go
```diff
@@ -326,7 +327,11 @@ func (s *Store) Watch(ctx context.Context, selector *Selector, watcher chan<- Fl
 			}
 
 			if err = ws.WatchCtx(ctx); err != nil {
-				s.logger.Error(fmt.Sprintf("error watching flags: %v", err))
+				if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+					s.logger.Debug(fmt.Sprintf("while watching flags for selector %s: %v", selector.ToLogString(), err))
```
toddbaert:
```suggestion
					s.logger.Debug(fmt.Sprintf("context cancellation while watching flags for selector %s: %v", selector.ToLogString(), err))
```

toddbaert: I think this gets the idea across

leakonvalinka: I literally applied this 1s before you changed it lmao, here we go again

core/pkg/store/store.go
```diff
@@ -313,7 +314,7 @@ func (s *Store) Watch(ctx context.Context, selector *Selector, watcher chan<- Fl
 			ws := memdb.NewWatchSet()
 			it, err := s.selectOrAll(selector)
 			if err != nil {
-				s.logger.Error(fmt.Sprintf("error watching flags: %v", err))
+				s.logger.Error(fmt.Sprintf("error getting flags for selector %s: %v", selector.ToLogString(), err))
```
erka: logger has `WithFields` which will be the best here. Structured logging please...

erka: https://pkg.go.dev/go.uber.org/zap/zapcore?utm_source=godoc#ObjectMarshaler

toddbaert: Agreed.

Maybe we should make an additional issue to make sure were making use of that generally.

core/pkg/store/store.go
toddbaert: Sorry!

core/pkg/store/store.go
leakonvalinka: I changed the logs to that now, I also think if we do it here we should make sure to use it everywhere.

core/pkg/store/store.go
```diff
@@ -313,7 +315,7 @@ func (s *Store) Watch(ctx context.Context, selector *Selector, watcher chan<- Fl
 			ws := memdb.NewWatchSet()
 			it, err := s.selectOrAll(selector)
 			if err != nil {
-				s.logger.Error(fmt.Sprintf("error watching flags: %v", err))
+				s.logger.WithFields(zap.Field{Key: "selector", String: selector.ToLogString()}, zap.Field{Key: "error", Interface: err}).Error("error getting flags")
```
erka:
```suggestion
				s.logger.WithFields(zap.String("selector", selector.ToLogString()), zap.Error(err)).Error("error getting flags")
```

erka: And ideally it should be 
```go
s.logger.WithFields(zap.Object("selector", selector), zap.Error(err)).Error("error getting flags")
```

leakonvalinka: Went with the first one, the second required another method to be implemented

aepfli (approved):

PR #1853, opened 2026-01-27
docs(ADR): Proposal for the languages support policy

cupofcat:
This PR introduces a new Architecture Decision Record (ADR) establishing a Tiered Support Model for flagd language providers.

Motivation
As the flagd ecosystem grows, the maturity and feature parity of different language providers have varied. This inconsistency can confuse adopters about which SDKs are "production-ready" versus experimental.

This policy aims to:

- Set clear expectations for users regarding stability, feature parity, and SLAs.
- Define a clear path for community contributions to graduate from "Incubation" to "Core Supported."
- Formalize governance roles (Core Maintainers vs. Language Maintainers).
- Set a clear expectations for releases

Related issues:
https://github.com/open-feature/flagd/issues/1750

PR #1867, opened 2026-02-06
docs: ADR proposing "rollout" operator

toddbaert:
Proposes a "rollout operator". Facilitates linear, progressive rollouts as a fundamental feature flag use case: gradually shifting traffic from one variant to another over time.

See ADR for justification and full proposal.

See also:
- draft implementation **WITH DEMO** in flagd: https://github.com/open-feature/flagd/pull/1868
  - run demo with: `make build-flagd && ./demo-rollout.sh`
- draft schema changes: https://github.com/open-feature/flagd-schemas/pull/205

output

### DIFF
--- a/docs/architecture-decisions/rollout-operator.md
+++ b/docs/architecture-decisions/rollout-operator.md
@@ -1,13 +1,16 @@
 ---
-status: draft
+status: rejected
 author: @toddbaert
 created: 2026-02-06
-updated: 2026-02-06
+updated: 2026-03-13
 ---
 
 # Rollout Operator
 
-The rollout operator enables time-based progressive feature rollouts.
+**Status: Rejected**: After discussion, this proposal was rejected in favor of enhancing the existing `fractional` operator to accept JSONLogic expressions as weight arguments (see the [Alternative Proposal](#alternative-proposal-enhanced-fractional-with-dynamic-weights) section below and the [Fractional Operator ADR](fractional.md)).
+The rollout/rollback operators are largely syntactic sugar over `fractional` with dynamic weights, and the additional operator surface area across all language SDKs is not justified at this time.
+The enhanced `fractional` approach has been proven to support both progressive rollouts and FILO rollbacks using only existing primitives.
+A dedicated operator may be reconsidered in the future if strong user need emerges, perhaps "compiled" to the fractional alternative described.
 
 ## Background
 
--- a/docs/architecture-decisions/fractional.md
+++ b/docs/architecture-decisions/fractional.md
@@ -2,7 +2,7 @@
 status: draft
 author: @toddbaert
 created: 2025-06-06
-updated: 2025-06-06
+updated: 2026-03-13
 ---
 
 # Fractional Operator
@@ -111,6 +111,21 @@ Note that in this example, we've also specified a custom bucketing value.
 }
 ```
 
+**Dynamic weights** are also supported: the weight argument in each variant array can be a JSONLogic expression that evaluates to a numeric value, not only a hard-coded integer.
+This enables use cases such as time-based progressive rollouts, where the weight changes dynamically based on the evaluation context (e.g., `$flagd.timestamp`).
+Negative weight values (which can result from dynamic expressions) must be clamped to 0.
+
+```jsonc
+// Time-based progressive rollout using dynamic weights:
+// the "on" weight grows as time advances, "off" weight shrinks.
+{
+  "fractional": [
+    ["on",  { "-": [{ "var": "$flagd.timestamp" }, 1740000000] }],
+    ["off", { "-": [1800000000, { "var": "$flagd.timestamp" }] }]
+  ]
+}
+```
+
 ### Consequences
 
 - Good, because Murmur3 is fast, has good avalanche properties, and we don't need "cryptographic" randomness
@@ -118,4 +133,3 @@ Note that in this example, we've also specified a custom bucketing value.
 - Good, because our bucketing algorithm is relatively stable when new variants are added
 - Bad, because we only support string bucketing values
 - Bad, because we don't have bucket resolution finer than 1:99
-- Bad, because we don't support JSONLogic expressions within bucket definitions


docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+// longhand: explicit from and to - from "old" to "new"
+{"rollout": [1704067200, 1706745600, "old", "new"]}
+
+// with custom bucketBy
+{"rollout": [{"var": "email"}, 1704067200, 1706745600, "old", "new"]}
+```
+
+Parameters:
+
+- `bucketBy` (optional): JSONLogic expression for bucketing value; defaults to `flagKey + targetingKey`, consistent with existing `fractional`
```
beeme1mr, line 52: What's the expected behavior if `targetingKey` is omitted?

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Proposal
+
+### Operator Syntax
+
+Three forms are supported, following JSONLogic array conventions:
+
+```jsonc
+// shorthand: roll from defaultVariant to "new"
+{"rollout": [1704067200, 1706745600, "new"]}
```
beeme1mr, line 41: What's the expected behavior if `defaultVariant` has been omitted?

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+
+// with custom bucketBy
+{"rollout": [{"var": "email"}, 1704067200, 1706745600, "old", "new"]}
+```
+
+Parameters:
+
+- `bucketBy` (optional): JSONLogic expression for bucketing value; defaults to `flagKey + targetingKey`, consistent with existing `fractional`
+- `startTime`: Unix timestamp (seconds) when rollout begins (0% on `to`)
+- `endTime`: Unix timestamp (seconds) when rollout completes (100% on `to`)
```
beeme1mr: Can you define the start and end percent or weight?

beeme1mr: I think you may cover this in the example section but it's not 100% clear to me.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+{"rollout": [1704067200, 1706745600, "new"]}
+
+// longhand: explicit from and to - from "old" to "new"
+{"rollout": [1704067200, 1706745600, "old", "new"]}
+
+// with custom bucketBy
+{"rollout": [{"var": "email"}, 1704067200, 1706745600, "old", "new"]}
+```
+
+Parameters:
```
beeme1mr, line 50: Do you think it's worth defining the behavior if priorities are omitted or invalid?

tangenti:
While I understand the requirements behind this proposal, I believe this logic belongs in the flag management system by adjusting the variant weights in the fractional evaluation, rather than part of the targeting rules itself.

I think time-based feature flag is considered as an anti-pattern:

- It allows you to skip all the enforcements of rollout schedules. For example, when there's a production freeze for Black Friday, the time-based feature flag can still enable a feature because the flag configs were rolled out before the freeze.

- Rollback is hard. When the operator detects an issue after the feature flag is enabled, the launch percentage will keep increasing until the change is fully rolled back. This could take multiple hours.

-  The feature can be enabled in an unexpected time, and become a time bomb. Especially if the engineer who created the feature flag left the team or no longer maintaining the system.

AFAIK we disallow time-based flags internally. Bumping the variant weight is done by the rollout (either manual or automatic) that is enforced by rollout rules (for example, you cannot bump from 1% to 100% directly on production)

docs/architecture-decisions/rollout-operator.md
toddbaert: Same behavior as other targeting rules in the same situation (previously flag not found, but soon we will support code defaulting, see: https://flagd.dev/architecture-decisions/support-code-default/?h=default)

docs/architecture-decisions/rollout-operator.md
toddbaert: Same as fractional, it's an error situation. I will add that.

docs/architecture-decisions/rollout-operator.md
toddbaert: > Can you define the start and end percent or weight?

You can do this by using a fractional in the start/end. This is all composition based.

toddbaert:
> While I understand the requirements behind this proposal, I believe this logic belongs in the flag management system by adjusting the variant weights in the fractional evaluation, rather than part of the targeting rules itself.
> 
> I think time-based feature flag is considered as an anti-pattern:
> 
> * It allows you to skip all the enforcements of rollout schedules. For example, when there's a production freeze for Black Friday, the time-based feature flag can still enable a feature because the flag configs were rolled out before the freeze.
> * Rollback is hard. When the operator detects an issue after the feature flag is enabled, the launch percentage will keep increasing until the change is fully rolled back. This could take multiple hours.
> 
> -  The feature can be enabled in an unexpected time, and become a time bomb. Especially if the engineer who created the feature flag left the team or no longer maintaining the system.
> 
> AFAIK we disallow time-based flags internally. Bumping the variant weight is done by the rollout (either manual or automatic) that is enforced by rollout rules (for example, you cannot bump from 1% to 100% directly on production)

I think this is a matter of policy which can be enforced independently of flagd - it's quite an opinionated stance. We make use of time-sensitive flags already in flagd with the [timestamp](https://flagd.dev/reference/flag-definitions/?h=timest#flagd-properties-in-the-evaluation-context) features, and we leverage them internally at Dynatrace.

I will add this point to the `Bad` section though, it's a good thought.

docs/architecture-decisions/rollout-operator.md
toddbaert: Schema should generally prevent that, but I'll specify this to be consistent with other operators.

tangenti:
Also, @cupofcat has sent https://github.com/open-feature/flagd/pull/1853 for review. 

@toddbaert  Do you think this `rollout` operator will be a good candidate to practice and verify the proposed workflow for supporting a core feature for all languages?

docs/architecture-decisions/rollout-operator.md
beeme1mr: Could you please add an explicit example with a short description?

toddbaert: Hmmm I think there already is one [here on line 94](https://github.com/open-feature/flagd/pull/1867/changes#diff-cf04e90bd9ce39559ca9a51487a60f291992452e3e933d4eb9a0999df95e061dR94). Maybe I'm misunderstanding. Can you clarify?

juanparadox:
@toddbaert  If I'm understanding this right - this means that percentage rollouts can be automatically incremented via flagd without having to manually adjust the percentage weights (although they technically can still be manually adjusted outside of the use of this operator aka the way it works today) or implement some kind of automatic adjustment system outside of flagd?

If this is the case, this is awesome.

We have often wondered what an automatic incrementation system would look like and it always pointed to some kind of job hanging out somewhere updating the flag targeting. However, this new operator will essentially eliminate the need for something like that if my understanding is correct?

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+### Operator Syntax
+
+Three forms are supported, following JSONLogic array conventions:
+
+```jsonc
+// shorthand: roll from defaultVariant to "new"
+{"rollout": [1704067200, 1706745600, "new"]}
+
+// longhand: explicit from and to - from "old" to "new"
+{"rollout": [1704067200, 1706745600, "old", "new"]}
```
juanparadox, line 44: Question: This looks to be outside of the scope of this operator, but based on the `startTime` and `endTime`, would there be a way to set 'periods' or points in time whereby we can be explicit about what the percent should be?

For example, I have a time range from now until two days from now and I want it to start at `0` at the `startTime`, `50` tomorrow, and `100` by the `endTime`? I'm thinking of cases where you may want or need more of a predictable incrementation of the percentages over time.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+
+Variants can be JSONLogic expressions, enabling composition:
+
+```jsonc
+// Rollout to a fractional split
+{
+  "rollout": [
+    1704067200, 1706745600,
+    "old",
+    {"fractional": [["a", 50], ["b", 50]]}
```
juanparadox, line 118: Question: What is this actually happening here? I'm trying to understand what happens to the weights for `a` and `b` over time.

docs/architecture-decisions/rollout-operator.md
toddbaert: > and I want it to start at 0 at the startTime, 50 tomorrow, 

@juanparadox You will be happy to know this is already possible by using the injected `$flagd.timestamp`. This config does a "stepped" rollout from 100% "off", to 50/50, to 100% "on":

```yaml
"variantMigration": {
      "state": "ENABLED",
      "variants": {
        "off": false,
        "on": true
      },
      "defaultVariant": "off",
      "targeting": {
        "if": [
          {
          # before Feb 10, 100% off
            "<": [
              { "var": ["$flagd.timestamp"] },
              1739164800
            ]
          },
          "off",
         # Feb 10-17: 50/50 split
          {
            "<": [
              { "var": ["$flagd.timestamp"] },
              1739251200
            ]
          },
          {
            "fractional": [
              ["off", 50],
              ["on", 50]
            ]
          },
          # after Feb 17, 100% on
          "on"
        ]
      }
    }
```

The limitation we have is you can only do explicit steps like this without the "rollout" operator I've proposed, which is gradual.

docs/architecture-decisions/rollout-operator.md
toddbaert: > Question: What is this actually happening here? I'm trying to understand what happens to the weights for `a` and `b` over time.

This configuration is gradually migrating from 100% "old" to 50/50 split of some other variants "a" and "b". At the start, 100% of the users are seeing "old". In the middle, 50% are seeing "old" and 50% are seeing a random spllt of "a" and "b". At the end, all users are either "a" or "b".

Basically, we are progressing linearly from what's on like 97 to waht's on like 98 over the given timeframe.

juanparadox: Thank you for the explanation! So the `to` in this case( which is the fractional) is the middle point in time based on the `startTime` and `endTime`? Can I force it to end all users to be `a` if I wanted to with an iteration of this configuration?

toddbaert: No, `to` is always the end, the destination. `to` in this case just happens to be fractional (it can be any configuration), so we are slowly migrating everyone *to* this final 50/50 split. Half-way through, half of our users will still be getting old, and half with be at the split a/b. To put it another way:

start: 100% old               
middle: 50% old, 25% a, 25% b      
end: 50% a, 50% b (we've fully reached the `to`)                       

Migrating _to_ a split is probably not extremely common, but the point is that the destination rule can be whatever you want.

juanparadox: Thank you, makes sense!

toddbaert:
> Also, @cupofcat has sent #1853 for review.
> 
> @toddbaert Do you think this `rollout` operator will be a good candidate to practice and verify the proposed workflow for supporting a core feature for all languages?

Yes, potentially. It could be a good opportunity to pilot that.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+- **Consistent hashing**: must use the same hashing strategy as `fractional` (MurmurHash3-32)
+- **Cross-language portable**: must use only integer arithmetic (no floating-point operations)
+- **JSONLogic conventions**: must follow established patterns for custom operators
+
+## Proposal
+
+### Operator Syntax
+
+Three forms are supported, following JSONLogic array conventions:
+
```
tangenti, line 38: Have you considered any name that is less generic than `rollout`?

docs/architecture-decisions/rollout-operator.md
tangenti, line 41: How do you plan to define the schema for this operator?

JSON schema seems to have a human-readable format for date and time: https://json-schema.org/understanding-json-schema/reference/type#dates-and-times, but not for timestamps.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,121 @@
...
+- **Time-based**: traffic distribution must change automatically based on the current timestamp
+- **Deterministic**: same user must get consistent results at a given point in time (no re-bucketing mid-request)
+- **Composable**: must support nested JSONLogic (e.g., rollout to a `fractional` distribution)
+- **Consistent hashing**: must use the same hashing strategy as `fractional` (MurmurHash3-32)
+- **Cross-language portable**: must use only integer arithmetic (no floating-point operations)
+- **JSONLogic conventions**: must follow established patterns for custom operators
+
+## Proposal
+
+### Operator Syntax
```
tangenti, line 35: With the change to allow non-string randomization unit for `fractional` and a more granular weight, I suppose it will be feasible to express the semantic with `fractional`?

docs/architecture-decisions/rollout-operator.md
toddbaert: Yes, good point. I feel "just OK" about the name itself. Other ideas I had:

- "linear" (implies the linear transition, but does't imply "time")
- "transition" (implies change over time, but probably not as "flag-specific" as rollout)

I'm very open to other names. Any suggestions?

docs/architecture-decisions/rollout-operator.md
toddbaert: I linked the proposed schema in the description - we could enhance it to put lower/upper bounds on this int. At the moment, these are just integers, consistent with the [$flagd.timestamp](https://flagd.dev/architecture-decisions/flag-configuration/?h=timest#system-provided-context) we inject into every evaluation context. This works great for rules generated by a UI, especially - we use a standard datepicker control in our internal implementation's UI to create these, so the UX is nice. It's certainly NOT nice to manually write these though.

@jonathannorris also is keen on supporting a more friendly date format, like you are suggesting, I think. I'm in favor of supporting a more user-friendly date format, but I think since we already use epochs, I would do that in a separate proposal, inject this new format and perhaps some operators specific to it, then update this to also accept that format.

If you feel like we should really do that in this proposal as well, I'm OK to widen the discussion, but I think it would be better to discuss the date topic in a separate proposal that tackles all the related features (injecting it in evals, short forms, timezones, related operators, etc).

docs/architecture-decisions/rollout-operator.md
toddbaert: Sorry @tangenti , can you rephrase? I'm worried I'm not understanding this one.

My intent is for the hash implemented here to be the same as what's used in the `fractional` (including the recently merged changes for non-strings, when that's been implemented).

toddbaert:
> > While I understand the requirements behind this proposal, I believe this logic belongs in the flag management system by adjusting the variant weights in the fractional evaluation, rather than part of the targeting rules itself.
> > I think time-based feature flag is considered as an anti-pattern:
> > 
> > * It allows you to skip all the enforcements of rollout schedules. For example, when there's a production freeze for Black Friday, the time-based feature flag can still enable a feature because the flag configs were rolled out before the freeze.
> > * Rollback is hard. When the operator detects an issue after the feature flag is enabled, the launch percentage will keep increasing until the change is fully rolled back. This could take multiple hours.
> > 
> > -  The feature can be enabled in an unexpected time, and become a time bomb. Especially if the engineer who created the feature flag left the team or no longer maintaining the system.
> > AFAIK we disallow time-based flags internally. Bumping the variant weight is done by the rollout (either manual or automatic) that is enforced by rollout rules (for example, you cannot bump from 1% to 100% directly on production)
> 
> I think this is a matter of policy which can be enforced independently of flagd - it's quite an opinionated stance. We make use of time-sensitive flags already in flagd with the [timestamp](https://flagd.dev/reference/flag-definitions/?h=timest#flagd-properties-in-the-evaluation-context) features, and we leverage them internally at Dynatrace.
> 
> I will add this point to the `Bad` section though, it's a good thought.

@tangenti I added a "Bad" point about this. We could also add additional documentation, and perhaps consider additional features to improve the observability of "baked-in" time-sensitive changes like this in general.

docs/architecture-decisions/rollout-operator.md
toddbaert: I'm now more in favor of `"rollout"` due to the "rollback" sister operator.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,158 @@
...
+- `endTime`: Unix timestamp (seconds) when rollout completes (100% on `to`)
+- `from`: Starting variant or expression (omit for shorthand to use `defaultVariant`)
+- `to`: Target variant or expression
+
+### Hashing Consistency
+
+The rollout operator uses the same hashing strategy as `fractional`:
+
+- MurmurHash3 (32-bit)
+- Same default bucketing value: `flagKey + targetingKey`
```
jonathannorris, line 65: @toddbaert we had discussed potentially prefixing the rollout hash to start with say `rollout-`/`rollback-` so that the distribution of users between a nested `rollout`/`rollback` and `fractional` operator don't exactly match. 

For example: 
- You have 10% of users who have linearly "rolled-out" so far, and there is a 50/50% fractional applied to those passing the `rollout` operator.
  - If the bucketing value is not different between the `rollout`/`rollback` and the `fractional` I believe those 10% of users would all be bucketed into the first variation of the 50/50% fractional.

docs/architecture-decisions/rollout-operator.md
beeme1mr: Do we want it to error or be falsy?

docs/architecture-decisions/rollout-operator.md
toddbaert: I was thinking about this. It definitely solves the problem of injecting some additional entropy between `fractional` and `rollout`, which as you point out is important. Exactly how we do it has implications. I see 2 options:

1. append `"rollout"` or some other salt, immediately before hashing, to whatever bucket value we retrieve
- good: this will solve the entropy problem for everyone
- good: it "just works" even for custom bucketing values
- bad: if anyone for some reason _wanted_ to have the same bucketing between fractional/rollout they can't (ie: they want the behavior we identify as problematic)

2. append rollout _just_ in the default bucketing, ie: `"rollout" + flagKey + targetingKey`
- good: solves the problem for default use-case
- good: gives users full control (they can remove this "salt" if they want to
- bad: means custom bucketing values need to manually include some "salt"

WDYT?

One other thing to note - is that both `rollout` and `rollback` would need to use the same salt to ensure `rollback` works as expected, or the hash inversion won't work.

jonathannorris: Hum... yea that's a challenging default behaviour to sort out. I think the correct application is to always apply a salt/prefix of some sort so that `rollout`/`rollback` distributions are always different than `fractional` distributions. This would obviously need to be made clear in the docs.

I personally can't think of a situation where you would want them to match, but I can imagine plenty of issues when running experiments and analyzing the data where having the bucketing distribution matching being a fairly large issue. So I think we should try and make the defaults here as "safe" as possible.

And good point on the same salt for `rollout` / `rollback`.

beeme1mr: I'd probably just add a salt prefix.

docs/architecture-decisions/rollout-operator.md
justinabrahms: Agreed that I like "rollout". It makes me want "linear" as a subtype. I can easily imagine that we'd want non-linear rollouts. (e.g. slow at start, then rapidly increasing as we gain confidence).. so thinking through how that might evolve (or if that's explicitly out of scope for flagd) feels meaningful.

docs/architecture-decisions/rollout-operator.md
toddbaert: Ya - I think I agree. I think somebody really _wanting_ the same bucketing is so unrealistic it's not worth supporting. I've made updates about this with this [commit](https://github.com/open-feature/flagd/pull/1867/changes/101ccccdb000b1d83b5cd69931dd2e2280575381).

jonathannorris (approved):

docs/architecture-decisions/rollout-operator.md
toddbaert: I added a "future proofing" section:

https://github.com/open-feature/flagd/pull/1867/changes#diff-cf04e90bd9ce39559ca9a51487a60f291992452e3e933d4eb9a0999df95e061dR148-R161

docs/architecture-decisions/rollout-operator.md
toddbaert: I think this is a bigger issue we need to address with @jonathannorris 's recent findings. I will open a new dicussion about that.

docs/architecture-decisions/rollout-operator.md
toddbaert: Same as https://github.com/open-feature/flagd/pull/1867#discussion_r2788096287

docs/architecture-decisions/rollout-operator.md
tangenti: I'm thinking of `linear-ramp-up` or simply `ramp-up`

docs/architecture-decisions/rollout-operator.md
tangenti: I tried this targeting rule on the playground and it seems to work well - I received `on` or `off` depending on the targeting key.

```
{
  "flags": {
    "my-progressive-feature": {
      "state": "ENABLED",
      "variants": {
        "on": true,
        "off": false,
        "error": false
      },
      "defaultVariant": "error",
      "targeting": {
        "fractional": [
          { "var": "targetingKey" },
          [
            "on",
            { "-": [{ "var": "$flagd.timestamp" }, 1740000000] }
          ],
          [
            "off",
            { "-": [1800000000, { "var": "$flagd.timestamp" }] }
          ]
        ]
      }
    }
  }
}
```

docs/architecture-decisions/rollout-operator.md
tangenti: Thanks. 
I worry less about the readability and more about the validation on the value. Having a lower/upper bound sounds a good start to me.

tangenti:
I'm so sorry for the delay of review.... Got sick last week and been busy with oncall this week.

cupofcat:
Hi @toddbaert, so sorry for the delay. Last 2 weeks were a bit busy for us. Here's my take:

I agree with @tangenti that this is a time-bomb. We would probably discourage our customers from using that but I understand other organizations have different policies and approaches so let me focus on what I think we should change.

# Naming

Personally, `rollout` and `rollback` don't sit well with me for a couple of reasons:
- it's a targeting operator that essentially is a function of time; a rollout is more of an action that happens and then it's done; it just does not fit for me semantically.
- `rollback` specifically seems redundant and also, potentially, misleading (it sounds like it should be instant, not gradual)

My proposal would be for `timed` or `time-weighted`, with additional `reversed` option. Or perhaps something like `auto-split` or `time-split`. Or even `progressive`.

Even if we want to stick with `rollout` I would lose `rollback` and go with `reversed` or something like that.

# Array-based / positional arguments

I think that is a bit brittle and potentially confusing. Also, future extensions, like curve parameter (also positional, also before the timestamps), compounds the ambiguity. How about we go with the object?
```
 {"time-weighted": {
    "bucketBy": {"var": "email"},
    "start": 1704067200,
    "end": 1706745600,
    "from": "old",
    "to": "new",
    "order": "reverses"
  }}
```
Some of these can be skipped as in the ADR.

I know JSONLogic conventions heavily favor arrays, but maybe we could use an object as the single argument in the array?

# Safety

1. Can we add additional warnings in documentation for some examples of situations when using this is unsafe? (e.g. the ones that @tangenti mentioned)

2. Time-based flags might be a bit annoying during an incident, or further compound the issues, because the output changes without a configuration deployment. Perhaps we should also add some sort of `disabled` option that instantly always returns the `from` variant? That said, someone still can put not the current / correct variant in `from`.

3. Confusion between unix `seconds` vs `millisecond` - if a user accidentally inputs a millisecond timestamp (1704067200000), the rollout will be scheduled for the year 56,000 AD :D The JSON Schema could enforce reasonable integer boundaries like a maximum of 3000000000 (the year 2065).

4. Better Observability (Reason Codes)
How about, when a user gets a variant from a `time-weighted`, the operator must append specific metadata to the evaluation response (e.g., reason: TIMED_PROGRESSION", metadata: {"percentage": 42.5}). If a system fails, incident responders need to see in their telemetry exactly why a user got a variant and at what progression state the flag was at the moment of failure.

5. Salting
The ADR states that to inject entropy, the UTF-8 bytes of "rollout" are appended to the bucketing value. If the default bucketing (flagKey + targetingKey) is used, this is safe. However, if a user specifies a custom bucketBy (e.g., `{"var": "email"}`), the hashed string becomes `bob@example.comrollout`, no? If a developer creates two different rollout flags using that same custom `bucketBy`, their hashes will perfectly correlate. The exact same users will get both features early. Although I guess, the same is the case for existing `fractional`

# Freezes

I think together with this ADR we should consider introducing some mechanism for code freezes that `time-weighted` respects. Perhaps `$flagd.freeze_start` and `$flagd.freeze_end`? Or, at least, document this and file and issue to design a freeze feature.

# Testing strategy

Can we explicitly specify things like:
- Implementations must accept a timestamp override (e.g., via evaluation context $flagd.timestamp) so tests can pin time
- Cross-implementation conformance: a test table with specific hash values, timestamps, and expected outcomes that all language implementations must pass (perhaps we can do this in gherkin?)

# Missing in the ADR

What happens when a rollout config is changed mid-rollout?

Scenario: A rollout is at 40%. An operator edits the flag config, changes the time window, swaps variants, or adds a `bucketBy`. What happens to users who were already bucketed? Can a config change mid-rollout can cause user experience discontinuities?

# Overflow and out of bands behavior

1. As Gemini said, if we end up in negative elapsed time (e.g. because we are still before `startTime`, or `endTime` is before `startTime`, etc) we risk overflow in some languages.

2. `startTime` == `endTime` case: duration is 0, bucket is 0, elapsed is 0. 0 < 0 is false. Is that the right semantics for an "instant" rollout?

# The rollback description is incorrect, no?

It says `rollback is identical to rollout except it applies bitwise NOT (~) to the hash before bucketing` but the shown example is different, I think? If the formula were truly identical-except-for-hash-inversion, then as elapsed grows, more users would transition to `to`.

To produce the described behavior (Alice "reverts to old at t=60%", Fred "already reverted immediately"), the rollback formula must also swap the return values, no?

```
// rollout
  if bucket < elapsed { return to }
    return from

// rollback - NOT just "same but with ~hash"
  bucket := (uint64(^hashValue) * uint64(duration)) >> 32
  if bucket < elapsed {
      return from  // swapped
  }
  return to  // swapped
```

At elapsed=0 everyone is on `to` (the feature stays), and as time progresses, users gradually move back to `from`, with last-adopters reverting first.

docs/architecture-decisions/rollout-operator.md
toddbaert: I'm ok with `ramp_up`, as then the sister inverse operator could easily be `ramp_down` as well.

docs/architecture-decisions/rollout-operator.md
toddbaert: That's very interesting (by the way, I've already started working on the associated improvements [here](https://github.com/open-feature/flagd/pull/1877))

@tangenti Unless I'm missing something, the enhanced fractional can do everything the `rollout` I've proposed can do. I wonder if it's worth implementing the rollout just for "UX" since it's easier to read than this. WDYT?

tangenti: For UX I agree. Do you see any good way of having this as a shared evaluator or something similar? Essentially expanding the operator into the equivalent `fractional` expression, which requires less efforts to impl it all languages and will grant us a better consistency on the behavior.

toddbaert:
Thanks for the thorough review @cupofcat! I'll try to respond to each:

## Naming

I don't mind much on the name - happy to take a vote on that as long as it's clear to users.

## Array-based / positional arguments

Every [JSONLogic](https://jsonlogic.com/operations.html) operator uses the form `"op": [...]` — this is a fundamental convention we can't break. Our existing operators (`fractional`, `sem_ver`, `starts_with`, etc.) all use positional arrays. Introducing an object here would be inconsistent with both JSONLogic itself and all existing custom flagd operators.

## Safety

1. **Warnings in documentation** — Agreed, will add explicit examples of unsafe scenarios (production freezes, etc.).
2. **`disabled` option** — This is already achievable: just replace the `rollout` rule with a static variant (or disable the flag). I don't think we need a dedicated option within the operator for this, unless I'm not understanding you.
3. **Seconds vs milliseconds confusion** — Good catch. The JSON Schema can enforce reasonable bounds (e.g., `maximum: 3000000000` / year ~2065) to catch this. I'll add this to the schema PR.
4. **Better observability (reason codes)** — This is an interesting idea worth exploring. It would be a broader change (not specific to this operator), but I could see value in richer evaluation metadata. I'll file an issue to discuss this separately. It might also apply to fractional (ie: indicate this assignment was pseudorandom)
5. **Salting / hash correlation with custom `bucketBy`**; This is addressed: the default bucketing includes `flagKey + targetingKey`, which decorrelates hashes across different flags. If a user opts into a custom `bucketBy`, they accept this tradeoff — same as with `fractional` today. Users can always include the flag key in their custom expression if they want decorrelation (we inject it with `$flagd.flagKey`.

## Freezes

Interesting idea, but I think this belongs in a separate proposal. Freezes are a broader operational concern that would apply to all time-sensitive features (including existing `$flagd.timestamp` rules), not just `rollout`.

## Testing strategy

Great point. I'll add a section to the ADR specifying:
- I like the idea of somehow override the timestamp, but I would somehow like to make sure this actually doesn't happen in prod - so far we have guards around it - is running flagd in a testing mode a good option here? WDYT?
- A cross-implementation conformance test table with specific hash values, timestamps, and expected outcomes should be defined (Gherkin in the test-harness is a good fit for this) it would be a requirement.

## Missing in the ADR — config changes mid-rollout

The operator is stateless — every evaluation is a pure function of `(bucketBy, startTime, endTime, currentTime)`. If you change the config mid-rollout:
- Changing the time window shifts the progression curve
- Changing variants immediately affects newly-evaluated users
- Changing `bucketBy` re-shuffles all users

There's no stored state to become inconsistent. I'll add a note clarifying this stateless property.

## Overflow and out-of-band behavior

1. **Negative elapsed / `endTime < startTime`**: The implementation validates `endTime > startTime` at parse time and clamps elapsed to `[0, duration]`. I'll make the clamping explicit in the ADR.
2. **`startTime == endTime`**: This is rejected by the `endTime > startTime` validation. An "instant rollout" should just be a direct variant assignment. I'll document this.

## "The rollback description is incorrect, no?"

You're right — good catch. The original text stated rollback is "identical to rollout except it applies bitwise NOT (~) to the hash," but that alone isn't sufficient. Inverting the hash only reverses the *user ordering* while still transitioning from `from` → `to`. A true rollback (transitioning from `to` back to `from`) also requires swapping the return values. I've updated the ADR to reflect this — the implementation section now shows both the hash inversion and the swap.

---

All that being said... please see: https://github.com/open-feature/flagd/pull/1867#issuecomment-4011771755

toddbaert:
@cupofcat @tangenti @beeme1mr @jonathannorris @justinabrahms 

~Honestly, the more I think about it, the more I think I want to close this proposal in favor of just recommending what @tangenti proposed [here](https://github.com/open-feature/flagd/pull/1867#discussion_r2858145362).~

~I'm not convinced we even need a "UX" wrapper for rollout - I think it's already complex enough that people are not likely to manually write it anyway - we will be generating such rollouts from UI controls, for instance. I think we can just make one small change to the flagd+schema to support dynamic values for the JSON weight argument in fractional, and call it "done".~

I don't think this solves everything we need, unfortunately. See content [here](https://github.com/open-feature/flagd/pull/1867/changes#r2906059160).

docs/architecture-decisions/rollout-operator.md
toddbaert: I'm not even sure such a UX enhancement is needed. I really love the elegance of your proposal.

cupofcat:
> @cupofcat @tangenti @beeme1mr @jonathannorris @justinabrahms
> 
> Honestly, the more I think about it, the more I think I want to close this proposal in favor of just recommending what @tangenti proposed [here](https://github.com/open-feature/flagd/pull/1867#discussion_r2858145362).
> 
> I'm not convinced we even need a "UX" wrapper for rollout - I think it's already complex enough that people are not likely to manually write it anyway - we will be generating such rollouts from UI controls, for instance. I think we can just make one small change to the flagd+schema to support dynamic values for the JSON weight argument in fractional, and call it "done".

Generally, sounds good. Can you explain a bit more what you mean exactly by dynamic values for JSON weight argument in fractional though?

jonathannorris:
> @cupofcat @tangenti @beeme1mr @jonathannorris @justinabrahms
> 
> Honestly, the more I think about it, the more I think I want to close this proposal in favor of just recommending what @tangenti proposed [here](https://github.com/open-feature/flagd/pull/1867#discussion_r2858145362).
> 
> I'm not convinced we even need a "UX" wrapper for rollout - I think it's already complex enough that people are not likely to manually write it anyway - we will be generating such rollouts from UI controls, for instance. I think we can just make one small change to the flagd+schema to support dynamic values for the JSON weight argument in fractional, and call it "done".

yea reluctantly I think I agree, I'm not a big fan of the developer experience of not having a `rollout` operator (there are developers hand-writing flagd configs, but I agree that they really shouldn't be for these complex rules). Maybe we can define an easier to write flagd intermediary language that would compile to flagd configs if there is a push for a more human-writable flagd configurations later. But that doesn't seem like a good reason to add a new operator here if we can achieve the same behaviour with the existing operators.

toddbaert:
> > @cupofcat @tangenti @beeme1mr @jonathannorris @justinabrahms
> > Honestly, the more I think about it, the more I think I want to close this proposal in favor of just recommending what @tangenti proposed [here](https://github.com/open-feature/flagd/pull/1867#discussion_r2858145362).
> > I'm not convinced we even need a "UX" wrapper for rollout - I think it's already complex enough that people are not likely to manually write it anyway - we will be generating such rollouts from UI controls, for instance. I think we can just make one small change to the flagd+schema to support dynamic values for the JSON weight argument in fractional, and call it "done".
> 
> Generally, sounds good. Can you explain a bit more what you mean exactly by dynamic values for JSON weight argument in fractional though?

~@cupofcat I just mean that the second argument for each split value is not dynamic - we proposed enhancements to make the variant dynamic, such that any JSONLogic was valid for the variant argument, but right now the numeric weight argument is necessarily an integer, and not an evaluated expression. This would be required to implement what @tangenti outlined, but it's easily done.~

See [here](https://github.com/open-feature/flagd/pull/1867#issuecomment-4023391910).

toddbaert:
After thinking about this more, I think @tangenti's approach [here](https://github.com/open-feature/flagd/pull/1867#discussion_r2858145362) can approximate _nearly_ everything in the `rollout` proposal, given:

1. PR #1877 (non-string/nesting fractional variants) - already in progress
2. High-precision fractional bucketing ADR - proposed
3. One additional change: allow the `fractional` weight argument to be a JSONLogic expression instead of only a hard-coded integer

However, I think two things are missing compared to the dedicated `rollout`/`rollback` operators:

**FILO rollback is not achievable.** The rollback operator inverts the hash (~hash) so users revert in reverse adoption order (first-in-last-out). With pure fractional, swapping the weight expressions gives you a gradual rollback, but it's FIFO; early adopters revert first, not last. Worse, if you swap weights mid-rollout, the formula starts at 100% "new" and ramps down, meaning users who never saw the new variant are suddenly exposed to it before being reverted; exactly the wrong behavior during an incident. There's no way to express bitwise NOT in JSONLogic, so this can't be composed with the existing `fractional`. This may or may not matter, because of course you can just instantly revert all users to a some variant if the rollout isn't going well, but it's a real behavioral difference.

**Hash decorrelation requires manual effort.** The `rollout` ADR automatically appends `"rollout"` to the bucketing value before hashing, so users' position in the rollout timeline doesn't correlate with their bucket in a nested `fractional`. With the pure-fractional approach, an outer and inner `fractional` use the same hash by default, meaning early-rollout users systematically land in the first inner bucket. Users *can* work around this by manually adding a salt via `cat`, but that's non-obvious.

If we're OK with these trade-offs (or willing to address them separately; e.g., document the salting pattern, accept FIFO rollback), I'm on board with closing this ADR in favor of the enhanced `fractional` approach... otherwise I think we should merge this. Personally I really like the elegance of what @tangenti proposed, but I think these tradeoffs are meaningful, so I'm slightly leaning to merging the proposal as is, but I'm interested in what others think.

cc @beeme1mr @jonathannorris @tangenti @cupofcat 

:warning: UPDATE:

After speaking with a few people about this (@jonathannorris and @beeme1mr , as well as others) it seems like a "FILO rollback" as described in this documentation is critical and as far as I can see can't be approximated with @tangenti 's suggestion (elegant though it is), so I'm inclined to merge this proposal as is. I've added this is a considered alternative though, [here](https://github.com/open-feature/flagd/pull/1867/changes#r2906059160).

:warning: UPDATE 2:

@cupofcat astutely pointed out that there were some oversights in my "inverted hash" solution to accomplish the rollback; I cherry-picked some "lucky" values that I incorrectly generalized. We need an additional parameter to achieve the rollback functionality we want, and an inverted hash doesn't help. This makes @tangenti 's proposal practical again, since the pivot point can be represented easily by arithmetic in the weight argument. I think the question is now entirely whether we want a dedicated operation for this or not, it seems like purely UX. See the commit [here](https://github.com/open-feature/flagd/pull/1867/changes/f6846268974a6195b6d84a7d3019c728b64584e1) for all the changes. Thanks @cupofcat !

docs/architecture-decisions/rollout-operator.md
toddbaert: See content [here](https://github.com/open-feature/flagd/pull/1867/changes#r2906059160). I added this as a considered alternative, but I don't think it quite covers all the use-cases we want.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,209 @@
...
+{"rollout": [{"var": "email"}, "linear|exponential", 1704067200, 1706745600, "old", "new"]}
+```
+
+```jsonc
+{"rollout": [{"var": "email"}, { some-json-logic-lambda }, 1704067200, 1706745600, "old", "new"]}
+```
+
+**Implementation of non-linear rollouts is out of the scope of this proposal.**
+
+### Considered Alternative: Enhanced `fractional` with Dynamic Weights
```
toddbaert: Considered alternative section based on @tangenti 's idea.

docs/architecture-decisions/rollout-operator.md
toddbaert: Other options:

- `shift/unshift`
- `ramp/deramp`
- `transition/untransition` (I like this one but it might but it have some political baggage)
- `promote/demote`
- `progress/regress`

cupofcat:
> ⚠️ UPDATE 2:
> 
> @cupofcat astutely pointed out that there were some oversights in my "inverted hash" solution to accomplish the rollback; I cherry-picked some "lucky" values that I incorrectly generalized. We need an additional parameter to achieve the rollback functionality we want, and an inverted hash doesn't help. This makes @tangenti 's proposal practical again, since the pivot point can be represented easily by arithmetic in the weight argument. I think the question is now entirely whether we want a dedicated operation for this or not, it seems like purely UX. See the commit [here](https://github.com/open-feature/flagd/pull/1867/changes/f6846268974a6195b6d84a7d3019c728b64584e1) for all the changes. Thanks @cupofcat !

If anyone's curious, here is my math:
* $T_S$: Time start
* $T_E$: Time end
* $e$: Elapsed time
* $d$: Duration
* $t$: Current time

#### 1. Timeline and Baseline Variables
$T_S = 340 \quad | \quad e_0 = (0)$
$t = 520 \quad | \quad e = (180)$
$T_E = 1120 \quad | \quad e = (780)$
$d = (1120 - 340) = 780$

#### 2. User Bucketing Examples ($t = 520$)
Assuming a 32-bit hash space ($2^{32}$), we map users and the current elapsed time:
* $\text{hash}(U_1) = \frac{160}{780} \cdot 2^{32} \approx 0.205 \cdot 2^{32}$
* $\text{hash}(U_2) = \frac{190}{780} \cdot 2^{32} \approx 0.243 \cdot 2^{32}$
* Target boundary at $t=520$: $\frac{180}{780} \cdot 2^{32} \approx 0.231 \cdot 2^{32}$

#### 3. Progression Scenarios ($e=180$)
**Rollout ($R\uparrow$):**
* $U_1 : 0.205 \cdot 2^{32} < 0.231 \cdot 2^{32} \implies \text{"to"}$
* $U_2 : 0.243 \not< 0.231 \implies \text{"from"}$

**Rollback ($R\downarrow$):** *(Testing the inverted boundary at 1 - 0.231 = 0.769)*
* $U_1 : 0.205 \cdot 2^{32} \ngtr 0.769 \cdot 2^{32} \implies \text{"to"}$
* $U_2 : 0.243 \ngtr 0.769 \implies \text{"to"} \ (!)$

#### 4. General Formula Derivation

**Rollout ($e\uparrow$):**
$$\frac{h \cdot d}{2^{32}} < e \implies \text{"to"}$$
$$h \cdot d < e \cdot 2^{32}$$
$$h < \frac{e}{d} \cdot 2^{32}$$

**Rollback ($R\downarrow$):**
$$\frac{\sim h \cdot d}{2^{32}} < e \implies \text{"from"}$$
$$\sim h \cdot d < e \cdot 2^{32}$$
$$(2^{32} - 1 - h) \cdot d < e \cdot 2^{32}$$
$$(2^{32} - 1 - h) < \frac{e}{d} \cdot 2^{32}$$
$$2^{32} - 1 - \frac{e}{d} \cdot 2^{32} < h$$
$$2^{32}\left(1 - \frac{e}{d}\right) - 1 < h$$
$$2^{32}\left(\frac{d-e}{d}\right) - 1 < h$$
$$h > \frac{d-e}{d} \cdot 2^{32} \quad (-1)$$

docs/architecture-decisions/rollout-operator.md
beeme1mr: I like `ramp_up/ramp_down` or `rollout/rollback`. I don't find the other terms very intuitive.

docs/architecture-decisions/rollout-operator.md
```diff
@@ -0,0 +1,308 @@
...
+{"rollout": [{"var": "email"}, "linear|exponential", 1704067200, 1706745600, "old", "new"]}
+```
+
+```jsonc
+{"rollout": [{"var": "email"}, { some-json-logic-lambda }, 1704067200, 1706745600, "old", "new"]}
+```
+
+**Implementation of non-linear rollouts is out of the scope of this proposal.**
+
+### Recommended Alternative: Enhanced `fractional` with Dynamic Weights
```
toddbaert:
```suggestion
### Alternative Proposal: Enhanced `fractional` with Dynamic Weights
```

cupofcat:
> ⚠️ UPDATE 2:
> 
> @cupofcat astutely pointed out that there were some oversights in my "inverted hash" solution to accomplish the rollback; I cherry-picked some "lucky" values that I incorrectly generalized. We need an additional parameter to achieve the rollback functionality we want, and an inverted hash doesn't help. This makes @tangenti 's proposal practical again, since the pivot point can be represented easily by arithmetic in the weight argument. I think the question is now entirely whether we want a dedicated operation for this or not, it seems like purely UX. See the commit [here](https://github.com/open-feature/flagd/pull/1867/changes/f6846268974a6195b6d84a7d3019c728b64584e1) for all the changes. Thanks @cupofcat !

My take on this is that documenting how to use fractional for this use case will strengthen fractional overall and avoid adding a new operator for all the langauges with potential for bugs, etc. Rollout/rollback proposal is mostly syntactic sugar on top of fractional. Perhaps the right approach is to add a generator, similar how OF CLI generates flag accessors? We could look into extending OF CLI to have vendor specific behaviors (e.g. `of rollout --flagd`) or perhaps have a path for that in flagd binary (`flagd generate rollout`), or some other tool. Management planes could also just generate these configs from the UI, etc. Improvements to fractional in the future might make the rollouts use case better too.

I would punt on custom operator for now until we see strong user need. WDYT?

beeme1mr:
> > ⚠️ UPDATE 2:
> > @cupofcat astutely pointed out that there were some oversights in my "inverted hash" solution to accomplish the rollback; I cherry-picked some "lucky" values that I incorrectly generalized. We need an additional parameter to achieve the rollback functionality we want, and an inverted hash doesn't help. This makes @tangenti 's proposal practical again, since the pivot point can be represented easily by arithmetic in the weight argument. I think the question is now entirely whether we want a dedicated operation for this or not, it seems like purely UX. See the commit [here](https://github.com/open-feature/flagd/pull/1867/changes/f6846268974a6195b6d84a7d3019c728b64584e1) for all the changes. Thanks @cupofcat !
> 
> My take on this is that documenting how to use fractional for this use case will strengthen fractional overall and avoid adding a new operator for all the langauges with potential for bugs, etc. Rollout/rollback proposal is mostly syntactic sugar on top of fractional. Perhaps the right approach is to add a generator, similar how OF CLI generates flag accessors? We could look into extending OF CLI to have vendor specific behaviors (e.g. `of rollout --flagd`) or perhaps have a path for that in flagd binary (`flagd generate rollout`), or some other tool. Management planes could also just generate these configs from the UI, etc. Improvements to fractional in the future might make the rollouts use case better too.
> 
> I would punt on custom operator for now until we see strong user need. WDYT?

I'm good with this now that we've proven that both rollouts and rollbacks work. We may want to add syntactic sugar at some point but I don't think it's necessary right now.

toddbaert:
> > > ⚠️ UPDATE 2:
> > > @cupofcat astutely pointed out that there were some oversights in my "inverted hash" solution to accomplish the rollback; I cherry-picked some "lucky" values that I incorrectly generalized. We need an additional parameter to achieve the rollback functionality we want, and an inverted hash doesn't help. This makes @tangenti 's proposal practical again, since the pivot point can be represented easily by arithmetic in the weight argument. I think the question is now entirely whether we want a dedicated operation for this or not, it seems like purely UX. See the commit [here](https://github.com/open-feature/flagd/pull/1867/changes/f6846268974a6195b6d84a7d3019c728b64584e1) for all the changes. Thanks @cupofcat !
> > 
> > 
> > My take on this is that documenting how to use fractional for this use case will strengthen fractional overall and avoid adding a new operator for all the langauges with potential for bugs, etc. Rollout/rollback proposal is mostly syntactic sugar on top of fractional. Perhaps the right approach is to add a generator, similar how OF CLI generates flag accessors? We could look into extending OF CLI to have vendor specific behaviors (e.g. `of rollout --flagd`) or perhaps have a path for that in flagd binary (`flagd generate rollout`), or some other tool. Management planes could also just generate these configs from the UI, etc. Improvements to fractional in the future might make the rollouts use case better too.
> > I would punt on custom operator for now until we see strong user need. WDYT?
> 
> I'm good with this now that we've proven that both rollouts and rollbacks work. We may want to add syntactic sugar at some point but I don't think it's necessary right now.

@cupofcat @tangenti @beeme1mr 

This [commit](https://github.com/open-feature/flagd/pull/1867/changes/965d7d38f8855da00d5cd36ef9c47db82f73d944) rejects the proposal and makes the small related amendment to fractional. TBH, I'm quite happy with this outcome. We learned _a lot_ here and I think this is a "less-is-more" situation. Spending a time learning you don't need something is time well-spent.

PR #1885, opened 2026-03-02
docs: add json/yaml switch to playground to fix schema validation, make playground fullscreen

toddbaert:
Adds a JSON/YAML switcher to the playground. This:

- brings back JSON validation in JSON mode (doesn't work in YAML editor)
- adds a button that converts from YAML/JSON
- starts in the mode indicated by the query param


playground-app/src/App.tsx
```diff
@@ -412,7 +454,7 @@ function App() {
                   theme={editorTheme}
                   width="100%"
                   height="80px"
-                  language="yaml"
+                  language="json"
```
toddbaert, line 457: Let's just keep the evaluation context always YAML.

playground-app/src/App.tsx
```diff
@@ -266,14 +298,13 @@ function App() {
   return (
     <div
       style={{
-        maxWidth: "825px",
+        maxWidth: "100%",
```
toddbaert, line 301: Let's remove the side nav when we are in the playground for more code-space.

PR #1919, opened 2026-04-01
docs: ADR for updated disabled flag behaviour

### LINKED ISSUE #1918: [FEATURE] Treat disabled flag evaluation as success with reason=DISABLED
### Requirements

`flagd` currently treats disabled flags as errors (`reason=ERROR`, `errorCode=FLAG_DISABLED`). This should instead resolve successfully with `reason=DISABLED`, returning the flag's defaultVariant value.

Expected behavior:
- Disabled flags resolve successfully across gRPC, OFREP, and in-process evaluation
- Missing flags still return `FLAG_NOT_FOUND`
- `ResolveAll` includes disabled flags instead of omitting them
- Telemetry records disabled evaluations as non-errors
- This would align flagd with the OpenFeature resolution reason model and avoid conflating intentionally disabled flags with missing flags.

suthar26:
## This PR

- adds ADR for disabled flag behavior updates

### Related Issues

Fixes #1918


docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
...
+- **Option 2: Return a successful evaluation with** `reason=DEFAULT` — Treat disabled flags as if they had no targeting, collapsing the disabled state into the default reason
+- **Option 3: Status quo** — Keep the current error behavior and document it as intentional divergence from the OpenFeature spec
+
+## Proposal
+
+We propose **Option 1: Successful evaluation with `reason=DISABLED` returning `defaultVariant`**.
+
+When a flag exists in the store but has `state: DISABLED`, the evaluator should return a successful evaluation with the following properties:
+
+- **value**: The flag's `defaultVariant` value (from the flag configuration)
```
aepfli: [q] should we specifically mention the code fallback here too, as the default variant from the configuration might be non existent?

aepfli (approved):

docs/architecture-decisions/disabled-flag-evaluation.md
suthar26: good catch, I added the code fallback here

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: Parth Suthar (@suthar26)
+created: 2026-04-01
+---
+# Treat Disabled Flag Evaluation as Successful with Reason DISABLED
+
+This ADR proposes changing flagd's handling of disabled flags from returning an error (`reason=ERROR`, `errorCode=FLAG_DISABLED`) to returning a successful evaluation with `reason=DISABLED` and the flag's `defaultVariant` value. A flag that does not exist in a flag set should remain a `FLAG_NOT_FOUND` error.
```
toddbaert: I'm not sure about this.

I would expect the behavior to be return the code-default in all cases when the flags is disabled. Otherwise I'm not really sure what the point is. If we use the `defaultVariant`, we are still delegating the evaluation do the flag management system, which contradicts the `DISABLED` idea to me.

I think we just just immediately return the code default.

Am I not thinking about this correctly? Maybe you had a different concept in mind.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
+author: Parth Suthar (@suthar26)
+created: 2026-04-01
```
toddbaert, line 5: Minor: the ADR template and other accepted ADRs include an `updated` field.

```suggestion
author: Parth Suthar (@suthar26)
created: 2026-04-01
updated: 2026-04-01
```

toddbaert (commented):
Reviewed focusing on in-process provider coverage, OFREP spec alignment details, and format.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
...
+3. **Conflation of "missing" and "disabled"**: gRPC v1 maps both `FLAG_NOT_FOUND` and `FLAG_DISABLED` to `connect.CodeNotFound`. These are semantically different situations: a missing flag is a configuration or deployment error, while a disabled flag is an intentional operational decision (incident remediation, environment-specific rollout, not-yet-ready feature).
+4. **Loss of observability**: When disabled flags are treated as errors, they pollute error metrics and alerting. Operators who disable a flag for legitimate reasons (ongoing incident remediation, feature not ready for an environment) see false error signals. Conversely, if they suppress these errors, they lose visibility into flag state entirely. A successful evaluation with `reason=DISABLED` would give operators a clean signal without noise.
+5. **Flag set use cases**: In multi-flag-set deployments, a flag may exist in a shared definition but be disabled in certain flag sets (e.g., disabled for `staging` but enabled for `production`). Treating this as an error forces the application into error-handling paths when the flag is simply not active — a normal operational state, not an exceptional one.
+
+Related context:
+
```
toddbaert, line 28: This is a good point but could be improved, I think. `FLAG_DISABLED` is not in the [OpenFeature spec's error code list](https://openfeature.dev/specification/types/#error-code) (`PROVIDER_NOT_READY`, `FLAG_NOT_FOUND`, `PARSE_ERROR`, `TYPE_MISMATCH`, `TARGETING_KEY_MISSING`, `INVALID_CONTEXT`, `PROVIDER_FATAL`, `GENERAL`); flagd invented it. Similarly, the [OFREP `evaluationFailure` schema](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml) only allows `errorCode` values of `PARSE_ERROR`, `TARGETING_KEY_MISSING`, `INVALID_CONTEXT`, `GENERAL`; there is no `FLAG_DISABLED`. So flagd's current behavior is arguably a spec violation on both the OpenFeature and OFREP sides. This deserves a mention in the Background since it strengthens the entire motivation..

```suggestion
3. **Conflation of "missing" and "disabled"**: gRPC v1 maps both `FLAG_NOT_FOUND` and `FLAG_DISABLED` to `connect.CodeNotFound`. These are semantically different situations: a missing flag is a configuration or deployment error, while a disabled flag is an intentional operational decision (incident remediation, environment-specific rollout, not-yet-ready feature).
4. **Non-standard error code**: `FLAG_DISABLED` is not in the [OpenFeature specification's error code list](https://openfeature.dev/specification/types/#error-code), nor is it a valid `errorCode` in the [OFREP `evaluationFailure` schema](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml) (which only allows `PARSE_ERROR`, `TARGETING_KEY_MISSING`, `INVALID_CONTEXT`, `GENERAL`). Conversely, `DISABLED` *is* already a valid `reason` in OFREP's [`evaluationSuccess` schema](https://github.com/open-feature/protocol/blob/main/service/openapi.yaml). flagd's current treatment of disabled flags as errors is a spec violation on both the OpenFeature and OFREP sides.
```

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
...
+- Should the `FlagDisabledErrorCode` constant be retained (but unused) for a deprecation period, or removed immediately?
+- How should in-process providers handle the transition? They evaluate locally and would need to be updated to return `DISABLED` reason instead of throwing an error.
+
+## More Information
+
+- [OpenFeature Specification - Resolution Reasons](https://openfeature.dev/specification/types/#resolution-reason)
+- [OpenFeature Specification - Error Codes](https://openfeature.dev/specification/types/#error-code) — notably, `FLAG_DISABLED` is not in the spec's error code list
+- [flagd Flag Definitions Reference](https://flagd.dev/reference/flag-definitions/)
+- [ADR: Support Explicit Code Default Values](./support-code-default.md)
+- [flagd Testbed](https://github.com/open-feature/flagd-testbed)
```
toddbaert, line 270: The in-process provider story is listed as an open question (line 280), but it is one of the three active evaluation surfaces (alongside gRPC v2 and OFREP; v1 is deprecated). It should be addressed concretely in the proposal body and timeline rather than left open.

The sync.proto payload (`flag_configuration` JSON) already includes disabled flags with `"state": "DISABLED"`; the wire format does not need to change. What *does* need to change is each language SDK's in-process evaluator: instead of raising an error when it encounters a disabled flag, it must return `reason=DISABLED` with code-default deferral (omitted value/variant). This is a coordinated rollout across all SDK in-process providers and should be called out I think so we don't miss it in implementation.

Consider adding a section like:

```markdown
### In-process providers

The sync.proto payload already includes disabled flags in the `flag_configuration` JSON. No wire-format changes are needed. Each language SDK's in-process evaluator must be updated to return `reason=DISABLED` with code-default deferral (omitted value/variant) instead of raising an error when encountering a disabled flag. This is functionally the same change as the flagd core evaluator, replicated in each SDK.
```

And the open question on line 280 can then be removed.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,284 @@
...
+  "metadata": {
+    "flagSetId": "my-app",
+    "scope": "production"
+  }
+}
+```
+
+**OFREP response** (single flag evaluation):
+
+```json
```
toddbaert, line 192: These OFREP example responses should clarify the HTTP status code. Currently, a disabled flag in OFREP single-flag evaluation returns HTTP 404 (because flagd rewrites `FLAG_DISABLED` to `FLAG_NOT_FOUND`). After this change, it would return HTTP 200 since it is a successful evaluation. This is a breaking change for HTTP clients that branch on status codes and should be called out explicitly, e.g.:

```markdown
**OFREP response** (single flag evaluation, HTTP 200 -- previously HTTP 404):
```

toddbaert (changes requested):
I think [this](https://github.com/open-feature/flagd/pull/1919#discussion_r3065771031) is a bit of a blocker for me; it seems a bit incoherent. Maybe I'm missing something though or not thinking about it correctly. I think if this was changed this would be an easy approval for me. Please help me understand if I'm off-base :sweat_smile:

docs/architecture-decisions/disabled-flag-evaluation.md
aepfli: i agree, it should use the code_default/fallback - the barriers in flagd are blurry, because if there is no default_variant on the flag, we also use the code_default. but in the case of a disabled flag, it should be the code_default

suthar26: no you are totally right, thats what we should be doing

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,270 @@
...
+## Proposal
+
+We propose **Option 1: Successful evaluation with `reason=DISABLED` deferring to code defaults**.
+
+When a flag exists in the store but has `state: DISABLED`, the evaluator should return a successful evaluation that always defers to the application's code-defined default value. A disabled flag means the flag management system is explicitly stepping aside — it has nothing to say about what the value should be. This is semantically different from returning a configured `defaultVariant`, which would still delegate the decision to the flag management system and contradict the meaning of `DISABLED`.
+
+The response has the following properties:
+
+- **value**: Omitted (signals code-default deferral). The SDK/provider uses the application's code-defined default value.
+- **variant**: Omitted (empty string).
```
beeme1mr: Why an empty string? I'd expect this to be completely omitted in supported languages.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,270 @@
...
+2. Update `resolve[T]` in `core/pkg/evaluator/json.go` to handle `DisabledReason` with empty variants (avoids `TYPE_MISMATCH`)
+3. Remove the disabled-flag skip in `ResolveAllValues` and update the `default` branch of the type switch to include disabled flags for gRPC v1 requests
+4. Update `SuccessResponseFrom` in `core/pkg/service/ofrep/models.go` to preserve `reason=DISABLED` (with field omission) for disabled flags deferring to code defaults
+5. Remove `FlagDisabledErrorCode` handling from `errFormat`, `errFormatV2`, and `EvaluationErrorResponseFrom`
+6. Update gRPC v1 service layer (`flag_evaluator_v1.go`) to handle nil values in `ResolveAll` responses for disabled flags deferring to code defaults
+7. Update each language SDK's in-process provider evaluator to return `reason=DISABLED` with code-default deferral instead of raising an error when encountering a disabled flag
+8. Update flagd-testbed with test cases for disabled flag evaluation across all surfaces
+9. Update OpenFeature providers to recognize `DISABLED` as a non-error, non-cacheable reason
+10. Update provider documentation and migration guides
```
beeme1mr: This isn't really a timeline.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Open questions
+
+- Should bulk evaluation include an option to exclude disabled flags for clients that prefer the current behavior (smaller payloads)?
+- How should existing dashboards and alerts that key on `FLAG_DISABLED` errors be migrated? Should we provide a deprecation period where both behaviors are available?
```
beeme1mr: I don't think this is something we can do. We just need to make it clear that's in the release notes that we're making the behavior change.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,270 @@
...
+- Does this change require a new flagd major version, or can it be introduced in a minor version with appropriate documentation given the spec alignment argument?
```
beeme1mr: It should be marked as a behavior-breaking change, which will update the minor version since we're sub-1.0.

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,270 @@
...
+- Should the `FlagDisabledErrorCode` constant be retained (but unused) for a deprecation period, or removed immediately?
```
beeme1mr: How would we retain this?

suthar26: ideally keeping it in the code base but on other hand we should remove it

beeme1mr (approved):

toddbaert (approved):
All my concerns, especially [the one referenced here](https://github.com/open-feature/flagd/pull/1919#pullrequestreview-4091423368), have been addressed. I think this is a good idea, and recent improvements in OFREP and flagd make it more feasible than it once was.

I think this adds value to flagd as a solution, so I support it.

I would love consideration from more non-Dynatracers, particularly @cupofcat , @juanparadox , @askpt or others.

Maybe also @oxddr

jonathannorris (approved):

askpt (approved):

juanparadox (approved):

oxddr:
I'd be happy to have a look, but I won't be able to do so earlier than next week.

oxddr (approved):

docs/architecture-decisions/disabled-flag-evaluation.md
```diff
@@ -0,0 +1,106 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: draft
```
beeme1mr:
```suggestion
status: accepted
```

jonathannorris:
Noting for the record: this change has a downstream impact on Multi-Provider implementations. Currently, `FLAG_DISABLED` errors can trigger fall-through to the next provider. Once disabled flags return a successful evaluation with `reason=DISABLED`, that fall-through signal is gone — Multi-Provider implementations that rely on error-based fall-through will need to be updated.

The multi-provider spec will need to define a new fall-through mechanism to account for this. Tracking as a required follow-up.

Issue #1913, opened 2026-03-27
[BUG] null values in targeting rules for object-valued flags when there is no defaultValue should evaluate to the code default


asppsa:
### Observed behavior

I think I have found a bug.  A minimal demonstration of the issue is the following:

```json
{
  "$schema": "https://flagd.dev/schema/v0/flags.json",
  "flags": {
    "my-object-flag": {
      "state": "ENABLED",
      "variants": {
        "on": { "payload": "hello" }
      },
      "targeting": {
        "if": [false, "on", null]
      }
    }
  }
}
```

If I evaluate the above with `getObjectDetails` (using OpenFeature in Node.js), I would expect get back whatever code default I supplied.  Instead I get:

```js
{
  value: undefined,
  reason: '',
  variant: undefined,
  flagMetadata: {},
  flagKey: 'my-object-flag'
}

```

### Expected Behavior

Based on https://flagd.dev/reference/flag-definitions/#targeting-rules,

>  If a null value is returned by the targeting rule, the defaultVariant is used. This can be useful for conditionally "exiting" targeting rules and falling back to the default (in this case the returned reason will be DEFAULT).... If defaultVariant is not defined or is null, and no variant is resolved from targeting, flagd providers will revert to the code default.

In the example, the targeting rule trivially returns null, and then because there is no `defaultValue`, I believe the correct behaviour should be to return the code default.

### Steps to reproduce

1. run `flagd start --uri file:./test.json` (with my test file above)
2. evaluate the following node.js code:

```js
import { FlagdProvider } from '@openfeature/flagd-provider'
import { OpenFeature } from '@openfeature/server-sdk'

await OpenFeature.setProviderAndWait(
  new FlagdProvider({
    host: '127.0.0.1',
    port: 8013,
    tls: false,
  }),
)

const client = OpenFeature.getClient()

console.dir(
  await client.getObjectDetails('my-object-flag', {}),
  { depth: null },
)
```

asppsa:
GitHub Copilot helped me work the above out.  It further claims that:

>  the bug has two layers, and only the first one is object-specific:
> 1. `objectResponseV2` is missing `SetReasonOnly` entirely — so the `responseV2[T]` type assertion silently fails and the response is completely empty (no Reason, no Value, nothing). This only affects object evaluation.
[...]

PR #1925, opened 2026-04-02
fix: object flags without `defaultVaraint` dont default in RPC

toddbaert:
`objectResponseV2` was missing the `SetReasonOnly` method, unlike all other V2 response types.

I added an assertion to prevent regressions.

 thanks @asppsa for the detailed report!
 
 Fixes: https://github.com/open-feature/flagd/issues/1913


suthar26:
@toddbaert would be nice to add a testcase for this to ensure something like this is not missed in another version bump

toddbaert:
@suthar26 added tests.

toddbaert:
See https://github.com/open-feature/flagd/issues/1913#issuecomment-4183925189 as well.

PR #1939, opened 2026-04-11
feat: add strict validation option to reject invalid flag configs

### LINKED ISSUE #1487: [BUG] flagd accepts invalid flag config with just a warning
### Observed behavior

When using a flag config that does not conform to the schema, flagd shows the following warning in the logs but accepts the config nevertheless without error:
`2024-12-19T10:26:53.980+0100	warn	evaluator/json.go:114	flag definition does not conform to the schema; validation errors:  1:flags.is-enabled: Must validate one and only one schema (oneOf)  2:flags.is-enabled.variants.off: Invalid type. Expected: number, given: boolean  3:flags.is-enabled: Must validate all the schemas (allOf) 
`

This leads to inconsistent behavior during evaluations (https://github.com/open-feature/flagd/issues/1481)

### Expected Behavior

There are a couple of behaviors that could make sense:
- when flag config validation fails, return an error and don't accept the updated config, essentially keep working with whatever was already in the flag store (preferred)
- possibly, accept only parts of the config that conform to the schema and return a warning / error (probably not a good idea)

### Steps to reproduce

Use the following flag config. The variants have mismatched types (number and boolean).

```
{
    "flags": {
      "is-enabled": {
        "defaultVariant": "off",
        "state": "ENABLED",
        "targeting": {
          "if": [
            {
              "<": [
                {
                  "%": [
                    {
                      "var": "request_id"
                    },
                    1000
                  ]
                },
                100
              ]
            },
            "on",
            "off"
          ]
        },
        "variants": {
          "on": 1,
          "off": false
        }
      }
    }
  }
```

buildingisfun23:
## Summary
- Adds `WithStrictValidation()` option to the JSON evaluator
- When enabled, flag configurations that fail schema validation are rejected with an error instead of accepted with a warning
- The flag store preserves its previous state when validation fails in strict mode
- Backward compatible: default behavior (accept with warning) is unchanged
- Includes 4 new tests covering strict/non-strict modes and state preservation

Fixes #1487


toddbaert:
@buildingisfun23 - I'm not opposed - I wonder if we should connect this directly to a flagd startup arg as well? WDYT?

buildingisfun23:
Good call @toddbaert! I've added `--strict-validation` as a flagd startup argument (also available as `FLAGD_STRICT_VALIDATION` env var via viper). It wires through the existing `WithStrictValidation()` evaluator option so operators can enable strict schema validation at startup without any code changes.

toddbaert:
Hey @buildingisfun23 - thanks again. I've been thinking about this and I think it made sense to make one more change...

I made--strict-validation fail-fast: invalid initial configs exit the process, and readiness now requires every configured source to have produced a valid config at least once. Help text got a cascading-failure warning so SREs know what they're opting into.

This is opt in, and it matches similar behavior in things like envoy/nginx. See [changes](https://github.com/open-feature/flagd/pull/1939/changes/f03059e8d056103da7321b87913ae1677b77a3f3).

WDYT? Does this make sense? Should we divide it into 2 options? I kept thinking it would be weird to have no flags if an initial sync fails, and it led be down this path.

toddbaert:
@buildingisfun23 Are you still interested in this?

PR #1940, opened 2026-04-14
fix(store): flag deduping for experimental incremental gRPC sync

erenatas:
## This PR
Adds a membership table to the in-memory store that tracks which flagSetIds reference which flags, enabling flag deduplication during incremental gRPC sync updates. Flags shared across flagSetIds are stored once per (key, source) in the flags table, while the membership table records (flagSetId, key, source) triples for per-flagSetId query resolution.  

Flags shared across flagSetIds are stored once per (key, source) in the flags table, while a membership table records (flagSetId, key, source) triples for query resolution. `Get`, `GetAll`, and `Watch` resolve flags through membership when a flagSetId selector doesn't match directly, iterating all entries to select the highest-priority flag. Content changes are upserted in place so updated targeting rules or variants are not dropped, and flagSetId is always  populated in metadata even when the original is nil. Watch registers scoped per-(key, source) channels instead of the entire flags table. 

All membership code paths are gated behind an `atomic.Bool` set only on the first incremental update, so non-incremental callers never enter them.

## Why

When many projects share most of the same flags, storing one copy per `flagSetId` wastes memory and can lead to OOMs. Deduplicating shared flags keeps the incremental sync path scalable without changing full-snapshot behavior.

### DIFF
--- a/core/pkg/store/query.go
+++ b/core/pkg/store/query.go
@@ -24,6 +24,10 @@ const flagSetIdSourceCompoundIndex = flagSetIdIndex + "+" + sourceIndex
 const keySourceCompoundIndex = keyIndex + "+" + sourceIndex
 const flagSetIdKeySourceCompoundIndex = flagSetIdIndex + "+" + keyIndex + "+" + sourceIndex
 
+// membership table for incremental update deduplication
+const membershipTable = "membership"
+const membershipFlagSetIdKeyIndex = flagSetIdIndex + "+" + keyIndex
+
 // flagSetId defaults to a UUID generated at startup to make our queries consistent
 // any flag without a "flagSetId" is assigned this one; it's never exposed externally
 var nilFlagSetId = uuid.New().String()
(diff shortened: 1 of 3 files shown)


toddbaert:
I understand the motivation; 8GB to 150MB is dramatic. But this PR fundamentally changes the store architecture by introducing a parallel state system with manual synchronization, read-time patching, and dual code paths through every query method. That's a lot of new surface area for bugs in the critical evaluation path, all for an experimental feature :sweat_smile: .

Maybe the 810x duplication of 938 common flags can be fixed on the sync server end?. Can we rethink this approach?

erenatas:
> I understand the motivation; 8GB to 150MB is dramatic. But this PR fundamentally changes the store architecture by introducing a parallel state system with manual synchronization, read-time patching, and dual code paths through every query method. That's a lot of new surface area for bugs in the critical evaluation path, all for an experimental feature 😅 .
> 
> Maybe the 810x duplication of 938 common flags can be fixed on the sync server end?. Can we rethink this approach?

Can you have a look now? I tried addressing your concerns. Non-incremental callers won't touch membership code and all experimental code is isolated behind the introduced gate

toddbaert:
Marking as draft for now after some conversations with @erenatas and @beeme1mr

PR #1949, opened 2026-04-16
fix: handle missing/null targeting keys in fractional evaluator

### LINKED ISSUE #1948: [BUG] missing targeting key handling differs across SDK implementations
## Parent Issue

Sub-issue of https://github.com/open-feature/flagd/issues/1904

## Summary

When fractional evaluation is used without an explicit bucketing property argument, bucketing should depend on the `targetingKey`. If the `targetingKey` is missing (or `null`) in the evaluation context, no bucket should be selected and the evaluator should return `null` so that the flag’s default variant (or code default, if no default variant exists) is used.

### Java

- targetingKey = null: explicit check, returns null
- targetingKey = "": treated as any other string (bucketing by flagkey+"")
- targetingKey missing entirely: is set to null -> same as targeting = null case

### Python

- targetingKey = nil: explicit check, returns null
- targetingKey = "": same as null
- targetingKey missing entirely: same as null

### JS

- targetingKey = null: explicit check, returns null
- targetingKey = "": same as null
- targetingKey missing entirely: same as null

### Go/flagd core

- targetingKey = nil: explicit check, returns null
- targetingKey = "": treated as a normal string (bucketing by flagkey+"")
- targetingKey missing entirely: same as null

### dotnet

- targetingKey = null: throws NullRef exception in `FlagdProperties.cs`
- targetingKey = "": treated as a normal string (bucketing by flagkey+"")
- targetingKey missing entirely: bucketing using flagKey only (`flagdProperties.FlagKey + flagdProperties.TargetingKey` -> treated as flagkey+"")

## EDIT: Suggested Fix

In all 3 cases (targeting key is null/nil, missing, or ""), we want to return null and fall back to the default variant.

utafrali:
Fixes #1948

The fractional evaluator wasn't handling missing or null targeting keys the same way other SDKs did. Added nil checks to return the default variant when the key is missing or null. Also added tests for the three cases: missing key, null key, and empty string.

### DIFF
--- a/core/pkg/evaluator/fractional.go
+++ b/core/pkg/evaluator/fractional.go
@@ -90,6 +90,10 @@ if len(valuesArray) < 1 {
 			return "", nil, fmt.Errorf("flag %q: bucketing value not supplied and no targetingKey in context", flagKey)
 		}
 
+		if targetingKey == "" {
+			return "", nil, nil
+		}
+
 		bucketBy = fmt.Sprintf("%s%s", properties.FlagKey, targetingKey)
 	}
 
--- a/core/pkg/evaluator/fractional.go
+++ b/core/pkg/evaluator/fractional.go
@@ -48,6 +48,10 @@ func (fe *Fractional) Evaluate(values, data any) any {
 		return nil
 	}
 
+	if feDistributions == nil {
+		return nil
+	}
+
 	hashValue := uint32(murmur3.StringSum32(valueToDistribute))
 	return distributeValue(hashValue, feDistributions)
 }
@@ -78,6 +82,9 @@ if len(valuesArray) < 1 {
 			valuesArray = valuesArray[1:]
 		}
 
+		if dataMap[targetingKeyKey] == nil {
+			return "", nil, nil
+		}
 		targetingKey, ok := dataMap[targetingKeyKey].(string)
 		if !ok {
 			return "", nil, fmt.Errorf("flag %q: bucketing value not supplied and no targetingKey in context", flagKey)
--- a/core/pkg/evaluator/fractional_test.go
+++ b/core/pkg/evaluator/fractional_test.go
@@ -442,6 +442,67 @@ func TestFractionalEvaluation(t *testing.T) {
 			expectedValue:   blueHex,
 			expectedReason:  model.TargetingMatchReason,
 		},
+		"null targetingKey returns default variant": {
+			flags: []model.Flag{{
+				Key:            "headerColor",
+				State:          "ENABLED",
+				DefaultVariant: redVariant,
+				Variants:       colorVariants,
+				Targeting: []byte(`{
+							"fractional": [
+								["blue", 50],
+								["green", 50]
+							]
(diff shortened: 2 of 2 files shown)


toddbaert:
Hey @utafrali  - can you do a `git commit --amend --signoff`? This is a legal requirement for CNCF projects. See the failing DCO action: https://github.com/open-feature/flagd/pull/1949/checks?check_run_id=71604066833

core/pkg/evaluator/fractional_test.go
```diff
@@ -442,6 +442,67 @@ func TestFractionalEvaluation(t *testing.T) {
...
+				Targeting: []byte(`{
+							"fractional": [
+								["blue", 50],
+								["green", 50]
+							]
+						}`),
+			}},
+			flagKey: "headerColor",
+			context: map[string]any{
+				"targetingKey": "",
```
leakonvalinka, line 500: Hi, this wasn't defined in the issue yet when you opened your PR, but the suggested fix is that `targetingKey = ""` should be treated the same as `targetingkey = nil` and a non-existent targeting key and therefore result in falling back to the default. :)

toddbaert: +1, see the testbed test for this:
https://github.com/open-feature/flagd-testbed/blob/484ada4dd979f663214c024f3cb601141328a1b5/gherkin/targeting.feature#L376-L381

utafrali:
Done, pushed the fix.

core/pkg/evaluator/fractional.go
```diff
@@ -48,6 +48,10 @@ func (fe *Fractional) Evaluate(values, data any) any {
 		return nil
 	}
 
+	if feDistributions == nil {
```
leakonvalinka, line 51: Can this ever be the case? 
Edit: Oh yes, I see, with your changes below.

core/pkg/evaluator/fractional.go
```diff
@@ -78,11 +82,18 @@ if len(valuesArray) < 1 {
 			valuesArray = valuesArray[1:]
 		}
 
+		if dataMap[targetingKeyKey] == nil {
+			return "", nil, nil
+		}
 		targetingKey, ok := dataMap[targetingKeyKey].(string)
 		if !ok {
```
leakonvalinka, line 89: Just a suggestion: You could also simply add a check here for an empty string.

leakonvalinka (approved):
I left a feedback comment, but just as a suggestion so not blocking of course!

utafrali:
Done, pushed the fix.

toddbaert:
> Hey @utafrali - can you do a `git commit --amend --signoff`? This is a legal requirement for CNCF projects. See the failing DCO action: https://github.com/open-feature/flagd/pull/1949/checks?check_run_id=71604066833

This still need to be done for both commits. @utafrali

utafrali:
Done, pushed the fix.

beeme1mr (approved):

PR #1950, opened 2026-04-17
fix: various custom operator conformance fixes

### LINKED ISSUE #1942: Update flagd-testbed to v3.6.0 and fix evaluation edge cases
## Summary

Update the flagd-testbed dependency to [v3.6.0](https://github.com/open-feature/flagd-testbed/releases/tag/v3.6.0) and fix the Go in-process evaluator to pass all new edge-case tests.

Testbed v3.6.0 adds coverage for the edge cases identified in:
- #1873 (sem_ver parsing inconsistencies)
- #1874 (custom operator error return values)
- #1875 ($evaluators/$ref resolution gaps)

### sem_ver (#1873)

1. **sem_ver returns `false` on parse failure instead of `null`**; should return `nil` to trigger default variant fallback, not take the false branch
2. **sem_ver returns `false` on unknown operator instead of `null`**; same issue as above

### Custom operator error returns (#1874)

3. **`ends_with` returns `false` on error while `starts_with` returns `nil`**; both should return `nil` consistently
4. **`fractional` returns `""` (empty string) on no-bucket-match**; should return `nil` to trigger default variant fallback instead of causing a variant lookup failure

### $evaluators/$ref resolution (#1875)

5. **Non-deterministic `$evaluators` replacement ordering** due to Go `map` iteration randomization; should use sorted or deterministic iteration
6. **No diagnostic feedback on unresolved `$ref`s at parse time**; misconfiguration only discovered at evaluation time with confusing error about unknown `$ref` operator

Parent: #1904

leakonvalinka:
* support V prefix in semver
* return null on errors in semver, ends_with, and fractional (already existed on starts_with)
* fix fractional fallback when total weight is zero

Fixes: https://github.com/open-feature/flagd/issues/1942

---

Related:
* https://github.com/open-feature/java-sdk-contrib/pull/1778
* https://github.com/open-feature/js-sdk-contrib/pull/1519
* https://github.com/open-feature/dotnet-sdk-contrib/pull/635
* https://github.com/open-feature/python-sdk-contrib/pull/386
* https://github.com/open-feature/go-sdk-contrib/pull/874


toddbaert (approved):
@leakonvalinka Thanks!

Looks great. I made some small tweaks for consistency with other implementations with this [commit](https://github.com/open-feature/flagd/pull/1950/changes/cc4f80eb4e8568858f80e67845f59586e6d85fba):

- used a JSON "round trip" to make the $ref replace much easier (because whitespace will be predictable)
- renamed a func (`parseSemanticVersion` to `normalizeVersion`)

I also updated the go-flagd-provider in use, which means we could update some of the excludes [here](https://github.com/open-feature/flagd/pull/1950/changes#diff-ca62391b26e30e01d1f8cee0eac7829ba8e9b5e22d80eab0e83881086fb4177bR208) (for instance, we were still using `fractional-v1`). I also added a comment here explaining why a lot of the exclusions are permanent and justified.

aepfli (approved):

askpt (approved):

PR #1952, opened 2026-04-20
docs: fix code block language for custom operation results

kamina-zzz:
## Summary
- Fix code block language from `shell` to `json` for JSON result examples in custom operation docs
- Affects fractional operation and string-comparison operation result examples

## Test plan
- [x] Verify rendered output shows JSON syntax highlighting instead of shell

(This supersedes #1951, which only covered the fractional operation fix.)


toddbaert (approved):
Thanks @kamina-zzz

PR #1960, opened 2026-05-01
feat: add custom headers support for sync providers

raflFaisal:
## Summary
Add support for custom headers in HTTP and gRPC sync providers

## Problem
flagd's sync providers cannot inject custom request headers into outbound requests. This limits integration with environments where intermediaries (API gateways, reverse proxies, service meshes) require specific headers for routing, tenant identification, or policy enforcement, resulting in errors like 421 Misdirected Request or misrouting.

## Solution
`flagd start --sources='[{"uri":"https://proxy-gateway.xyz.com/flags/sync","provider":"http","headers":{"X-Proxy-Gateway-Host":"backend-api.service"}}]' `

Headers are sent as HTTP request headers for HTTP sync and as gRPC metadata for gRPC sync.

## Examples:

_gRPC with headers_:
```
flagd start \
  --sources='[{"uri":"proxy-gateway-server.xyz.com","provider":"grpc","selector":"features","headers":{"X-Routing-Target-Host":"b.flags-server.service"}}]'
```

_Using environment variables_:
```
export FLAGD_SOURCES='[{"uri":"https://proxy-gateway.bk-eu-west6.service-mesh.dqs.booking.com/v2/flags/sync?project=flags","provider":"http", "headers":{"X-Proxy-Gateway-Host":"b-flags-api.service"}}]' 
./bin/flagd start  --debug 
```

## Scope & compatibility

Fully backward compatible, headers are optional and default to empty
No behavior change for existing configurations
Covers both HTTP and gRPC sync providers

## Test plan
Unit tests for HTTP sync header injection
Unit tests for gRPC sync metadata injection


toddbaert:
IMO it'd be simpler to drop the global `--sync-headers` flag and keep headers per-source only, like other source-level params (`authHeader`, `certPath`, etc.).

The global + per-source merging adds complexity:
- case-insensitive precedence logic with `EqualFold`
- env var parsing fallback since viper can't handle `StringToString` from env
- implicit override behavior that users need to reason about
- `map[string]string` is natural in the per-source JSON config but awkward as a CLI flag (custom parsing, comma-separated key=value strings)

If someone has multiple sources needing the same headers, duplicating them in the source config is straightforward and explicit. WDYT?

core/pkg/sync/http/http_sync_test.go
```diff
@@ -320,6 +320,83 @@ func TestHTTPSync_Fetch(t *testing.T) {
 	}
 }
 
+func TestNewHTTP_PassesHeaders(t *testing.T) {
+	headers := map[string]string{"X-Custom": "value"}
+	config := sync.SourceConfig{
```
toddbaert, line 325: Nit: `"X-Custom"` is already in canonical form, so this test would pass even without the `http.CanonicalHeaderKey` canonicalization code in `NewHTTP`. Using a non-canonical key like `"x-custom"` and asserting it becomes `"X-Custom"` would actually verify the behavior.

toddbaert (changes requested):
I'm in favor of this, but unless there's strong reason to, I'd prefer not to duplicate this as both a source option and a global option: https://github.com/open-feature/flagd/pull/1960#issuecomment-4373598366

docs/reference/sync-configuration.md
```diff
@@ -147,6 +148,26 @@ sources:
...
+### Custom Headers
+
+Custom headers can be injected into all outbound sync requests (HTTP and gRPC) using the
+`--sync-headers` flag:
+
+```sh
+# CLI flag
+flagd start --sync-headers="X-Proxy-Gateway-Host=myhost,X-Tenant-ID=tenant1" \
+  --sources='[{"uri":"http://my-flags.com/flags","provider":"http"}]'
+
```
toddbaert:
```suggestion
```
```

I think the closing ``` is missing?

raflFaisal:
Thanks @toddbaert for the review.
Makes sense, agreed. I've dropped the global --sync-headers flag and kept headers per-source only. This removes the precedence logic and keeps the solution simple.

toddbaert (approved):

beeme1mr (approved):

Issue #186, opened 2022-10-04
[FEATURE] Prometheus metric endpoint


AlexsJones:
### Requirements

In order to capture data we should add a prometheus metrics endpoint.

Initial proposal:
```
flagd_total_request_count
flagd_request_bucket
flagd_request_per_minute
flagd_resolve_boolean_value_count
flagd_resolve_string_value_count
flagd_resolve_int_value_count
flagd_resolve_object_value_count
````

AlexsJones:
> I like this idea.
> 
> Somewhat related: I'd also like to add OpenTelemetry spans. Do you think it's worth creating a new issue for that? It would allow use to get timing data for various aspects of evaluation (parsing, evaluation, etc). I can create one if you think it's a good idea

I would say a new issue, it's a different kind of implementation.
Thing about spans are you need to send them somewhere, otherwise they are just firing into nothing.

beeme1mr:
It may be worth considering OTel metrics. It looks like the metric support is still alpha though.

https://github.com/open-telemetry/opentelemetry-go

The push model of OTel may work better when FlagD is running as a sidecar.

AlexsJones:
I wouldn't want to use a push model for flagD, I would caution that as we're adding dependencies and requirements. 
A Prometheus scrappable endpoint feels like a light weight enough implementation to be useful without mandating additional infrastructure.

AlexsJones:
I wouldn't want to use a push model for flagD, I would caution that as we're adding dependencies and requirements. 
A Prometheus scrappable endpoint feels like a light weight enough implementation to be useful without mandating additional infrastructure.

PR #152, opened 2022-08-22
feat: http unix socket support

james-milligan:
Adds unix socket support to bridge the connection between the http proxy and grpc service.

When a unix socket path is provided, this new implementation is used in favour of the standard tcp conection

PR #1967, opened 2026-05-25
chore: resolve open dependabot security alerts

jonathannorris:
## Summary

Resolved 2 of 7 open Dependabot security alerts. The remaining 5 alerts are for `github.com/docker/docker` and currently have no patched version available upstream.

## Dependabot Alerts Resolved

| Alert | Package | Severity | Fix |
|-------|---------|----------|-----|
| #190 | `js-cookie` | **high** | Added npm override to `>=3.0.7` in `playground-app/package.json` (transitive via `react-use`) |
| #189 | `github.com/containerd/containerd/v2` | **high** | Bumped to `v2.2.4` via `go get` + `go mod tidy` in `test/integration/go.mod` |

## Unresolvable Alerts

These Docker-related alerts in `test/integration/go.mod` (transitive via `testcontainers-go`) have no upstream patched version yet:

- #184 (high): Docker `PUT /containers/{id}/archive` executes container binary on the host
- #185 (medium): Docker race condition in `docker cp` allows creation of arbitrary empty files
- #186 (high): Docker race condition in `docker cp` allows bind mount redirection
- #151 (medium): Moby off-by-one error in plugin privilege validation
- #152 (high): Moby AuthZ plugin bypass — advisory references `29.3.1` but no such release exists yet


playground-app/package.json
```diff
@@ -17,6 +17,9 @@
     "react-dom": "^19.0.0",
     "react-use": "^17.6.0"
   },
+  "overrides": {
+    "js-cookie": "^3.0.7"
+  },
```
toddbaert, line 22: Can this override be avoided? Is there a transitive dep we can update?

jonathannorris: `react-use` is the only dep pulling in `js-cookie`, and it's a production dependency. The latest version (`17.6.0`) still requires `js-cookie: ^2.2.1`, so there's nothing to bump. That said, I don't think we're actually affected: the vulnerability only triggers when a JSON-derived object is passed as the `attributes` argument to `Cookies.set` or similar, and the playground only imports `useMedia` and `useObservable` from `react-use`. No cookie APIs are used. I'd lean toward dismissing the alert.

jonathannorris:
Superseded by #1980 (on the canonical `chore/dependabot-alerts` branch), which carries the same containerd + js-cookie fixes. Closing to keep a single active dependabot PR.

PR #1970, opened 2026-05-27
docs: remove outdated State Resync Event section

### LINKED ISSUE #1809: [DOC] Remove outdated Resync Event section for Sync sources
### Change in the documentation

**Context:**  
Within the [State Resync Events documentation](https://flagd.dev/concepts/syncs/#state-resync-events), we describe how `flagd` handles synchronization when flags are deleted from one source. Historically, a resync was triggered in cases where a flag might also exist in another source, ensuring that the state from the secondary (less prioritized) source was taken into account. This was intended to mitigate potential user errors but introduced complexity and unpredictable behavior.

**Problem:**  
With the changes introduced in #1797 , it was clarified that flagd does not currently handle cases where **the same flagset ID is defined across multiple sources**. This limitation is not reflected in the documentation, which still implies behavior that is not implemented. Specifically, flagd does not perform a resync or reconciliation between sources in such scenarios, leaving users to manage this complexity themselves. If unaddressed, this could lead to confusion or unexpected outcomes for users who rely on the outdated documentation.

**Focus of this Ticket:**  
- **Update the Documentation:**  
  Remove the outdated explanation of resync behavior in the [State Resync Events](https://flagd.dev/concepts/syncs/#state-resync-events) section.  
- **Clarify Limitations:**  
  Explicitly document the current guideline: users **must not specify flags with the same flagset ID in multiple sources.** Highlight that `flagd` does not handle such scenarios and that users must configure their sources accordingly.  
- **Prevent Misunderstanding:**  
  Ensure the updated documentation communicates the potential risks or unexpected behavior that could occur if this guideline is not followed.

By addressing this, we will align the documentation with the actual behavior of `flagd`, ensuring clarity for users and reducing the likelihood of misconfiguration or confusion.

kunalworldwide:
## Description

Removes the "State Resync Events" section from  as it describes behavior that flagd does not actually support.

## Context

Issue #1797 clarified that flagd does not handle cases where the same flag key is defined across multiple sources. The removed section incorrectly implied that when a higher-priority source deletes its version of a flag, a resync event would restore the flag definition from a lower-priority source. This is not accurate behavior.

The section being removed included:
- The "State Resync Events" heading and all its descriptive paragraphs
- References to flag-merge-3.svg and flag-merge-4.svg diagrams

This is a minimal documentation-only fix with no code changes.

Closes #1809

### DIFF
--- a/docs/concepts/syncs.md
+++ b/docs/concepts/syncs.md
@@ -149,21 +149,3 @@ When `flagd` is started with the command defined above, `source-B` takes priorit
 Using the above example, if a flag key is duplicated across all 3 sources, then the definition from `source-C` would be the only one stored in the merged state.
 
 ![flag merge 2](../images/flag-merge-2.svg)
-
-### State Resync Events
-
-Given the above example, the `source-A` and `source-B` 'versions' of flag definition the `foo` have been discarded, so if a delete event in `source-C` results in the removal of the `foo`flag, there will no longer be any reference of `foo` in flagd's store.
-
-As a result of this flagd will return `FLAG_NOT_FOUND` errors, and the OpenFeature SDK will always return the default value.
-
-To prevent flagd falling out of sync with its flag sources during delete events, resync events are used.
-When a delete event results in a flag definition being removed from the merged state, the full set of definition is requested from all flag sources, and the merged state is rebuilt.
-As a result, the value of the `foo` flag from `source-B` will be stored in the merged state, preventing flagd from returning `FLAG_NOT_FOUND` errors.
-
-![flag merge 3](../images/flag-merge-3.svg)
-
-In the example above, a delete event results in a resync event being fired, as `source-C` has deleted its 'version' of the `foo`, this results in a new merge state being formed from the remaining definition.
-
-![flag merge 4](../images/flag-merge-4.svg)
-
-Resync events may lead to further resync events if the returned flag definition result in further delete events, however the state will eventually be resolved correctly.


toddbaert:
Removal looks correct. Could you also add the note from #1809 that users shouldn't reuse a flagset ID across sources (and the risks if they do)? That was part of the ask here.

docs/concepts/syncs.md
```diff
@@ -149,21 +149,3 @@ When `flagd` is started with the command defined above, `source-B` takes priorit
...
-
-Given the above example, the `source-A` and `source-B` 'versions' of flag definition the `foo` have been discarded, so if a delete event in `source-C` results in the removal of the `foo`flag, there will no longer be any reference of `foo` in flagd's store.
-
-As a result of this flagd will return `FLAG_NOT_FOUND` errors, and the OpenFeature SDK will always return the default value.
-
-To prevent flagd falling out of sync with its flag sources during delete events, resync events are used.
-When a delete event results in a flag definition being removed from the merged state, the full set of definition is requested from all flag sources, and the merged state is rebuilt.
-As a result, the value of the `foo` flag from `source-B` will be stored in the merged state, preventing flagd from returning `FLAG_NOT_FOUND` errors.
-
-![flag merge 3](../images/flag-merge-3.svg)
```
toddbaert, line 163: `flag-merge-3.svg` and `flag-merge-4.svg` were only used here; they're orphaned now. Might be worth deleting them too.

kunalworldwide:
@toddbaert good eye — those two SVGs are indeed orphaned after removing the State Resync Events section. I'll delete `docs/images/flag-merge-3.svg` and `docs/images/flag-merge-4.svg` and include that in the next push.

PR #1971, opened 2026-05-28
feat(sync): forward S3 URI query parameters to the bucket URL

### LINKED ISSUE #1961: [FEATURE] Add support for use_path_style to S3 syncs.
### Requirements

The flagd daemon seems to make use of the gocloud library for blob syncs. The documentation for s3blob.URLOpener, https://pkg.go.dev/gocloud.dev/blob/s3blob#URLOpener, states that the argument use_path_style is supported by s3blob. However, the flagd daemon does not seem to offer a clean mechanism to enable the use_path_style option. 

I don't have a strong opinion on what such a mechanism should look like, but the following comes to mind:

Consume an environment in flagd similar to how the various "AWS_"-prefixed environment variables are consumed within s3blob.

Or, perhaps allow for query parameter parsing in flagd such that those arguments can be passed onto s3blob.  For example, 

`flagd start --uri "s3://my-bucket/my-flag.json?use_path_style=true"`

---
For reference, the s3blob documentation states the following:

The following S3-specific query options are also supported:
- ssetype: The type of server side encryption used (AES256, aws:kms, aws:kms:dsse)
- kmskeyid: The KMS key ID for server side encryption
- accelerate: A value of "true" uses the S3 Transfer Accleration endpoints
- use_path_style: A value of true sets the UsePathStyle option.
- s3ForcePathStyle: Same as use_path_style, for backwards compatibility with V1.
- disable_https: A value of true disables HTTPS in the Endpoint options.

mvanhorn:
## Summary

Forward S3 URI query parameters (`use_path_style`, `region`, `accelerate`, `disable_https`, etc.) to the bucket URL so gocloud's `s3blob` driver actually reads them. This lets flagd sync from path-style or non-AWS S3-compatible endpoints (MinIO, LocalStack, Ceph).

## Why this matters

`newS3` in `core/pkg/sync/builder/syncbuilder.go` splits the source URI with `regS3` (`^s3://.+?/`), which matches only up to the first `/`. For `s3://my-bucket/my-flag.json?use_path_style=true` that leaves the query string on the object name (`my-flag.json?use_path_style=true`) rather than on the bucket URL, which is where `s3blob.URLOpener` reads its options. The option is silently dropped and the object key is corrupted with a trailing `?...`.

This matches what the #1961 thread found: @kawingliu6 hit `strconv.ParseBool: parsing "true/"` when passing the parameter through, and @erka asked whether `s3://my-bucket/my-flag.json?use_path_style=true` should work. With this change it does.

## Changes

`newS3` now splits any `?query` off the URI before the bucket/object regex split, then re-attaches it to the bucket URL (trimming the trailing slash so the host stays well-formed) and leaves the object key clean. gcs and azblob are untouched; the fix is scoped to S3, where the issue and `s3blob`'s documented query options apply.

## Testing

Added a `Test_S3Config` case asserting `s3://bucket/path/to/object?use_path_style=true` resolves to bucket `s3://bucket?use_path_style=true` and object `path/to/object`. `go test ./pkg/sync/builder/...` passes in the `core` module; `gofmt` clean.

Closes #1961

### DIFF
--- a/core/pkg/sync/builder/syncbuilder.go
+++ b/core/pkg/sync/builder/syncbuilder.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/open-feature/flagd/core/pkg/logger"
 	"github.com/open-feature/flagd/core/pkg/sync"
@@ -291,8 +292,14 @@ func (sb *SyncBuilder) newS3(config sync.SourceConfig, logger *logger.Logger) (*
 	// Extract bucket uri and object name from the full URI:
 	// s3://bucket/path/to/object results in s3://bucket/ as bucketUri and
 	// path/to/object as an object name.
-	bucketURI := regS3.FindString(config.URI)
-	objectName := regS3.ReplaceAllString(config.URI, "")
+	rawURI, query, hasQuery := strings.Cut(config.URI, "?")
+	bucketURI := regS3.FindString(rawURI)
+	objectName := regS3.ReplaceAllString(rawURI, "")
+	if hasQuery && query != "" {
+		// s3blob reads use_path_style/region/etc. from the bucket URL query string;
+		// the bucket host must not carry a trailing slash before "?".
+		bucketURI = strings.TrimSuffix(bucketURI, "/") + "?" + query
+	}
 
 	interval, poller, err := newPoller(config)
 	if err != nil {
--- a/core/pkg/sync/builder/syncbuilder_test.go
+++ b/core/pkg/sync/builder/syncbuilder_test.go
@@ -464,6 +464,13 @@ func Test_S3Config(t *testing.T) {
 			expectedInterval: defaultInterval,
 		},
 		{
+			name:             "bucket options in query",
+			uri:              "s3://bucket/path/to/object?use_path_style=true",
+			expectedBucket:   "s3://bucket?use_path_style=true",
+			expectedObject:   "path/to/object",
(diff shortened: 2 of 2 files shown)


toddbaert (approved):

beeme1mr (approved):
Thanks @mvanhorn! By the way, the June oven looks pretty cool 😄

erka:
This PR solves some part of the issue but it panics in the end.

If I run it 
```sh
 flagd start --uri "s3://my-bucket/example_flags.json?use_path_style=true&region=garage&endpoint=http://127.0.0.1:3900"
```
I see this output

```sh
panic: source s3://my-bucket?use_path_style=true&region=garage&endpoint=http://127.0.0.1:3900example_flags.json is not registered in the store
```

and the known source is `s3://my-bucket/example_flags.json?use_path_style=true&region=garage&endpoint=http://127.0.0.1:3900`

The panic occurs in https://github.com/open-feature/flagd/blob/3aec90a1fc88ff1569e6dde1c3f30513de37d35d/core/pkg/store/store.go#L235-L237

@mvanhorn

mvanhorn:
Thanks @beeme1mr for landing the S3 URI query-parameter forwarding for the sync source.

mvanhorn:
Really appreciate the merge @beeme1mr. Forwarding the S3 URI query parameters through makes the sync config a lot more flexible.

mvanhorn:
Appreciate the merge @beeme1mr, S3 query parameter forwarding should make presigned-style bucket configs workable.

PR #1974, opened 2026-05-29
fix(sync): panic on s3 URI with query string

### LINKED ISSUE #1973: S3 sync URI with query string panics on Update
After #1971, starting flagd with an S3 sync URI that contains query parameters panics:

```sh
flagd start --uri "s3://my-bucket/example_flags.json?use_path_style=true&region=garage&endpoint=http://127.0.0.1:3900"
```

```
panic: source s3://my-bucket?use_path_style=true&region=garage&endpoint=http://127.0.0.1:3900example_flags.json is not registered in the store
```

The store registers the original `--uri`, but `blob.Sync` emits `Source` as `Bucket + Object`. #1971 moved the query onto the bucket URL (so gocloud reads `use_path_style` etc.), making the concatenation diverge from the registered source. `Store.Update` then panics here: https://github.com/open-feature/flagd/blob/3aec90a1fc88ff1569e6dde1c3f30513de37d35d/core/pkg/store/store.go#L235-L237

Originally reported by @erka in https://github.com/open-feature/flagd/issues/1971#issuecomment-4567497796.

toddbaert:
#1971 put the query string on the bucket URL (so gocloud reads it), but `blob.Sync` still emits `Source` as `Bucket + Object`, so it no longer matches the registered URI and `Store.Update` panics.

Pulled the split/join into `internal/bloburi` and rebuild the URI when emitting `DataSync.Source` so that things stay consistent if there's changes here - added a test for that as well to lock them as inverses.

Verified manually/locally against MinIO.

Closes #1973

FYI @mvanhorn

### DIFF
--- a/core/pkg/sync/internal/bloburi/bloburi.go
+++ b/core/pkg/sync/internal/bloburi/bloburi.go
@@ -1,0 +1,32 @@
+// Package bloburi splits and rejoins blob sync URIs (s3://, gs://, azblob://).
+package bloburi
+
+import (
+	"regexp"
+	"strings"
+)
+
+// Split parses "scheme://bucket/key?opt=1" into bucket URL ("scheme://bucket?opt=1")
+// and object key ("key"). The query moves to the bucket URL because gocloud blob
+// drivers read driver options (e.g. s3blob use_path_style, region) from there.
+// schemeRegex must match through the first "/" after the scheme (e.g. "^s3://.+?/").
+func Split(uri string, schemeRegex *regexp.Regexp) (bucket, object string) {
+	raw, query, hasQuery := strings.Cut(uri, "?")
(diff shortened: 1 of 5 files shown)


core/pkg/sync/internal/bloburi/bloburi.go
```diff
@@ -0,0 +1,32 @@
+// Package bloburi splits and rejoins blob sync URIs (s3://, gs://, azblob://).
```
toddbaert, line 1: I extracted these because they need to be inverses of each-other for the logic to work. I've added tests that ensure that as well.

beeme1mr (approved):

erka (approved):
well done @toddbaert !

PR #1982, opened 2026-06-30
feat(evaluator): add lower/upper string casing operators

### LINKED ISSUE #1916: Support case-insensitive string matching in targeting rules
## Context

At KubeCon EU 2026, the flagd discussion ([recap](https://openfeature.dev/blog/kubecon-eu-2026-recap)) identified case-insensitive matching as a missing capability in targeting rules.

## Problem

Flag evaluation rules currently use case-sensitive string matching. This can lead to unexpected behavior when targeting context values have inconsistent casing (e.g. `user@example.com` vs `User@Example.com`).

## Proposal

Add `lower` and/or `upper` string transform operations that can be composed with existing operators. This avoids duplicating every string operator with a case-insensitive variant and fits naturally with the JSON Logic model.

Example:

```json
{ "==": [{ "lower": [{ "var": "email" }] }, "user@example.com"] }
```

This approach is composable with any existing string operator (`==`, `in`, `starts_with`, etc.) rather than needing dedicated `equals_ignore_case`, `contains_ignore_case` variants for each.

## Related

- #1842: Unified flagd-evaluator PoC

anxkhn:
This adds two single-argument JSONLogic operators, `lower` and `upper`, that fold a string to ASCII lower/upper case. Composed with the existing operators (`==`, `in`, `starts_with`, ...) they enable case-insensitive targeting without duplicating every comparison operator with an ignore-case variant, which is the shape requested in the issue:

```json
{ "==": [{ "lower": [{ "var": "email" }] }, "user@example.com"] }
```

Casing is restricted to ASCII (`A-Z`/`a-z`) per @toddbaert's recommendation on the issue: `strings.ToLower`/`ToUpper` give simple 1:1 mapping in Go, but Java/Python/Rust/JS use full (and Java locale-sensitive) mappings, so a fold like `ß -> SS` would diverge across providers. ASCII-only is trivially identical everywhere and covers the real use case (emails, non-IDN domains). All non-ASCII bytes are left unchanged. This is documented as the intended scope.

### Changes

- `core/pkg/evaluator/string_casing.go`: `lower`/`upper` operators, mirroring the `string_comparison.go` (`starts_with`/`ends_with`) structure. Single arg accepted as bare string or 1-element array; non-string/invalid input logs and returns `nil`, matching the existing operators' error idiom.
- `core/pkg/evaluator/json.go`: register both operators in `NewResolver`.
- `docs/schema/v0/targeting.json`: `stringCasingRule` added to `anyRule`.
- Docs: new custom-operation page, operator-table row, mkdocs nav entry.
- `core/pkg/evaluator/string_casing_test.go`: table-driven tests (compose with `==`/`starts_with`, ASCII-only proof, non-string fallback).

Closes #1916

### Notes

- Restricted to ASCII per the maintainer comment; intentionally does not use `strings.ToLower/ToUpper`.
- `make test-core` (`-race`), `golangci-lint` v2.7.2, and `markdownlint` all pass locally; evaluator coverage ~87.6%.


anxkhn:
Good catch. The two happy-path `wantErr` closures returned `false`, which tripped the `if !tt.wantErr(...) { return }` guard and skipped `assert.Equalf`, so the `"a"` outputs for the bare-string and single-element-array cases were never checked. I flipped them to `return true` (matching the error-case closures) so the loop now reaches the value assertion. Verified by temporarily corrupting `wantProperty`: it passed before the change and fails after, so both cases are genuinely asserted now.

toddbaert:
@anxkhn Thanks... I'm open to this, and I think it's a good idea; but it needs to be a much larger effort. We can't just do this in flagd itself - we have to do it consistently across the ecosystem.

Support needs to be added:

- here (flagd itself)
- all providers ([java](https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd), [js](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd), [python](https://github.com/open-feature/python-sdk-contrib/tree/main/providers/openfeature-provider-flagd), etc)
- the [JSON schema](https://github.com/open-feature/flagd-schemas/tree/main/json)
- coverage needs to be added in the [gherkin suite](https://github.com/open-feature/flagd-testbed)

I'm happy to create issues for all of these, and help, certainly it's not all your responsibility. The best place to start though would probably be an [ADR](https://github.com/open-feature/flagd/tree/main/docs/architecture-decisions), so the community can agree on the behavior of the operator.

toddbaert (changes requested):
This is a bigger effort than one PR - blocking for now until we have some consensus: https://github.com/open-feature/flagd/pull/1982#issuecomment-5025729572.

anxkhn:
closing this implementation pr until the adr and cross-repository design are agreed; i'll revisit it after that coordination.

PR #1987, opened 2026-07-05
fix(flagd-proxy): route management-port gRPC health by the correct content-type

anxkhn:
```markdown
### This PR

Fixes the h2c content-type routing on the flagd-proxy management/metrics port so
that gRPC traffic reaches the registered gRPC health server.

`startMetricsServer` in `flagd-proxy/pkg/service/server.go` registers a
`grpc_health_v1` health server behind an h2c handler that dispatches HTTP/2
requests to gRPC based on the request `Content-Type`. The predicate matched the
prefix `application/grpcServer`:

```go
// if this is 'application/grpcServer' and HTTP2, handle with gRPC, otherwise HTTP.
if request.ProtoMajor == 2 && strings.HasPrefix(request.Header.Get("Content-Type"), "application/grpcServer") {
```

`application/grpcServer` is not a real media type. Per the gRPC-over-HTTP2
protocol, gRPC clients send `Content-Type: application/grpc` (optionally with a
`+proto` / `+json` suffix). Because no client sends `application/grpcServer`,
the `HasPrefix` check never matched real gRPC traffic: every gRPC request on the
management port fell through to the plain HTTP mux (which only serves
`/healthz`, `/readyz`, `/metrics`), so the registered gRPC health server on that
port was unreachable.

This changes the prefix (and its comment) to `application/grpc`, which makes the
dispatcher identical to the equivalent one in flagd's connect service
(`flagd/pkg/service/flag-evaluation/connect_service.go`), the authoritative
reference for the intended behavior. After this change, `application/grpcServer`
no longer appears anywhere in the repository.

### Regression test

Adds `flagd-proxy/pkg/service/server_test.go`
(`TestMetricsServerRoutesGRPCHealth`): it starts the management server on an
ephemeral port, dials a real gRPC client over plaintext HTTP/2 (h2c), and
asserts a `grpc_health_v1.Check` is routed to the health server and returns
`SERVING`.

- Before the fix, the check is served by the plain HTTP mux and fails
  (HTTP 404, `text/plain`, `code = Unimplemented`).
- After the fix, it returns `SERVING`.

The test uses only the standard library plus gRPC packages already used in the
module (`google.golang.org/grpc`, `.../health/grpc_health_v1`,
`.../credentials/insecure`); `go.mod` and `go.sum` are unchanged.

### Validation

- `cd flagd-proxy && go build ./...`, `go vet ./pkg/service/...`
- `go test -race -covermode=atomic -cover -short ./pkg/...` (all pass; new test
  run with `-race -count` shows no flake and no data race)
- `golangci-lint run ./...`

The h2c gRPC routing path on the management port was previously untested, so
this also closes a coverage gap.
```

### DIFF
--- a/flagd-proxy/pkg/service/server.go
+++ b/flagd-proxy/pkg/service/server.go
@@ -137,8 +137,8 @@ func (s *Server) startMetricsServer() error {
 	mux.Handle("/metrics", promhttp.Handler())
 
 	handler := http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
-		// if this is 'application/grpcServer' and HTTP2, handle with gRPC, otherwise HTTP.
-		if request.ProtoMajor == 2 && strings.HasPrefix(request.Header.Get("Content-Type"), "application/grpcServer") {
+		// if this is 'application/grpc' and HTTP2, handle with gRPC, otherwise HTTP.
+		if request.ProtoMajor == 2 && strings.HasPrefix(request.Header.Get("Content-Type"), "application/grpc") {
 			grpcServer.ServeHTTP(writer, request)
 		} else {
 			mux.ServeHTTP(writer, request)
--- a/flagd-proxy/pkg/service/server_test.go
+++ b/flagd-proxy/pkg/service/server_test.go
@@ -1,0 +1,97 @@
+package service
+
+import (
+	"context"
+	"fmt"
+	"net"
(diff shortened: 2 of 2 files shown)


toddbaert:
Looks like this used to be correct and was broken with a bulk find/replace. Thanks again @anxkhn

toddbaert (approved):

PR #1990, opened 2026-07-07
fix(sync): handle non-string YAML map keys in config parsing

anxkhn:
**What**

A YAML flag config that contains a nested mapping with a non-string key (for
example an unquoted numeric key inside an object variant, or under `metadata`)
currently fails to load entirely, and the whole file is rejected with a cryptic
internal error:

```
json: unsupported type: map[interface {}]interface {}
```

This affects every YAML sync source (`file`, `http`, `blob`), all of which route
through `utils.ConvertToJSON` -> `utils.YAMLToJSON`.

**Reproduction**

A minimal, realistic flag definition that uses an object variant with numeric
keys is enough to trigger it:

```yaml
flags:
  myObjectFlag:
    state: ENABLED
    defaultVariant: config
    variants:
      config:
        1: first
        2: second
```

Loading this config (via a `file:`, `http:`, or `blob:` sync) fails the entire
file rather than just that flag, with the `map[interface {}]interface {}` error
above.

**Root cause**

`YAMLToJSON` decodes the document into `map[string]interface{}` with
`gopkg.in/yaml.v3` and then `json.Marshal`s the result. yaml.v3 only forces
string keys for the *top-level* map (because the decode target type does).
Any *nested* mapping with a non-string key is decoded as
`map[interface{}]interface{}`, which `encoding/json` cannot marshal, so the
marshal step fails and the whole config is rejected.

**Fix**

Recursively convert nested `map[interface{}]interface{}` into
`map[string]interface{}` (stringifying the keys) before marshaling. yaml.v3 is
kept as the parser, so YAML 1.2 value semantics are unchanged:

- unquoted `on/off/yes/no/NO` still parse as strings (no "Norway problem"),
- timestamps keep their existing normalization,
- duplicate keys and top-level scalar inputs still error exactly as before.

The change is purely additive: configs that already loaded are unaffected, and
configs with nested non-string keys now load instead of failing. No dependency
change.

**Tests**

Adds two table cases to `core/pkg/utils/yaml_test.go`:

- `nested non-string map keys` - the object-variant config above now converts to
  JSON with stringified keys (`"1"`, `"2"`); red before this change, green after.
- `unquoted boolean-like keys and values stay strings` - pins YAML 1.2 value
  semantics so a future parser swap that reintroduced boolean coercion would be
  caught. Passes on `main` too.

`go test -race -covermode=atomic -cover -short ./pkg/utils/...` passes (97.2%
coverage); `golangci-lint` (v2.7.2) is clean; `go mod tidy` leaves go.mod/go.sum
unchanged.

### DIFF
--- a/core/pkg/utils/yaml.go
+++ b/core/pkg/utils/yaml.go
@@ -18,10 +18,39 @@ func YAMLToJSON(rawFile []byte) (string, error) {
 		return "", fmt.Errorf("error unmarshaling yaml: %w", err)
 	}
 
-	r, err := json.Marshal(ms)
+	r, err := json.Marshal(convertMapKeys(ms))
 	if err != nil {
 		return "", fmt.Errorf("error marshaling json: %w", err)
 	}
 
 	return string(r), err
+}
+
+// convertMapKeys recursively converts any map[interface{}]interface{} into
+// map[string]interface{} so it can be marshaled by encoding/json. gopkg.in/yaml.v3
+// decodes a nested mapping that has a non-string key (e.g. an unquoted numeric key
+// in an object variant) as map[interface{}]interface{}, which json.Marshal cannot
+// handle; stringifying the keys mirrors how YAML-to-JSON conversion is expected to
+// behave and preserves YAML 1.2 value semantics.
+func convertMapKeys(in interface{}) interface{} {
+	switch v := in.(type) {
+	case map[string]interface{}:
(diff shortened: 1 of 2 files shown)


toddbaert (approved):

toddbaert:
Nice fix @anxkhn ... I'm curious if you ran into this or just found it in code?

PR #1991, opened 2026-07-07
feat: improve ETag and body hash support to blob/http sync

### LINKED ISSUE #1988: feat(sync): dedupe unchanged blob payloads via ETag and content hash
Blob sync (`core/pkg/sync/blob/blob_sync.go:82-95`) dedupes on `ModTime` only. Any bucket op that bumps ModTime without changing bytes (re-upload, tag/ACL edit, replication, terraform re-apply) causes a redundant republish.

HTTP sync already handles this correctly using ETag + SHA-3 body hash (`core/pkg/sync/http/http_sync.go:180-201, 208-223`).

### Proposed change

Bring blob to parity with HTTP:
- Use `attrs.ETag` when available (skip fetch on match).
- Fall back to existing ModTime check.
- After fetch, SHA-3 the body; skip publish if unchanged, as a backup for lack of etag or ModTimes with no material changes.

### Test

Unit tests with mock bucket covering:
- ETag unchanged -> no fetch, no publish.
- ETag changed, ModTime bumped, body identical -> fetch but no publish (SHA guard).
- ETag absent, ModTime unchanged -> no fetch, no publish.
- ETag absent, ModTime bumped, body identical -> fetch but no publish (SHA guard).
- ETag absent, ModTime bumped, body changed -> fetch and publish.
- ETag changed, body changed -> fetch and publish.

leakonvalinka:
- introduces ETag & hash of body's content to the blob sync to avoid publishing "fake" changes
- improves body hashing reliability (immaterial changes don't fire the sync).

Fixes: https://github.com/open-feature/flagd/issues/1988

### DIFF
--- a/core/pkg/sync/blob/blob_sync.go
+++ b/core/pkg/sync/blob/blob_sync.go
@@ -3,10 +3,12 @@ package blob
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"path/filepath"
+	stdsync "sync"
 	"time"
 
 	"github.com/open-feature/flagd/core/pkg/logger"
@@ -28,6 +30,7 @@ type Sync struct {
 	Poller      polling.Poller
 	Logger      *logger.Logger
 	Interval    uint32
+	mu          stdsync.Mutex
 	ready       bool
 	lastUpdated time.Time
 	lastETag    string
@@ -45,6 +48,8 @@ func (hs *Sync) Init(_ context.Context) error {
 }
 
 func (hs *Sync) IsReady() bool {
+	hs.mu.Lock()
+	defer hs.mu.Unlock()
 	return hs.ready
 }
 
@@ -58,7 +63,9 @@ func (hs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 		return err
 	}
 
+	hs.mu.Lock()
 	hs.ready = true
+	hs.mu.Unlock()
 
 	hs.Logger.Debug(fmt.Sprintf("polling %s/%s every %ds (offset: %ds)",
 		hs.Bucket, hs.Object, hs.Interval, hs.Poller.Offset()))
@@ -101,17 +108,15 @@ func (hs *Sync) sync(ctx context.Context, dataSync chan<- sync.DataSync, skipCha
 		return fmt.Errorf("couldn't get object: %v", err)
 	}
 
-	// only publish if the content actually differs from what we last saw.
-	if !skipChangeDetection && bodySHA == hs.lastBodySHA {
+	// only publish if the content actually differs
+	if !skipChangeDetection && hs.bodyUnchanged(bodySHA) {
 		hs.Logger.Debug("configuration hasn't changed, skipping publishing")
 		hs.updateState(attrs, bodySHA)
 		return nil
 	}
 
 	hs.Logger.Debug(fmt.Sprintf("configuration updated: %s", msg))
-	if !skipChangeDetection {
-		hs.updateState(attrs, bodySHA)
-	}
+	hs.updateState(attrs, bodySHA)
 	dataSync <- sync.DataSync{FlagData: msg, Source: bloburi.Join(hs.Bucket, hs.Object)}
 	return nil
 }
@@ -120,6 +125,8 @@ func (hs *Sync) sync(ctx context.Context, dataSync chan<- sync.DataSync, skipCha
 // based on its attributes alone, allowing us to skip fetching the full object.
 // prefers the ETag and falls back to the mod time for stores that don't expose one.
 func (hs *Sync) attributesUnchanged(attrs *blob.Attributes) bool {
+	hs.mu.Lock()
+	defer hs.mu.Unlock()
 	if attrs.ETag != "" {
 		return hs.lastETag == attrs.ETag
 	}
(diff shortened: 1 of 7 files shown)


toddbaert:
Initially, I misread the code and thought that we weren't properly updating the bodySha and using it on the next request; we are (I misread) but I think we are missing a test that specifically asserts this. The existing `TestBlobSync_ChangeDetection` cases prove state is set after the *baseline* sync (since the "ETag unchanged -> no fetch" case would fail otherwise), but nothing exercises the full path involving an update to the blob while syncing. A small three-sync case would give confidence here: initial publish -> change + publish -> identical tick asserts no fetch and no publish. That way if someone later moves `updateState` back into only the skip branch, we'd catch it.

spec
toddbaert, line 1: Please revert the submodule change.

schemas
toddbaert, line 1: Please revert the submodule change.

test-harness
toddbaert, line 1: Please revert the submodule change.

core/pkg/sync/blob/blob_sync.go
```diff
@@ -113,32 +150,38 @@ func (hs *Sync) getBucket(ctx context.Context) (*blob.Bucket, error) {
...
+		return "", "", fmt.Errorf("error converting blob data to json: %w", err)
 	}
-	return json, nil
+	return json, hs.generateSha(data), nil
+}
+
+func (hs *Sync) generateSha(body []byte) string {
+	hasher := sha3.New256()
+	hasher.Write(body)
+	return base64.URLEncoding.EncodeToString(hasher.Sum(nil))
```
toddbaert: Instead of hashing the body directly, hash a canonical representation of the payload; i.e. JSON-parse the data first, then hash the re-serialized output. Go's encoding/json sorts map keys alphabetically and has strong stability guarantees on output. Hashing that means upstream key-ordering differences (not stable per the JSON spec) and YAML comments/whitespace won't cause spurious "changes". Ideally we'd apply this across all sync sources (file, blob, http), not just blob. It's a bit of scope creep, but IMO a good thing to add here.

core/pkg/sync/blob/blob_sync.go
```diff
@@ -73,38 +84,81 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
...
+	hs.updateState(attrs, bodySHA)
 	dataSync <- sync.DataSync{FlagData: msg, Source: bloburi.Join(hs.Bucket, hs.Object)}
 	return nil
 }
 
+// attributesUnchanged reports whether the object can be considered unchanged
+// based on its attributes alone, allowing us to skip fetching the full object.
+// prefers the ETag and falls back to the mod time for stores that don't expose one.
+func (hs *Sync) attributesUnchanged(attrs *blob.Attributes) bool {
+	hs.mu.Lock()
```
alexandraoberaigner, line 125: since we are adding a locking mechanism, would it make sense to add some kind of small/lightweight concurrency/race-safety test? 
e.g. concurrent IsReady(), ReSync(), and regular sync calls on the same Sync, mainly intended to run under go test -race (which is already set up for our tests. see [makefile](https://github.com/open-feature/flagd/blob/main/Makefile#L47)). Just to make sure that there are no race conditions or deadlocks are happening around that data thats protected via the mutex

toddbaert: I did my best to add this here. It basically exercises a lot of the locks. I commented some out after adding this test and indeed I see race errors:

```
Goroutine 50 (running) created at:
  github.com/open-feature/flagd/core/pkg/sync/blob.TestBlobSync_ConcurrentChanges()
      /home/local_toddbaert/git/flagd/core/pkg/sync/blob/blob_sync_test.go:398 +0x3fa
  testing.tRunner()
      /home/local_toddbaert/sdk/go1.25.8/src/testing/testing.go:1934 +0x21c
  testing.(*T).Run.gowrap1()
      /home/local_toddbaert/sdk/go1.25.8/src/testing/testing.go:1997 +0x44
==================
--- FAIL: TestBlobSync_ConcurrentChanges (0.05s)
    testing.go:1617: race detected during execution of test
```

core/pkg/sync/blob/mock_blob.go
```diff
@@ -1,72 +0,0 @@
-package blob
```
toddbaert, line 1: I renamed this because it was being shipped.

core/pkg/utils/hash.go
```diff
@@ -0,0 +1,37 @@
...
+	dec := json.NewDecoder(bytes.NewReader(body))
+	dec.UseNumber()
+
+	var parsed interface{}
+	if err := dec.Decode(&parsed); err != nil {
+		return body
+	}
+	// check for leftover garbage after valid json
+	var extra json.RawMessage
+	if err := dec.Decode(&extra); err != io.EOF {
```
toddbaert:
```suggestion
	if dec.Decode(&extra) != io.EOF {
```

core/pkg/sync/blob/blob_sync.go
```diff
@@ -73,38 +81,81 @@ func (hs *Sync) ReSync(ctx context.Context, dataSync chan<- sync.DataSync) error
 	return hs.sync(ctx, dataSync, true)
 }
 
-func (hs *Sync) sync(ctx context.Context, dataSync chan<- sync.DataSync, skipCheckingModTime bool) error {
+func (hs *Sync) sync(ctx context.Context, dataSync chan<- sync.DataSync, skipChangeDetection bool) error {
```
toddbaert: I think the name `skipChangeDetection` is confusing, because it skips our check for skipping publishes, and leads to double negatives in code (`!skipChangeDetection`). I think ` forcePublish` or something like that might be better.

toddbaert (approved):
Approving with 2 nits:

- https://github.com/open-feature/flagd/pull/1991/changes#r3597874620
- https://github.com/open-feature/flagd/pull/1991/changes#r3597889032

I also added the text @alexandraoberaigner recommended, mostly because I was curious how effective this loop/goroutine method would be

core/pkg/sync/blob/blob_sync.go
leakonvalinka: Thank you! I didn't get around to it before

core/pkg/sync/blob/blob_sync.go
leakonvalinka: Yea makes sense, I updated it!

core/pkg/sync/blob/blob_sync.go
leakonvalinka: For the record: I also changed it in the http sync now, but left the file sync as is.

alexandraoberaigner (approved):

PR #1992, opened 2026-07-10
fix(sync): avoid send on closed channel in fileinfo watcher Close

anxkhn:
The `FILEINFO` file watcher (`core/pkg/sync/file/fileinfo_watcher.go`) can crash flagd with `panic: send on closed channel` at shutdown.

The watcher's timer goroutine is the sole sender on `evChan` and `erChan`, but `Close()` closed both channels without coordinating with that goroutine. If the ticker fires while `Close()` runs, the goroutine can be mid-send on `evChan` and hit the panic. The realistic trigger is `filepath` `Sync()` returning on context cancel: its `defer fs.watcher.Close()` runs while the same context is stopping the timer goroutine, so a send can still be in flight exactly as the channel is closed.

This only affects the non-default `fileinfo` watcher (selected by `watchType == "fileinfo"`); the default `fsnotify` watcher is unaffected. The window is narrow, but the panic is unrecovered and takes the process down.

### Fix

`Close()` now stops the timer goroutine before it closes the channels the goroutine sends on:

- Add a `done` channel (closed once, guarded by a `sync.Once`) and a `sync.WaitGroup` to `fileInfoWatcher`.
- `Close()` signals `done`, waits for the goroutine to exit (`wg.Wait()`), and only then closes `evChan`/`erChan`, so no send can race with the close.
- `done` is also added to the timer goroutine's `select` and to every send in `update()`, so an in-flight send is aborted promptly and shutdown cannot deadlock on a full, unread buffer.

No happy-path behavior changes.

### Changes

- `core/pkg/sync/file/fileinfo_watcher.go`: add `done`/`stopOnce`/`wg`; make `Close()` stop-then-close; make the timer goroutine and the three sends (`erChan` in `run`, both `evChan` sends in `update`) abortable on `done`.
- `core/pkg/sync/file/fileinfo_watcher_test.go`: add `Test_fileInfoWatcher_Close_racesWithTimer`, which drives the timer at a fast tick with a stat func that emits an event every tick, then calls `Close()` with no reader draining the channels. It panics before this change and passes after. `makeTestWatcher` gains the `done` field so the existing close test keeps compiling.

### Notes

- `make test-core` (`-race`), `golangci-lint` v2.7.2, and `gofmt` all pass locally; `core/pkg/sync/file` coverage is ~69.9% and the full `core/pkg/...` `-race` suite has no regressions.
- The new test is red -> green: reverting only the `Close`/`run`/`update` coordination (keeping the field so it compiles) reproduces the exact `panic: send on closed channel`; with the fix it is race-clean across repeated runs.

### DIFF
--- a/core/pkg/sync/file/fileinfo_watcher.go
+++ b/core/pkg/sync/file/fileinfo_watcher.go
@@ -35,7 +35,7 @@ type fileInfoWatcher struct {
 	wg sync.WaitGroup
 }
 
-// NewFsNotifyWatcher returns a new fsNotifyWatcher
+// NewFileInfoWatcher returns a new fileInfoWatcher
 func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger) Watcher {
 	fiw := &fileInfoWatcher{
 		evChan:   make(chan fsnotify.Event, 32),
@@ -57,12 +57,13 @@ func (f *fileInfoWatcher) Close() error {
 	// signal the timer goroutine to stop and abort any in-flight send, then wait
 	// for it to exit before closing the channels it sends on. Closing before the
 	// goroutine has stopped would race with its sends -> send on closed channel.
+	// stopOnce guards the whole sequence so Close is idempotent.
 	f.stopOnce.Do(func() {
 		close(f.done)
+		f.wg.Wait()
+		close(f.evChan)
+		close(f.erChan)
 	})
-	f.wg.Wait()
-	close(f.evChan)
-	close(f.erChan)
 	return nil
 }
 
--- a/core/pkg/sync/file/fileinfo_watcher.go
+++ b/core/pkg/sync/file/fileinfo_watcher.go
@@ -26,6 +26,13 @@ type fileInfoWatcher struct {
 	// thread-safe interface to underlying files we are watching
 	mu      sync.RWMutex
 	watches map[string]fs.FileInfo // filename -> info
+	// done signals the timer goroutine to stop and aborts any in-flight send,
+	// so Close never closes evChan/erChan while the goroutine is still sending
+	done chan struct{}
+	// stopOnce guards done so Close is safe to call more than once
+	stopOnce sync.Once
+	// wg tracks the timer goroutine so Close can wait for it to exit
+	wg sync.WaitGroup
 }
 
 // NewFsNotifyWatcher returns a new fsNotifyWatcher
@@ -36,6 +43,7 @@ func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger) Watcher {
 		statFunc: getFileInfo,
 		logger:   logger,
 		watches:  make(map[string]fs.FileInfo),
+		done:     make(chan struct{}),
 	}
 	fiw.run(ctx, (1 * time.Second))
 	return fiw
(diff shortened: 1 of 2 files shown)


core/pkg/sync/file/fileinfo_watcher.go
```diff
@@ -44,9 +52,15 @@ func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger) Watcher {
 // fileInfoWatcher explicitly implements file.Watcher
 var _ Watcher = &fileInfoWatcher{}
 
-// Close calls close on the underlying fsnotify.Watcher
```
toddbaert, line 47: Whoops, that was inaccurate thanks for fixing.

core/pkg/sync/file/fileinfo_watcher.go
```diff
@@ -26,6 +26,13 @@ type fileInfoWatcher struct {
...
+	// done signals the timer goroutine to stop and aborts any in-flight send,
+	// so Close never closes evChan/erChan while the goroutine is still sending
+	done chan struct{}
+	// stopOnce guards done so Close is safe to call more than once
+	stopOnce sync.Once
+	// wg tracks the timer goroutine so Close can wait for it to exit
+	wg sync.WaitGroup
 }
 
 // NewFsNotifyWatcher returns a new fsNotifyWatcher
```
toddbaert:
```suggestion
// NewFileInfoWatcher returns a new fileInfoWatcher
```

core/pkg/sync/file/fileinfo_watcher_test.go
```diff
@@ -38,6 +39,37 @@ func Test_fileInfoWatcher_Close(t *testing.T) {
...
+		"foo": &mockFileInfo{name: "foo", modTime: time.Now()},
+	})
+	// every stat reports a newer modtime, so update() emits a Write event on
+	// every tick, keeping the timer goroutine busy sending on evChan
+	watcher.statFunc = func(_ string) (fs.FileInfo, error) {
+		return &mockFileInfo{name: "foo", modTime: time.Now().Add(time.Hour)}, nil
+	}
+
+	// nothing reads Events()/Errors(), so the buffer fills and the goroutine ends
+	// up blocked mid-send, which is exactly when the unguarded Close() panicked
```
toddbaert: >  so the buffer fills

Am I confused?

the 512 buffer never fills with ~20 events; the real mechanism is context.Background() never cancels, so the goroutine keeps ticking/sending after Close, I think... right?

I think the test is correct, but this isn't the right mechanism by which it works, unless I'm confused.

core/pkg/sync/file/fileinfo_watcher_test.go
```diff
@@ -38,6 +39,37 @@ func Test_fileInfoWatcher_Close(t *testing.T) {
...
+	time.Sleep(20 * time.Millisecond)
+
+	if err := watcher.Close(); err != nil {
+		t.Errorf("fileInfoWatcher.Close() error = %v", err)
+	}
+
+	// Close must have stopped the timer goroutine and closed both channels;
+	// draining the (buffered) events chan blocks forever unless it is closed
+	for range watcher.Events() {
+	}
```
toddbaert:
```suggestion
	for range watcher.Events() {
	    // Close must have stopped the timer goroutine and closed both channels;
	    // draining the (buffered) events chan blocks forever unless it is closed
	}
```

I think this will fix the sonar complaint.

toddbaert (approved):
This improves things, but I think coderabbit did spot some remaining issues that are similar [here](https://github.com/open-feature/flagd/pull/1992#discussion_r3561585017) and [here](https://github.com/open-feature/flagd/pull/1992#discussion_r3561585000). I think they are worth addressing.

Also some nits from me: 

- https://github.com/open-feature/flagd/pull/1992#discussion_r3598602841
- https://github.com/open-feature/flagd/pull/1992/changes#r3598596898

PR #1994, opened 2026-07-10
fix(certreloader): re-check reload condition under write lock

anxkhn:
## This PR

`GetCertificate` in `core/pkg/certreloader/certreloader.go` uses a
double-checked-locking pattern, but the second check is missing. It reads the
reload condition under the read lock, releases it, takes the write lock, and then
reloads the key pair from disk **unconditionally**:

```go
r.mu.RLock()
shouldReload := r.ReloadInterval != 0 && r.nextReload.Before(now)
r.mu.RUnlock()
if shouldReload {
    r.mu.Lock()
    defer r.mu.Unlock()
    cert, err := r.loadCertificate() // always runs, even if another caller just reloaded
    ...
    r.cert = &cert
    r.nextReload = now.Add(r.ReloadInterval)
    return r.cert, nil
}
```

The function's own comment states the intended behavior:

> If a reload is in progress this will block and we will skip reloading in the
> current call once we can continue

but nothing skips: after blocking on `r.mu.Lock()`, every caller that passed the
read-lock check calls `loadCertificate()` again.

`GetCertificate` is wired as `tls.Config.GetCertificate`, which the standard
library invokes per handshake and can run in parallel. So once `ReloadInterval`
has elapsed, a burst of concurrent handshakes all pass the read-lock check, then
each grabs the write lock in turn and re-reads and re-parses the cert/key pair
from disk, even though the first caller already reloaded and advanced
`nextReload`. That is a thundering herd of redundant disk reads at every reload
boundary on a busy TLS listener.

This is not a correctness/crash bug: access is serialized by the lock and the
served certificate is always valid. It defeats the single-reload optimization the
code is clearly trying to implement.

## The fix

Add the missing second check inside the write-locked critical section, so only
the first caller past the interval reloads and the rest return the
already-loaded certificate:

```go
r.mu.Lock()
defer r.mu.Unlock()
if r.ReloadInterval != 0 && r.nextReload.Before(now) {
    cert, err := r.loadCertificate()
    ...
    r.cert = &cert
    r.nextReload = now.Add(r.ReloadInterval)
}
return r.cert, nil
```

The re-check reuses `now` captured at the top of the function. The first caller
sets `nextReload = now.Add(ReloadInterval)`; any caller that was blocked on the
write lock observes a `now` no earlier than the first caller's, so for any real
interval its re-check is false and it correctly skips. Production behavior is
otherwise unchanged: a single caller past the interval still reloads exactly as
before.

To make the behavior testable, `tls.LoadX509KeyPair` is called through a
package-level variable (`var loadX509KeyPair = tls.LoadX509KeyPair`) so a test
can count reloads. Production always uses the stdlib function; this is a one-line
seam plus one changed call site, and the runtime path is byte-identical.

## Testing

Adds `TestConcurrentCertificateReload`, which fires 50 concurrent
`GetCertificate` calls past the interval and asserts the cert is reloaded from
disk exactly once. It parks all callers on the read lock, releases them together
so they evaluate the reload condition concurrently, and counts reloads via the
seam.

```
cd core
go build ./...
go test -race -covermode=atomic -cover -short ./pkg/certreloader/...
```

All green under `-race` (stable over repeated runs). Package coverage for
`core/pkg/certreloader` is 92.6%. `golangci-lint` (repo-pinned v2.7.2) reports 0
issues on the package.

Red/green: with only the new write-lock re-check reverted, the test reports 50
reloads for 50 concurrent callers and fails; with the fix it reports exactly 1.

### DIFF
--- a/core/pkg/certreloader/certreloader.go
+++ b/core/pkg/certreloader/certreloader.go
@@ -48,19 +48,28 @@ func (r *CertReloader) GetCertificate() (*tls.Certificate, error) {
 		// Need to release the read lock, otherwise we deadlock
 		r.mu.Lock()
 		defer r.mu.Unlock()
-		cert, err := r.loadCertificate()
-		if err != nil {
-			return nil, fmt.Errorf("failed to load TLS cert and key: %w", err)
+		// Re-check the condition now that we hold the write lock: a concurrent
+		// caller may have already reloaded while we were blocked, in which case
+		// we skip the redundant reload (double-checked locking).
+		if r.ReloadInterval != 0 && r.nextReload.Before(now) {
+			cert, err := r.loadCertificate()
+			if err != nil {
+				return nil, fmt.Errorf("failed to load TLS cert and key: %w", err)
+			}
+			r.cert = &cert
+			r.nextReload = now.Add(r.ReloadInterval)
 		}
-		r.cert = &cert
-		r.nextReload = now.Add(r.ReloadInterval)
 		return r.cert, nil
 	}
 	return r.cert, nil
 }
 
+// loadX509KeyPair loads a certificate/key pair from disk. It is a package
+// variable so tests can observe reloads; production always uses the stdlib.
+var loadX509KeyPair = tls.LoadX509KeyPair
+
 func (r *CertReloader) loadCertificate() (tls.Certificate, error) {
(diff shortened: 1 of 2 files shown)


toddbaert (approved):
Thanks again @anxkhn .

PR #1999, opened 2026-07-14
feat(sync): configurable gRPC keepalive enforcement on sync server

### LINKED ISSUE #1998: [FEATURE] Configurable gRPC keepalive enforcement policy on the flag-sync server
### Requirements

**Problem**

The flag-sync service exposes a long-lived `SyncFlags` gRPC stream that in-process providers consume. To keep that stream healthy across idle periods, providers send gRPC keepalive pings.

The sync gRPC server is currently constructed with a bare `grpc.NewServer(...)` and no `KeepaliveEnforcementPolicy`, so grpc-go's default enforcement applies: `MinTime: 5m` and `PermitWithoutStream: false`. Any client that pings more frequently than every 5 minutes — or pings while it holds no active stream — accumulates ping strikes, and the server responds with `GOAWAY ENHANCE_YOUR_CALM` (`too_many_pings`), tearing down the sync stream and forcing a reconnect.

This is a **server-side default mismatch**, not a single-client bug. flagd's own sync stream is long-lived and idle-tolerant by design, yet the server inherits grpc-go's generic-RPC defaults that assume pings are rare. Provider SDKs across languages set their own keepalive cadences; the Go provider is the sharpest case (a hardcoded 30s ping with `PermitWithoutStream: true` and no override), producing a GOAWAY roughly every 60–90s, but any provider — current or future, in any language — whose keepalive interval is tighter than 5 minutes is exposed to the same teardown. There is no server-side setting to correct this today.

**Why fix it on the server**

Keepalive *enforcement* is the server's contract: the server is the authority on "how often is too often." Patching each language provider's client keepalive instead would mean coordinated changes across every provider repo (Go, Java/Kotlin, TypeScript, …), wouldn't help third-party or future clients, and would leave the server rejecting perfectly reasonable keepalives by default. One server-side change fixes every client, in every language, at the correct layer.

**Proposal**

Set a streaming-appropriate `KeepaliveEnforcementPolicy` on the flag-sync gRPC server, with sane defaults and CLI/env overrides:

- `--keep-alive-min-time` (env `FLAGD_KEEP_ALIVE_MIN_TIME`, duration, default `30s`) — minimum interval the server permits between client keepalive pings.
- `--keep-alive-permit-without-stream` (env `FLAGD_KEEP_ALIVE_PERMIT_WITHOUT_STREAM`, bool, default `true`) — permit pings when the client has no active stream.

Defaults are strictly **more permissive** than grpc-go's silent 5-minute default, so no currently-working client regresses; clients that are currently GOAWAY'd stop being torn down. Operators who want a tighter or looser policy can tune it without a code change.

**Estate-wide / operator tuning**

Because the policy is exposed as flags/env vars, it is configurable wherever flagd runs — including operator-managed flagd (sidecar or standalone) via container env/args. A single setting on the flagd server tunes keepalive tolerance for *every* consumer connecting to it, across languages and providers, without editing any gRPC client in any workload. (A dedicated field in the OpenFeature Operator CRD is a natural follow-up but not required — env/args already flow through.)

**Scope**

flag-sync gRPC server only. flagd-proxy and the evaluation/metrics servers are out of scope.

**Acceptance**

- Enforcement policy configurable via the two flags/env vars above, with the stated defaults.
- Auto-generated CLI reference updated.
- Unit test for the config→policy mapping; a behavioural test that an aggressively-pinging client is no longer GOAWAY'd.
- ADR added under `docs/architecture-decisions/`.

PR coming for review.

jabenedicic:
## This PR

Sets a streaming-appropriate gRPC keepalive **enforcement policy** on the flag-sync server, configurable via two new flags, so in-process providers stop being disconnected with `GOAWAY ENHANCE_YOUR_CALM` (`too_many_pings`).

Closes #1998.

## Problem

The flag-sync service exposes a long-lived `SyncFlags` gRPC stream that in-process providers keep alive with periodic keepalive pings. The sync server is created with a bare `grpc.NewServer(...)` and no `KeepaliveEnforcementPolicy`, so it inherits grpc-go's defaults — `MinTime: 5m`, `PermitWithoutStream: false` — which are tuned for short unary RPC, not a long-lived, often-idle stream.

Any client pinging more often than every 5 minutes (or pinging during an idle gap with no active stream) accrues ping strikes, and the server responds with `GOAWAY ENHANCE_YOUR_CALM` and closes the connection. Provider SDKs choose their own cadence; the Go provider's in-process resolver pings every 30s with `PermitWithoutStream: true` and no override, so it is disconnected roughly every 60–90s. This is a server-side default mismatch that affects every client language, and there is no server-side setting to correct it today.

## Why fix it on the server

Keepalive *enforcement* is the server's contract — the server decides how often a client may ping. Adjusting each provider's client keepalive instead would need coordinated releases across every provider repo, wouldn't help third-party or future clients, and would leave the server rejecting reasonable keepalives by default. One server-side change fixes every client, in every language, at the correct layer.

## Change

New flags on `flagd start` (env vars auto-derived), applied to the flag-sync gRPC server:

| Flag | Env | Type | Default |
|---|---|---|---|
| `--keep-alive-min-time` | `FLAGD_KEEP_ALIVE_MIN_TIME` | duration | `30s` |
| `--keep-alive-permit-without-stream` | `FLAGD_KEEP_ALIVE_PERMIT_WITHOUT_STREAM` | bool | `true` |

The defaults are strictly **more permissive** than grpc-go's, so no client that works against flagd today regresses; clients that were being disconnected stop being disconnected. Because these are flags/env vars, deployments (including operator-managed flagd via container env/args) can tune keepalive tolerance for every consumer from one place, without editing any gRPC client.

An ADR is included at `docs/architecture-decisions/sync-server-keepalive-enforcement.md`.

## Validation

Reproduced end to end with an in-process provider (Go provider `v0.6.0`, default 30s keepalive) against two builds of the sync server on `:8015`, identical in every other respect:

- **Before** (this repo's `main`): server logs `got too many pings from the client`; client receives `GoAway ... ENHANCE_YOUR_CALM ... "too_many_pings"` at ~105s and the sync stream is torn down and retried.
- **After** (this branch, default `MinTime 30s`): 0 GOAWAYs, 0 sync errors, stream stable for the full 4-minute window (2.4× past the point where `main` failed).

## Test plan

- `make workspace-init && make test` — passes (added a unit test for the config→policy mapping and a raw-HTTP/2 behavioural test that drives the server's enforcement directly: strict policy → `GOAWAY ENHANCE_YOUR_CALM`, permissive policy → healthy, pings-without-stream rejected when not permitted).
- `make lint` — clean on the changed packages.
- `make generate-docs` — regenerated CLI reference is committed.
- `make markdownlint` — clean.

### DIFF
--- a/docs/architecture-decisions/sync-server-keepalive-enforcement.md
+++ b/docs/architecture-decisions/sync-server-keepalive-enforcement.md
@@ -1,0 +1,127 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: proposed
+author: Jason Benedicic (@jabenedicic)
+created: 2026-07-14
+updated: 2026-07-14
+---
+# gRPC keepalive enforcement on the flag-sync server
+
+The flag-sync gRPC server currently sets no keepalive enforcement policy, so it
+inherits grpc-go's defaults, which are tuned for generic request/response RPC
+rather than a long-lived streaming sync. Clients that ping to keep the sync
+stream alive can be disconnected with `GOAWAY ENHANCE_YOUR_CALM`. This ADR
+proposes setting a streaming-appropriate enforcement policy on the server, with
+sane defaults and operator-facing overrides.
+Tracking issue: [#1998](https://github.com/open-feature/flagd/issues/1998).
+
+## Background
+
+flagd's in-process mode has providers consume a long-lived `SyncFlags` gRPC
+stream from the flag-sync server. gRPC clients keep such streams healthy by
+sending periodic keepalive pings, especially during idle periods when no flag
+changes are flowing.
+
+gRPC servers enforce a *keepalive enforcement policy* that decides how often a
+client is allowed to ping. grpc-go's default is `MinTime: 5m` with
+`PermitWithoutStream: false`: a client may ping at most once every five minutes,
+and only while it has an active stream. A client that exceeds this — pings more
+often, or pings during a gap with no active RPC — accrues "ping strikes", and
+after a small number the server sends `GOAWAY` with `ENHANCE_YOUR_CALM` /
+`too_many_pings` and closes the connection. Those defaults suit short unary RPC
+traffic; they are a poor fit for a service whose primary job is a long-lived,
+often-idle stream.
+
+The flag-sync server is constructed with a bare `grpc.NewServer(...)` and no
+enforcement policy, so it inherits those defaults. Provider SDKs choose their
+own keepalive cadence, and several ping far more frequently than every five
+minutes to detect dead connections quickly. The Go provider is the clearest
+example — it pings every 30s with `PermitWithoutStream: true` and exposes no
+override — so it is disconnected roughly every 60–90s and must reconnect. This
+is not specific to one provider: any client, in any language, whose keepalive
+interval is shorter than five minutes is exposed to the same teardown, and there
+is no server-side setting today to widen the tolerance.
+
+## Requirements
+
+* The flag-sync server must tolerate the keepalive cadences that provider SDKs
+  use in practice, without emitting `GOAWAY ENHANCE_YOUR_CALM`.
+* The fix must apply to every client language, not one provider.
+* The policy must be operator-configurable (CLI flag and environment variable),
+  so deployments can tune it without a code change.
+* The default must not regress any client that works against flagd today.
+
+## Considered Options
+
+* Set a fixed, streaming-appropriate enforcement policy on the server, exposed
+  via CLI flags / environment variables (chosen).
+* Set a fixed enforcement policy on the server with no configuration surface.
+* Change each language provider's client keepalive settings instead of the
+  server.
+* Leave the server on grpc-go's defaults (status quo).
+
+## Proposal
+
+Set a `KeepaliveEnforcementPolicy` on the flag-sync gRPC server, configurable
+through two new options with defaults chosen to match how providers actually
+keep the stream alive:
+
+* `--keep-alive-min-time` (env `FLAGD_KEEP_ALIVE_MIN_TIME`, duration, default
+  `30s`) — the minimum interval the server permits between client keepalive
+  pings.
+* `--keep-alive-permit-without-stream` (env
+  `FLAGD_KEEP_ALIVE_PERMIT_WITHOUT_STREAM`, bool, default `true`) — whether the
+  server permits pings while the client holds no active stream.
+
+The `30s` default matches the tightest keepalive cadence a shipped provider uses
+today; because a ping sent every 30s arrives at the server `30s + RTT` apart, it
+clears a 30s minimum with margin. `PermitWithoutStream: true` is required
+because providers ping during idle gaps when, transiently, no stream is active.
+Both defaults are strictly more permissive than grpc-go's `5m` /
+`PermitWithoutStream: false`, so any client that works against flagd today
+continues to work; clients that were being disconnected stop being disconnected.
+
+Choosing the server as the point of control is deliberate. Keepalive enforcement
+is a property of the server — it is the party that decides how often a client
+may ping — so the server is the correct and complete place to fix a
+too-aggressive default. The alternative, changing keepalive in each provider,
+would require coordinated releases across every provider repository, would not
+help third-party or future clients, and would leave the server rejecting
+reasonable keepalives out of the box. One server-side change covers every client
+in every language.
+
+### API changes
+
+Two new CLI flags on `flagd start`, with matching `FLAGD_`-prefixed environment
+variables, documented in the auto-generated CLI reference. No changes to the
+gRPC or flag-configuration APIs. No wire-format or schema changes.
+
+### Consequences
+
+* Good, because a single server-side change resolves `ENHANCE_YOUR_CALM`
+  disconnects for every provider language at once, at the layer that owns the
+  policy.
+* Good, because the defaults are more permissive than grpc-go's, so no
+  currently-working client regresses.
+* Good, because deployments — including operator-managed flagd — can tune the
+  policy through environment variables or arguments on the flagd container,
+  giving estate-wide, cross-language keepalive tuning from one place, without
+  editing any gRPC client in any workload.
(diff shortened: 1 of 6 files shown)


jabenedicic:
Addressed the review feedback in `1492ba8` (tests only):

- **Port-reuse race (CodeRabbit):** each subtest now waits for the sync server goroutine to exit via `t.Cleanup` (cancel context + wait for `Start` to return) before the next subtest reuses the port, removing the listener-shutdown race. Verified with `go test -race -count=5`.
- **Cognitive complexity (SonarCloud):** extracted `dialWithRetry` and `watchForGoAway` helpers from `floodKeepalivePings` to bring it back under the threshold.

Intentionally not changed:
- **Docstring coverage warning:** the added functions already carry doc comments; the reported 40% is dominated by table-driven test closures, and I didn't want to add noise comments to hit an arbitrary threshold on a test-heavy diff.
- **`uint16(port)` narrowing nit:** `port` is a constant `18017`, and this matches the existing pattern in the surrounding sync-service tests.

Note: the `build` workflow is still showing `action_required` — it looks like it needs a maintainer to approve running Actions for a first-time fork contributor. Happy to rebase or adjust anything else.

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -65,26 +68,38 @@ func loadTLSCredentials(certPath string, keyPath string) (credentials.TransportC
 	return credentials.NewTLS(config), nil
 }
 
+// keepAliveEnforcementPolicy derives the gRPC keepalive enforcement policy for
+// the sync server from the service configuration. The sync server serves a
+// long-lived SyncFlags stream that in-process providers keep alive with
+// periodic keepalive pings, so the defaults deliberately widen grpc-go's
+// generic-RPC enforcement (MinTime 5m, PermitWithoutStream false) to tolerate
+// those pings instead of responding with GOAWAY ENHANCE_YOUR_CALM.
```
toddbaert: Nit: this is probably too verbose for an in-line comment, but that's only my opinion.

docs/architecture-decisions/sync-server-keepalive-enforcement.md
```diff
@@ -0,0 +1,127 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: proposed
+author: Jason Benedicic (@jabenedicic)
+created: 2026-07-14
+updated: 2026-07-14
```
toddbaert, line 6: We probably didn't need an ADR - not very controversial - but it doesn't hurt either.

docs/architecture-decisions/sync-server-keepalive-enforcement.md
```diff
@@ -0,0 +1,127 @@
...
+  operators may need to understand.
+* Bad, because a misconfigured very-low `--keep-alive-min-time` could let a
+  misbehaving client ping more aggressively than intended; the default is
+  conservative and this is opt-in.
+
+### Open questions
+
+* Should the OpenFeature Operator expose these as first-class `Flagd` CRD fields,
+  or is passing them through container env/args sufficient? (Follow-up; out of
+  scope for this change.)
```
toddbaert, line 120: Periodically we update OFO's CRDs to have such fields, might be worth an issue.

flagd/pkg/service/flag-sync/sync_service_test.go
```diff
@@ -242,6 +247,193 @@ func TestSyncServiceDeadlineEndToEnd(t *testing.T) {
...
+			wantEnhanceYourCalm: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			port := 18017
+			flagStore, sources := getSimpleFlagStore(t)
+
+			ctx, cancelFunc := context.WithCancel(context.Background())
```
toddbaert, line 313:
```suggestion
			ctx, cancelFunc := context.WithCancel(context.Background())
			defer cancelFunc()
```

Sonar wants you to do this here, instead of in cleanup.

flagd/pkg/service/flag-sync/sync_service.go
jabenedicic: Agreed — condensed to three lines keeping just the GOAWAY rationale in 6721b1a.

flagd/pkg/service/flag-sync/sync_service_test.go
jabenedicic: Done in 5ffd844 — moved the cancel to `defer cancelFunc()` at the creation site. Kept the `<-serverDone` wait in t.Cleanup so the listener is released before the next subtest reuses the port; the deferred cancel runs before Cleanup callbacks, so cancellation still precedes the wait. Passes 3x under -race.

docs/architecture-decisions/sync-server-keepalive-enforcement.md
jabenedicic: Thanks, wasn't sure on the process so kind of winged it a bit. Will know better for future contributions. Was good for me to go through the motions though, made me think a bit more about it.

flagd/pkg/service/flag-sync/sync_service.go
jabenedicic: Fixed

docs/architecture-decisions/sync-server-keepalive-enforcement.md
jabenedicic: Have linked an issue and PR for this over in the other repo.

toddbaert (approved):
Just for completeness, something we've found is that proxies and service meshes _do not_ forward gRPC/HTTP2 keepalive/PING frames - they are a connection level thing, not a RPC-level thing. Due to this, you can have a healthy connection to an Envoy for instance, but that Envoy has dropped it's upstream connection to flagd, and you have no idea because the keepalives are still happily indicating the connection is healthy. 🙃

One of the recommendations we have is to use the `streamDeadlineMs` knob as a way to ensure streams are fresh - since new RPCs transit the intermediate proxy and keep the connection health accurate.

PR #2001, opened 2026-07-15
fix(sync): re-establish file watch after delete and restore

### LINKED ISSUE #1876: [BUG] Feature flags synchronization failure if file with feature flags is removed and restored
### Observed behavior

While trying to evaluate feature flags using `/ofrep/v1/evaluate/flags` endpoint, I encountered an issue related to feature flags synchronization. In standard way of defined flags management, if definition of given flag changes (like its default variant, metadata etc.), appropriate event is being raised and its centralized state is updated.

**Example (clarification for what I'm referring to in the beginning)**
I have defined following flag inside `feature_flags.json` file:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```

When I evaluate flags (i.e. call mentioned endpoint) first, I receive following response:
```
C:\>curl -X POST http://localhost:8016/ofrep/v1/evaluate/flags
{"flags":[{"value":false,"key":"myFeature","reason":"STATIC","variant":"off","metadata":{"description":"Some description"}}],"metadata":{}}
```

After altering feature flag in file, I see its state has changed.
feature_flags.json:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "on", // <- changed from "off" to "on"
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```

Response (flag value changed to "on" variant):
```
C:\>curl -X POST http://localhost:8016/ofrep/v1/evaluate/flags
{"flags":[{"value":true,"key":"myFeature","reason":"STATIC","variant":"on","metadata":{"description":"Some description"}}],"metadata":{}}
```

However, if I remove mentioned feature_flags.json file and restore it back, synchronization between running flagd and feature flag source breaks, making ofrep endpoint response unchanged (assuming You have just one flags definition source)

Before removing and restoring feature_flag.json file back
---------------------------------------------------------------------
feature_flags.json file content:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "on",
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```
Ofrep endpoint response:
```
C:\>curl -X POST http://localhost:8016/ofrep/v1/evaluate/flags
{"flags":[{"value":true,"key":"myFeature","reason":"STATIC","variant":"on","metadata":{"description":"Some description"}}],"metadata":{}}
```

After removing and restoring feature_flag.json file back, I defined 2 feature flags in feature_flag.json file:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		},
		"myFeature2": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```

Ofrep endpoint response:
```
C:\>curl -X POST http://localhost:8016/ofrep/v1/evaluate/flags
{"flags":[{"value":true,"key":"myFeature","reason":"STATIC","variant":"on","metadata":{"description":"Some description"}}],"metadata":{}}
```

Seems like this removing and restoring file procedure, removes a source watcher permanently, so flagd cannot track/keep in sync defined flag state.

### Expected Behavior

Removing and restoring file with feature flags definitions shouldn't break synchronization between flagd and sources. If such situation occurs, file watcher should be rebuilt, making sure that managed flags are always in sync with specified sources

### Steps to reproduce
Prerequisites:
* OS version: Windows 11 Pro, v10.0.26100
* Used flagd version: flagd/v0.13.2 (https://github.com/open-feature/flagd/tree/flagd/v0.13.2)

1. Create "feature_flags.json" file with following content:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```
2. Start flagd process using following command:
`flagd.exe start --debug --uri file:C:\dev\misc\github\feature_flags.json --ofrep-port 8016 --sync-port 8015 --port 8013 --management-port 8014`
3. Send a POST request to `/ofrep/v1/evaluate/flags` endpoint to evaluate defined feature flags
4. Remove created `feature_flags.json` file and restore it back (delete and use ctrl+z right away)
5. Define new feature flag within `feature_flags.json` - see example:
```
{
	"flags": {
		"myFeature": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		},
		"myFeature2": {
			"state": "ENABLED",
			"variants": {
				"on": true,
				"off": false
			},
			"defaultVariant": "off",
			"metadata": {
				"description": "Some description"
			}
		}
	}
}
```
6. Send a POST request to `/ofrep/v1/evaluate/flags` endpoint to evaluate defined feature flags
7. Compare responses received in steps 3 and 6. Response for request sent in step 6 should contain information regarding both `myFeature` and  `myFeature2` flag, whereas response from step 3 should provide information about `myFeature` flag only

mvanhorn:
## Summary

flagd now recovers its file watch when a watched flag-definition file is deleted and quickly restored, instead of silently going stale.

## Why this matters

The reporter of #1876 hit this on Windows by deleting the flag file and immediately undoing (ctrl+z): after the restore, `/ofrep/v1/evaluate/flags` kept returning stale data and later edits to the file were never picked up. A maintainer confirmed the report, labeled it `bug` / `help wanted`, and left it open.

The cause is in `core/pkg/sync/file/filepath_sync.go`. On a `Remove` event the sync loop re-added the watcher with a single `fs.watcher.Add(fs.URI)` call (originally there to handle Kubernetes ConfigMap symlink swaps). During a delete+restore race the file can be momentarily absent at that instant, so `Add` fails, the code logs "error restoring watcher" and gives up. The watch is then lost permanently even though the file exists again a moment later.

## Changes

- Extracted the re-add into a bounded-retry helper, `reAddWatcher(ctx)`, that retries `Add` a few times with a short backoff and aborts early on `ctx` cancellation. A file that returns within the retry window is re-watched. The retry is watcher-agnostic, so it fixes both the `fsnotify` and `fileinfo` paths. The retry bound and backoff are package constants (`reAddMaxAttempts`, `reAddBackoff`).

## Testing

`go test ./pkg/sync/file/` (including `-race`). Added three unit tests against a mock `Watcher`:

- transient failure then success (delete + restore) re-establishes the watch,
- a truly deleted file exhausts the bounded attempts and returns the error (existing default-state behavior preserved),
- a context cancelled during backoff returns promptly instead of running out the attempt budget.

The existing `TestSimpleSync` delete/update cases (real `fsnotify` over a temp dir) stay green, confirming single-delete and ordinary-write behavior is unchanged.

Fixes #1876

### DIFF
--- a/core/pkg/sync/file/filepath_sync.go
+++ b/core/pkg/sync/file/filepath_sync.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	msync "sync"
+	"time"
 
 	"github.com/fsnotify/fsnotify"
 	"github.com/open-feature/flagd/core/pkg/logger"
@@ -18,6 +19,17 @@ import (
 const (
 	FSNOTIFY = "fsnotify"
 	FILEINFO = "fileinfo"
+)
+
+const (
+	// reAddMaxAttempts bounds how many times we try to re-establish the watch
+	// after a remove event before giving up. A delete followed by a quick
+	// restore (e.g. an editor's atomic save, or a manual delete + undo) can
+	// leave the file momentarily absent when the first Add is attempted, so a
+	// single attempt is not enough to recover the watch.
+	reAddMaxAttempts = 5
+	// reAddBackoff is the delay between re-add attempts.
+	reAddBackoff = 100 * time.Millisecond
 )
 
 type Watcher interface {
@@ -114,7 +126,10 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 			case event.Has(fsnotify.Remove):
 				// K8s exposes config maps as symlinks.
 				// Updates cause a remove event, we need to re-add the watcher in this case.
-				err := fs.watcher.Add(fs.URI)
+				// A delete quickly followed by a restore can leave the file
+				// momentarily absent, so retry the re-add for a short window
+				// rather than giving up after a single failed attempt.
+				err := fs.reAddWatcher(ctx)
 				if err != nil {
 					// the watcher could not be re-added, so the file must have been deleted
 					fs.Logger.Error(fmt.Sprintf("error restoring watcher, file may have been deleted: %s", err.Error()))
@@ -147,6 +162,29 @@ func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 			return nil
 		}
 	}
+}
+
+// reAddWatcher re-establishes the watch on fs.URI after a remove event. It
+// retries fs.watcher.Add a bounded number of times with a short backoff so a
+// file that is deleted and quickly restored is picked back up instead of
+// silently losing its watch. It returns the last Add error if every attempt
+// fails, or ctx.Err() if the context is cancelled while backing off.
+func (fs *Sync) reAddWatcher(ctx context.Context) error {
+	var err error
+	for attempt := 0; attempt < reAddMaxAttempts; attempt++ {
+		if err = fs.watcher.Add(fs.URI); err == nil {
+			return nil
+		}
+		if attempt == reAddMaxAttempts-1 {
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(reAddBackoff):
+		}
+	}
+	return err
 }
 
 func (fs *Sync) sendDataSync(ctx context.Context, dataSync chan<- sync.DataSync) {
--- a/core/pkg/sync/file/filepath_sync.go
+++ b/core/pkg/sync/file/filepath_sync.go
@@ -22,14 +22,10 @@ const (
 )
 
 const (
-	// reAddMaxAttempts bounds how many times we try to re-establish the watch
-	// after a remove event before giving up. A delete followed by a quick
-	// restore (e.g. an editor's atomic save, or a manual delete + undo) can
-	// leave the file momentarily absent when the first Add is attempted, so a
-	// single attempt is not enough to recover the watch.
-	reAddMaxAttempts = 5
-	// reAddBackoff is the delay between re-add attempts.
-	reAddBackoff = 100 * time.Millisecond
+	// bounded re-add attempts after a remove event; 6 x 200ms backoff = ~1s window to survive a non-atomic replace (delete then recreate)
+	reAddMaxAttempts = 6
+	// delay between re-add attempts
+	reAddBackoff = 200 * time.Millisecond
 )
 
 type Watcher interface {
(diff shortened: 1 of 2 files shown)


core/pkg/sync/file/filepath_sync.go
```diff
@@ -20,6 +21,17 @@ const (
...
 
+const (
+	// reAddMaxAttempts bounds how many times we try to re-establish the watch
+	// after a remove event before giving up. A delete followed by a quick
+	// restore (e.g. an editor's atomic save, or a manual delete + undo) can
+	// leave the file momentarily absent when the first Add is attempted, so a
+	// single attempt is not enough to recover the watch.
+	reAddMaxAttempts = 5
+	// reAddBackoff is the delay between re-add attempts.
+	reAddBackoff = 100 * time.Millisecond
```
leakonvalinka: Not sure if this is enough time to catch manual restores after accidental deletes? 
I wonder if we should make it configurable too, but that's probably overkill for an edge case.

leakonvalinka (approved):
Looks good! I left a question but that's not a blocker for me.

core/pkg/sync/file/filepath_sync.go
toddbaert:
```suggestion
	reAddMaxAttempts = 6
	// reAddBackoff is the delay between re-add attempts.
	reAddBackoff = 200 * time.Millisecond
```

I think we should change this to `200`, x 6, which with the logic below equals 1s total, which seems a bit less arbitrary than the `400ms` total we get now.

toddbaert: Pushed this.

toddbaert:
Hey @mvanhorn ... I think this is a good change overall. I shortened some particularly wordy comments and tweaked your constants so that we re-attach for up to 1s [here](https://github.com/open-feature/flagd/pull/2001/changes/ed4c05de08148c6cee147d7306cbdc4056cfe5ec).

Thanks!

core/pkg/sync/file/filepath_sync.go
toddbaert: Ya, I had the same thought. I think we are less concerned with manual delete/add-back, and mostly concerned with automations/scripts and editors, some non-atomic mount/volume update mechanisms (plain hostPath scripts, certain network filesystems) that don't use K8s-style atomic swaps... But for the sake of picking a nice round number, I've made it 1s.

toddbaert (approved):

mvanhorn:
Thanks for landing this, @toddbaert - the file watch surviving delete-and-restore closes a real sync gap.

PR #2012, opened 2026-08-04
feat: SSE support for OFREP endpoints

JamieSinn:
## This PR
- Adds an SSE server compliant with ADR-0008 for OFREP https://github.com/open-feature/protocol/pull/63
- Implement the ADR-0008 OFREP response body

The implementation of the wire protocol/message format is the bare SSE object that was defined in ADR-0008, not a custom protocol.
The SSE server is using the LaunchDarkly EventSource server which we've proven in DevCycle/Dynatrace that it's quite stable and has the ideal functionality that we want (channels, separation, and highly performant).
### Related Issues

https://github.com/open-feature/protocol/pull/63

### Notes
The URL path for this `/ofrep/v1/sse` is very open to discussion - I have no strong opinions on this, but this just felt the most logical.

### How to test

This was tested manually; I can add a script or something to the test folder if that's easier.


flagd/pkg/service/flag-evaluation/ofrep/sse/service.go
```diff
@@ -0,0 +1,91 @@
...
+	"time"
+
+	"github.com/launchdarkly/eventsource"
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/store"
+)
+
+// defaultHeartbeatInterval must stay well below the advertised inactivity delay so idle
+// connections are not closed by clients or proxies.
+const defaultHeartbeatInterval = 30 * time.Second
```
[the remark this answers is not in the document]
JamieSinn: Inactivity delay is for clients to keep the connection open in the background - not the connection inactivity/keepalive.

core/pkg/service/ofrep/models.go
```diff
@@ -22,6 +22,24 @@ type EvaluationSuccess struct {
...
+	// EventStreams advertises SSE endpoints clients can subscribe to for change
+	// notifications, per OpenFeature protocol ADR-0008. Omitted when SSE is disabled.
+	EventStreams []EventStream `json:"eventStreams,omitempty"`
+}
+
+// EventStream describes a Server-Sent Events endpoint a client can subscribe to in order to
+// be notified (via a `refetchEvaluation` event) when the flag configuration changes.
+type EventStream struct {
+	Type               string               `json:"type"`
+	Endpoint           *EventStreamEndpoint `json:"endpoint"`
```
leakonvalinka, line 34: What about the `url` option? I saw in it in the adr and if I understood the discussion correctly, it should even be default? Or has that changed

JamieSinn: I actually wrote it with the URL method the first time I wrote this out, but then I realized we have a small issue of that we don't know the required path/host path that clients use to reach the OFREP endpoint as people regularly put flagd behind a proxy, SSL termination, etc.

The endpoint method allows us to use the reliance on the OFREP client to fall back to the same host as the OFREP request is being made to for the SSE endpoint. This means we don't need to configure an endpoint/host on the flagd instance. 

Though - I did add a configuration property to allow configuring the hostname if you know it, or want to use a different host/route path.

toddbaert:
I think I see an issue here...

The bulk ETag here is derived from the flag config, but the `/ofrep/v1/evaluate/flags` response contains evaluated values that also depend on the eval context; so flagd can return a `304` for a request whose context changed and would resolve differently. The OFREP spec (https://github.com/open-feature/protocol/blob/main/service/openapi.yaml) doesn't actually specify this: **_it defines the bulk endpoint as "static context" and defines the ETag as "the current state of all flags," but never states whether the `RFC 7232` "cache-validator" covers context, or alternatively, that clients must drop their ETag on a context change._**

That ambiguity is already possibly a problem: the reference [ofrep-web-provider](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts#L186) only clears its cached ETag when `targetingKey` changes, so any other context attribute change (email, country, plan, etc.) sends a stale `If-None-Match` and gets served old values. IMO we should pin down the semantics in the [protocol repo](https://github.com/open-feature/protocol) first, then decide whether flagd folds a context hash into the ETag or explicitly documents this as config-scoped and relies on the client to dump its ETag.

cc @jonathannorris

JamieSinn:
> I think I see an issue here...
> 
> The bulk ETag here is derived from the flag config, but the `/ofrep/v1/evaluate/flags` response contains evaluated values that also depend on the eval context; so flagd can return a `304` for a request whose context changed and would resolve differently. The OFREP spec (https://github.com/open-feature/protocol/blob/main/service/openapi.yaml) doesn't actually specify this: **_it defines the bulk endpoint as "static context" and defines the ETag as "the current state of all flags," but never states whether the `RFC 7232` "cache-validator" covers context, or alternatively, that clients must drop their ETag on a context change._**
> 
> That ambiguity is already possibly a problem: the reference [ofrep-web-provider](https://github.com/open-feature/js-sdk-contrib/blob/main/libs/providers/ofrep-web/src/lib/ofrep-web-provider.ts#L186) only clears its cached ETag when `targetingKey` changes, so any other context attribute change (email, country, plan, etc.) sends a stale `If-None-Match` and gets served old values. IMO we should pin down the semantics in the [protocol repo](https://github.com/open-feature/protocol) first, then decide whether flagd folds a context hash into the ETag or explicitly documents this as config-scoped and relies on the client to dump its ETag.


Definitely, this is also the case for any changes that happen as a result of the evaluation itself (metadata, SSE URL changing, or even any other top level field that is in the response).

This is something that should be separately addressed in both Flagd and OFREP, happy to make the issues for both to document and track. I'll setup a separate PR to fix the ETag calculation to include the entire response body of the OFREP response in the ETag spec - which is more in line with how the concept of the ETag is intended (to cover the entire response and help CDN's cache data - https://httpwg.org/specs/rfc9110.html#field.etag)

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,270 @@
...
+	copy(sorted, flags)
+	sort.Slice(sorted, func(i, j int) bool {
+		if sorted[i].FlagSetId != sorted[j].FlagSetId {
+			return sorted[i].FlagSetId < sorted[j].FlagSetId
+		}
+		if sorted[i].Source != sorted[j].Source {
+			return sorted[i].Source < sorted[j].Source
+		}
+		return sorted[i].Key < sorted[j].Key
+	})
```
toddbaert: There's some double duty happening here - finderprinting is fine, but this is also _filtering_ by flagSetId - something that the storage/query-layer actually does for you much more efficiently using immutable radix trees; basically you can pass the selector into `store.Watch` and it will only emit events when the set of flags selected is impacted. In the future we might support selectors like "all boolean flags" etc - anything we can query/index - so we need to lean on that here.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,270 @@
...
+func (t *Tracker) Run(ctx context.Context) {
+	if t.store == nil {
+		if t.logger != nil {
+			t.logger.Warn("ofrep sse tracker has no flag store; change notifications disabled")
+		}
+		return
+	}
+
+	watcher := make(chan store.FlagQueryResult, 1)
+	t.store.Watch(ctx, &store.Selector{}, watcher)
```
toddbaert: Pass a selector from the request (the channel might have to be a selector expression and not a flagSetId purely.

toddbaert (changes requested):
The store's `Watch` already selector-scopes via `ToQuery` (including compound indices), so the tracker should watch per-channel, with that channel's _real selector_ rather than one "empty-selector" watch plus grouping after the fact. That single change:

- lets channels be arbitrary selectors
- removes the whole-store re-fingerprint on every change and manual filtering
- makes the ETag scope exactly match the eval scope "for free"

This is how the [EventStream](https://github.com/open-feature/flagd/blob/main/flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go#L79) currently works, which is basically a gRPC equivalent of the SSE you are writing. You can basically power your SSE entirely with the same logic.

flagd/pkg/service/flag-evaluation/ofrep/ofrep_service.go
```diff
@@ -9,11 +9,16 @@ import (
...
 	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/store"
 	"github.com/open-feature/flagd/core/pkg/telemetry"
+	"github.com/open-feature/flagd/flagd/pkg/service/flag-evaluation/ofrep/sse"
 	corsmw "github.com/open-feature/flagd/flagd/pkg/service/middleware/cors"
 	"golang.org/x/sync/errgroup"
 )
 
+// ssePath is the endpoint OFREP clients subscribe to for change notifications (ADR-0008).
+const ssePath = "/ofrep/v1/sse"
```
erka:
```suggestion
const ssePath = "GET /ofrep/v1/sse/{channel}"
```

flagd/pkg/service/flag-evaluation/ofrep/ofrep_service.go
```diff
@@ -26,30 +31,68 @@ type SvcConfiguration struct {
...
+	// the request-body limit is not applied to the stream. Everything else falls through to the
+	// existing OFREP handler.
+	mux := http.NewServeMux()
+	var evaluateHandler http.Handler = ofrepHandler
 	if cfg.MaxRequestBodyBytes > 0 {
-		h = http.MaxBytesHandler(h, cfg.MaxRequestBodyBytes)
+		evaluateHandler = http.MaxBytesHandler(evaluateHandler, cfg.MaxRequestBodyBytes)
+	}
+	if sseService != nil {
+		mux.Handle(ssePath, sseService.Handler())
```
erka: It works but it doesn't look modern.  It may be better to have `GET /ofrep/v1/sse/{channel}` and use `r.PathValue("channel")` in the handler.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,238 @@
...
+
+// Close stops accepting subscriptions and waits for every watch goroutine to exit. It must
+// complete before the eventsource server is closed, since a publish after that blocks forever.
+func (t *Tracker) Close() {
+	t.mu.Lock()
+	t.closed = true
+	subs := make([]*subscription, 0, len(t.subs))
+	for _, sub := range t.subs {
+		subs = append(subs, sub)
+	}
```
erka:
```suggestion
	subs := t.subs
```
I don't think we need an extra copy

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,238 @@
...
+func (t *Tracker) publish(channel string, sub *subscription) {
+	v := sub.ver.Load()
+	if v == nil {
+		return
+	}
+
+	id := strconv.FormatInt(t.eventID.Add(1), 10)
+	t.es.Publish([]string{channel}, newRefetchEvent(id, v.etag, v.lastModified))
+	if t.logger != nil {
+		t.logger.Debug(fmt.Sprintf("published refetch event to channel %q (etag=%s)", channel, v.etag))
```
erka:
```suggestion
		t.logger.Debug("published refetch event", zap.String("channel", channel), zap.String("etag", v.etag)
```

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,238 @@
...
+	if ctx.Err() != nil {
+		return // our own teardown
+	}
+
+	// The store hit a selector/iterator error, so this channel will never fire again. Stop
+	// serving its ETag, which would otherwise answer 304 from a frozen fingerprint forever.
+	sub.ver.Store(nil)
+	if t.logger != nil {
+		t.logger.Error(fmt.Sprintf(
+			"ofrep sse watch for channel %q ended unexpectedly; its cached ETag is invalidated", channel))
```
erka:
```suggestion
		t.logger.Error("ofrep sse watch ended unexpectedly; its cached ETag is invalidated", zap.String("channel", channel))
````

File: flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go [hunk 10]
```diff
@@ -0,0 +1,238 @@
...
+
+// publish must be called only after the new version is stored, so a client that refetches the
+// instant it receives the event cannot read the previous ETag and be served a stale 304.
+func (t *Tracker) publish(channel string, sub *subscription) {
+	v := sub.ver.Load()
+	if v == nil {
+		return
+	}
+
+	id := strconv.FormatInt(t.eventID.Add(1), 10)
```
erka: It's probably only make sense for one immortal server. Any restart or replicas will break this sequence. It would be great to find another way to generate a stable event id.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,238 @@
...
+	// This loop must run until the channel closes: store.Watch's send does not select on
+	// ctx.Done(), so abandoning the channel would leak the store's goroutine.
+	for res := range watcher {
+		fp := fingerprint(res.Flags)
+		if prev := sub.ver.Load(); prev != nil && prev.etag == fp {
+			// A no-op wakeup: an identical re-sync, or a coarser radix watch channel firing
+			// for a change outside this selector.
+			continue
+		}
+		sub.ver.Store(&version{etag: fp, lastModified: time.Now().Unix()})
```
erka: is this a race? `CompareAndSwap` probably is the method for this.

erka (commented):
Great work! I just have a few nits.

Also I have a question about metrics. Do we want to keep track of the number of subscribers?

flagd/pkg/service/flag-evaluation/ofrep/ofrep_service.go
JamieSinn: So the gotcha by including it in the route path directly is that would violate the ADR-0008 around logging of the channel/subscribe tokens (#7 here - https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md#open-questions)

Having the channel be a query param allows it to be filtered more easily/readily for this.


I do agree though, having it in the path is much prettier.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: The sequence count is mostly irrelevant, the only benefit this provides is the possibility of a "replay X events since ID Y" situation on reconnection which doesn't really make sense for the concept of SSE. (Ably implements this, but it's not something super useful for us in the current design for SSE referring to config refetching). 

Realistically, it doesn't even need to be numeric, it could be a hash of the current timestamp. Thoughts on that?

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: The intent here is to isolate the current list, and process that separately from the mutex'd path here so that the channel can close faster. I guess technically this is an over optimization, but i was treating it as a queue drain in isolation so that closing isn't blocked on this queue having a race condition/concurrency fight for it. 

If you think it's fine, happy to just do this in a single drain for shutdown.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: Better safe than sorry, will swap.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
erka: The next line does `t.subs = map[string]*subscription{}`, so I believe it's extra memory allocation with that loop.

JamieSinn: it is, but during shutdown (when this is intended) i wasn't too worried about it given it's only doubling the number of connections. Happy to swap this to a blocking shutdown if you prefer.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: actually, sequence wise the current epoch millis is a completely fine message id. Will swap that.

flagd/pkg/service/flag-evaluation/ofrep/ofrep_service.go
erka: how does request query change that? I believe it will be logged as well by most http servers

JamieSinn: true, functionally it's mostly for the upstream observability tooling that ingests the request logs. 

curious  - @toddbaert any opinions on path vs query param?

toddbaert: I don't think this is an issue for us - 7 says:

> Providers must not log or persist SSE URLs 

Emphasis on _providers_. A provider has no idea if a SSE URL is sensitive - the SERVER does. flagd purposely relies on other "cloud-native" components to do auth (think Istio, etc) - the channel here would just be a selector, which is not secret or sensitive.

JamieSinn: pretty path it is then!

toddbaert: Just make sure you do any required encoding/decoding - selectors may contain more chars in the future.

JamieSinn:
> Also I have a question about metrics. Do we want to keep track of the number of subscribers?

Conceptually I don't think these are super valuable metrics here; I think it might be better to move the metrics idea/discussion to a larger initiative that accounts for more than just the evaluation flow.

toddbaert:
> > Also I have a question about metrics. Do we want to keep track of the number of subscribers?
> 
> Conceptually I don't think these are super valuable metrics here; I think it might be better to move the metrics idea/discussion to a larger initiative that accounts for more than just the evaluation flow.

Ya this is a new custom metric that deserves it's own issue. I think we have similar cases.

flagd/pkg/service/flag-evaluation/ofrep/sse/handler.go
```diff
@@ -0,0 +1,39 @@
...
+		// already percent-decoded by the mux, so a source containing "/" arrives intact
+		channel := r.PathValue(channelPathVar)
+		selector, err := store.NewSelector(channel)
+		if err != nil {
+			// not echoing the expression back: it is unescaped client input
+			http.Error(w, "invalid selector in the channel path segment", http.StatusBadRequest)
+			return
+		}
+
+		release, err := svc.tracker.Subscribe(channel, selector)
```
toddbaert, line 30: Nit: since `channel` is keyed by the raw selector string, equivalent expressions like `myflags.json` and `source=myflags.json` open two separate watches for the same flags; keying subscriptions by the parsed selector (e.g. `selector.ToLogString()`) would dedupe them.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/launchdarkly/eventsource"
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/model"
+	"github.com/open-feature/flagd/core/pkg/store"
+	"go.uber.org/zap"
+)
+
+type version struct {
+	etag         string
```
toddbaert: Nit: since this token is config-scoped and distinct from the HTTP representation `ETag` (it's really the spec's `flagConfigEtag`), consider naming it `configEtag` internally ; `version.etag`, `Version()`'s return, etc. They happen to be the same value today, but once a context-aware validator is implemented for the `ETag` header they'll diverge, and the name will otherwise imply they're the same thing. The SSE JSON wire field (`event.go:13`, `json:"etag"`) should stay as-is to match the OFREP spec.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
```diff
@@ -0,0 +1,229 @@
...
+// fingerprint produces a deterministic hash of the result set a channel's watch resolves to.
+func fingerprint(selector *store.Selector, flags []model.Flag) string {
+	digests := make([]string, 0, len(flags))
+	for _, f := range flags {
+		h := sha256.New()
+		h.Write([]byte(f.Key))
+		h.Write([]byte{0})
+		h.Write([]byte(f.Source))
+		h.Write([]byte{0})
+		b, _ := json.Marshal(f)
```
toddbaert, line 211: Non-blocking: `fingerprint` is a second definition of "flag changed," parallel to the RPC path's [`NewFromFlags`](https://github.com/open-feature/flagd/blob/main/core/pkg/notifications/notifications.go#L43)/`flagsEqual` ; a denylist (`Flag.MarshalJSON`) vs an allowlist. Fine today, but they'll drift. Since both are the de-dup gate that hits notifications, not just payloads. IMO both should sit on one equality primitive with a guard test. Happy to fast-follow.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
toddbaert: > isolate the current list, and process that separately from the mutex'd path here so that the channel can close faster.

I think that's already achieved by the `t.subs = map[...]{}` reassignment on the next line; the map swap already isolates the old map, so the slice copy is redundant if I'm reading correctly. I would take @erka 's change I think unless I'm missing something.

toddbaert (approved):
Approving with 3 nits, and and I think @erka might be right about [this](https://github.com/open-feature/flagd/pull/2012#discussion_r3846424670), because we are already copying the memory in question.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: I wanted to adopt the RPC path, but given this is a bit more one off I didn't feel like it was super applicable to do all the plumbing that would be needed for just SSE to use it. Happy to make this a fast follow later.

flagd/pkg/service/flag-evaluation/ofrep/sse/tracker.go
JamieSinn: good idea - will clarify

flagd/pkg/service/flag-evaluation/ofrep/sse/handler.go
JamieSinn: good catch - will fix

toddbaert:
@erka I think @JamieSinn applied your suggestion.

I'll merge this today unless I hear objections.

flagd/pkg/service/flag-evaluation/ofrep/sse/handler.go
toddbaert: My dedupe suggestion caused the SSE test failures ; keying the sub by `ToLogString()` (which was built for log formatting, not channel identity) while the client still binds `es.Handler(channel)` to the raw path lands publishes and subscribers on different channels. I'll revert that; let's defer the dedupe to a proper follow-up. (reverted in the latest fixup)

erka (approved):

PR #2013, opened 2026-08-05
docs: add adr for targeting metadata

suthar26:
## This PR

- adds a new ADR for targeting metadata in flagd schema

### DIFF
--- a/docs/architecture-decisions/targeting-variant-metadata.md
+++ b/docs/architecture-decisions/targeting-variant-metadata.md
@@ -1,0 +1,255 @@
+---
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: proposed
+author: Parth Suthar
+created: 2026-07-30
+updated: 2026-09-02
+---
+
+# Per-evaluation metadata from targeting rules
+
+Let a targeting rule return `{ "variant": "<key>", "reason": "<string>", "metadata": { ... } }` in place of a plain variant string, so the _branch that fired_ can annotate the evaluation with extra metadata, custom reasons, and other auxiliary information. Existing string returns are unchanged.
+
+## Background
+
+An evaluation today returns `value`, `variant`, `reason`, and `metadata`. The `reason` is a coarse enum (`TARGETING_MATCH`, `DEFAULT`, `STATIC`, …); `metadata` carries only the static blocks defined at flag-set and flag level. Neither answers the debugging question we hit most often: **which branch of the targeting expression fired?**
+
+For a nested `if` / `and` / `or` tree the resolver just returns the winning variant key. Two rules landing on the same variant are indistinguishable in the response — the only way to know _why_ today is to fetch the flag config and re-run the logic by hand. Encoding the branch identity into the variant key (`"clubs-eu-rollout-a"`) is the workaround, and it pollutes the variant space with debug info consumers then have to parse back out.
+
+The plumbing to carry metadata already exists on most paths: `AnyValue.Metadata` (`core/pkg/evaluator/ievaluator.go`) is threaded through the evaluator and surfaced via OFREP (single and bulk) and single-resolution gRPC responses as `flagMetadata`. Two paths don't carry it yet — bulk gRPC `ResolveAll` and `RecordEvaluation` telemetry — but wiring those up is separate follow-up work.
+
+## Proposal
+
+In `definitions.primitive` in `schemas/json/targeting.json`, the `string` entry (the variant-key return) becomes a choice of the plain string or the tagged object — the other `primitive` entries (`null`, `boolean`, `number`, `array`) are untouched:
+
+```json
+{
+  "description": "When returned from rules, strings are used as keys to retrieve the associated value from the \"variants\" object. Be sure that the returned string is present as a key in the variants! As of <version>, an object of the form { \"variant\": ..., \"reason\": ..., \"metadata\": ... } may be used instead, to attach a custom reason and arbitrary metadata to the branch that fired.",
+  "oneOf": [
+    { "type": "string" },
+    {
+      "type": "object",
+      "required": ["variant"],
+      "additionalProperties": false,
+      "properties": {
+        "variant": { "type": "string" },
+        "reason": { "type": "string" },
+        "metadata": {
+          "allOf": [
+            {
+              "$ref": "https://flagd.dev/schema/v0/flags.json#/definitions/metadata"
+            },
+            {
+              "properties": {
+                "details": {
+                  "type": "string",
+                  "description": "Free-text explanation of why this branch fired."
+                }
+              }
+            }
+          ]
+        }
+      }
+    }
+  ]
+}
+```
+
+### Field types
+
+- **`variant`** — string. The only **required** property, and the exact equivalent of the plain-string return: `"red"` and `{ "variant": "red" }` are interchangeable terminals selecting the variant `red`.
+- **`reason`** — string, optional. Free-form, operator-defined; overrides the top-level evaluation `reason` for this branch. The schema does not constrain it to an enum (see [Narrowing the coarse `reason`](#narrowing-the-coarse-reason)).
+- **`metadata`** — object, optional, shallow. Reuses the existing `metadata` definition, so its values stay restricted to `string | number | boolean`, matching flag and flag-set metadata. Nested objects and arrays are not permitted.
+- **`metadata.details`** — string, optional. A well-known key, declared in the schema so editors autocomplete and validate it, but carried inside `metadata` rather than promoted to a peer of `variant`.
+
+### Evaluator changes
+
+In `evaluateVariant` (`core/pkg/evaluator/json.go`), the current unconditional quote-strip of the JsonLogic result gains an object case ahead of it:
+
+- object with a `variant` field → use it as the variant key; shallow-merge the rule's `metadata` object into the returned metadata; if `reason` is present, use it in place of `TARGETING_MATCH`
+- everything else → unchanged from today: `null` exits targeting to `defaultVariant`; strings, booleans, and numbers are quote-stripped into a variant key (so a boolean `true` still resolves the `"true"` variant, as `schemas/json/targeting.json` documents and the existing tests cover); a key absent from `variants` errors exactly as it does now
+
+Metadata merge precedence, lowest → highest, so more specific wins: flag-set metadata → flag metadata → **rule-returned metadata**.
+
+!!! warning "Breaking change to the config schema"
+
+    Existing configs keep working on a new binary, but the reverse does not hold: a config using the tagged-object return, evaluated by an older flagd binary, fails — the old resolver quote-strips the JsonLogic result and gets a non-existent variant key (or panics on the single-key `variant` map, which it has no operator registered for).
+
+    There's no schema/feature version negotiation today, so the config schema version has to gate it: operators on an older binary must stay on the older schema version, and flagd has to be upgraded fleet-wide before any config starts emitting the new shape. This needs a schema version bump and a migration note in the release.
+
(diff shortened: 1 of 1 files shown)


aepfli:
I generally like the idea, but I am missing the migration path. How do we plan to move from old version to new version, without breaking migration patterns?

I feel like we need to point out how we plan to migrate without breaking installations for others. In-process providers might be used in deprecated services and we need to give people time to migrate from one to the other.

Is it maybe time to have a version contract in the request for the flag configuration?

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+# Valid statuses: draft | proposed | rejected | accepted | superseded
+status: proposed
+author: Parth Suthar
+created: 2026-07-30
+updated: 2026-07-30
+---
+
+# Per-evaluation metadata from targeting rules
+
+Let a targeting rule return `{ "variant": "<key>", "reason": "<enum>", "details": "<text>", "metadata": { ... } }` in place of a plain variant string, so the _branch that fired_ can annotate the evaluation with extra information. Existing string returns are unchanged.
```
toddbaert:
```suggestion
Let a targeting rule return `{ "variant": "<key>", "reason": "<enum>", "details": "<text>", "metadata": { ... } }` in place of a plain variant string, so the _branch that fired_ can annotate the evaluation with extra metadata, custom reasons, and other auxiliary information. Existing string returns are unchanged.
```

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+        "metadata": {
+          "$ref": "https://flagd.dev/schema/v0/flags.json#/definitions/metadata"
+        }
+      }
+    }
+  ]
+}
+```
+
+`reason` and `details` are optional plain strings — `reason` is an operator-chosen enum value (e.g. `SPLIT_GRADUAL`, `ALLOWLIST`, `OVERRIDE`) that classifies _why_ this branch fired, and `details` is a longer free-text explanation. They're first-class schema-validated properties rather than an ad-hoc convention in freeform `metadata`, so consumers get stable key names instead of every config author inventing their own (`rule`, `ruleId`, `why`, …).
```
toddbaert: Reasons _could_ be totally free form. Is there any reason to constrain them? (enum to me implies they are limited). If we want we COULD limit them in the schema - they should at least be limited to strings, but it could be an enum.

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+The `metadata` property reuses the existing definition, so its values stay restricted to `string | number | boolean`, matching flag and flag-set metadata.
+
+In `evaluateVariant` (`core/pkg/evaluator/json.go`), replace the current string-strip of the JsonLogic result with a typed decode:
+
+- string → variant key, as today
+- object with `variant` field → use it as the variant key; write `reason`/`details` (if present) into the returned metadata, then shallow-merge the rule's `metadata` object on top. Top-level `reason`/`details` win if the rule's `metadata` also sets keys of those names.
```
toddbaert: To confirm, then, `variant` is the only REQUIRED attribute in the new object format, yes?

We should probably make that explicit. I don't think any of the others are needed.. basically:

```json
"red"
```

and 

```json
{
  "variant": "red"
}
```

Would be exactly equivalent terminals for selecting a variant called "red".

toddbaert: Maybe I'm misunderstanding, but I would make `reason` a peer of `variant` because that's how OpenFeature describes them in various situations. This makes the "object terminals" look very similar to OpenFeature's evaluation/resolution objects, which is very intuitive. Here's a real, full example: 


```jsonc
{
  "flags": {
    "enable-mainframe-access": {
      "state": "ENABLED",
      "defaultVariant": "off",
      "variants": {
        "on": true,
        "off": false
      },
      "targeting": {
        "if": [
          {
            "ends_with": [
              {
                "var": "email"
              },
              "@ingen.com"
            ]
          },
          // new terminus (this node used to be "on"
          {
            "variant": "on", // required
            "reason": "my_super_cool_custom_reason", // optional
            "metadata": {} // custom metadata
           }
        ]
      }
    }
  }
}
```

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+- anything else → `PARSE_ERROR`, same as an unrecognized return today
+
+Merge precedence, lowest → highest, so more specific wins: flag-set metadata → flag metadata → **rule-returned metadata**.
+
+!!! warning "Breaking change to the config schema"
+
+    Existing configs keep working on a new binary, but the reverse does not hold: a config using the tagged-object return, evaluated by an older flagd binary, fails — the old resolver quote-strips the JsonLogic result and gets a non-existent variant key (or panics on the single-key `variant` map, which it has no operator registered for).
+
+    There's no schema/feature version negotiation today, so the config schema version has to gate it: operators on an older binary must stay on the older schema version, and flagd has to be upgraded fleet-wide before any config starts emitting the new shape. This needs a schema version bump and a migration note in the release.
```
toddbaert, line 78: Yep - we've done this before.

For me this is a sufficient migration plan (just gate via json schema) - existing releases will see these configs as errors; the providers and flagd embed a snapshot of the schema in each version so they will complain if somebody tried to use this new feature with them.

The only thing we might want to do is wait until we have it implemented in a few places before we update the HOSTED flagd schema used in editors, etc at https://flagd.dev/schema/v0/flags.json 

cc @aepfli

docs/architecture-decisions/targeting-variant-metadata.md
toddbaert: Oh sorry - this is what you describe below - but reading this I thought you were proposing putting `reason` under `metadata`, but your examples below look the same as mine.

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+                  ["off", 90]
+                ]
+              },
+              "on"
+            ]
+          },
+          {
+            "variant": "on",
+            "reason": "SPLIT_RANDOM",
+            "detail": "us-10pct-rollout"
```
toddbaert: This sort of "promotes" detail to the same level as `variant` and `value`. I wonder if it's better to put it into the metadata, instead of adding this new "first class" field. I'm not 100% sure though. WDYT?

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+
+Note that the OpenFeature spec ([evaluation details, requirement 6.1](https://openfeature.dev/specification/sections/evaluation-context)) types `reason` as a free-form string — the values in the spec (`TARGETING_MATCH`, `SPLIT`, `DEFAULT`, …) are _recommended_, not exhaustive, and providers are explicitly allowed to emit their own.
+That means letting a branch-scoped `reason` override the top-level `reason` is spec-legal without any SDK contract change; SDK hooks and telemetry sinks already treat the field as opaque.
+
+The user facing reason and mechanism does not match what the falgd engine is evaluating for differnet variants of `SPLIT`. Today split covers both fractional and gradual rollout. Adding a reason override for from the engine result allows for an easy win in this case. Also there are differnet type of `OVERRIDES` which can benefit from this.
+
+```json
+{ "variant": "on", "reason": "SPLIT_GRADUAL", "details": "5%→50% ramp, week 3" }
+{ "variant": "on", "reason": "SPLIT_STEPPED", "details": "stage 2 of 4, cohort=eu-west" }
+{ "variant": "on", "reason": "SPLIT_RANDOM",  "details": "50/50 A/B, salt=exp-4231" }
```
toddbaert: Related to https://github.com/open-feature/flagd/pull/2013/changes#r3883050353... 

I think even if we put `details` here in the flagd schema, it would have to go in metadata in the actual openfeature evaluation (I don't think we want to add this as a field in every flag resolution in all our OpenFeature SDKS, that would be a huge spec change)... so it might be better to put the `details` on metadata from start, directly in the config. We could define it in the schema so it would autocomplete, and be validated (I guess it needs to be a string?).

toddbaert (approved):
I love this idea, and I think it's super useful/powerful and relatively easy to implement.

I left some significant comments, but they all seem resolvable and relatively minor, so I approve.

One other nit would be to explicitly mention the data-type of all these new fields (`variant` has to be a string, for example, `details` as well, I guess), metadata an object (shallow)...

toddbaert:
Oh, another thought, it might be worth a small section mentioning that we'd enhance the flagd gherkin suite to test this as well.

docs/architecture-decisions/targeting-variant-metadata.md
JamieSinn: I agree - why not make them just arbitrary strings that the operator can define?

docs/architecture-decisions/targeting-variant-metadata.md
JamieSinn: I'd prefer metadata too, personally.

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+## Consequences
+
+- **Good** — direct answer to "which branch fired" without overloading the variant key; no wire-format changes. `flagMetadata` is already emitted on OpenFeature evaluation events and picked up by SDK hooks (OpenTelemetry, logging, custom exporters), so anything reading it — OTel spans, A/B dashboards, debug logs — picks up rule-level attribution automatically.
+- **Good** — backwards compatible for existing configs: string returns are byte-for-byte unchanged on an upgraded flagd binary.
+- **Bad** — not forward compatible: a config using the new object-return shape, evaluated by an older flagd binary, fails (the returned object doesn't quote-strip into a valid variant key). No schema/feature version negotiation exists today, so rollouts must upgrade flagd before configs start using the new shape. Needs an explicit migration note.
+- **Bad** — automatic telemetry attribution requires the separate `RecordEvaluation` work called out in Background; without it, rule ids ride `flagMetadata` for SDK-side hooks but don't reach flagd's own OTel metrics.
+
+## Open questions
+
+- **Config Version.** How would we want to version the flagd schema?
```
JamieSinn: v0.1?

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+- **Fractional buckets.** `fractional` returns a variant string directly, so it can't tag the picked bucket with metadata without a separate operator extension (e.g. an optional third element per weight tuple). Proposal: defer — ship the base object-return shape first.
```
JamieSinn: agreed, fractional is a lot more complicated especially with the nesting.

docs/architecture-decisions/targeting-variant-metadata.md
```diff
@@ -0,0 +1,194 @@
...
+- **JsonLogic operator registration for `variant`.** Registering `variant` as a passthrough operator is a workaround for a dependency quirk, not a documented contract of `github.com/diegoholiveira/jsonlogic`. Pin the behavior with an integration test so a future engine swap or upgrade doesn't silently regress single-key tagged-object returns.
+- **Metadata size.** No metadata field is capped today; consistency says leave uncapped, but a soft warning at config load is cheap if reviewers want it.
```
JamieSinn: I don't think we need a limit here. This is up to the operator to limit based on their business requirements.

JamieSinn (approved):

docs/architecture-decisions/targeting-variant-metadata.md
suthar26: sorry lost in updates. that is what i had before but i will update so the reason and details are sibling properties to variamt

juanparadox (approved):

erka (approved):

PR #2016, opened 2026-08-12
fix: record OFREP evaluation metrics

### LINKED ISSUE #2015: [BUG] OFREP endpoints never record flag evaluation metrics (feature_flag.flagd.impression / .result.reason)
## Description

flagd's OFREP HTTP handlers (`/ofrep/v1/evaluate/flags/{key}` and the bulk endpoint) never emit the `feature_flag.flagd.impression` and `feature_flag.flagd.result.reason` metrics, even though `/metrics` is healthy and these metrics work correctly for the same flags evaluated over Connect/gRPC.

## Root cause

In [`flagd/pkg/service/flag-evaluation/ofrep/handler.go`](https://github.com/open-feature/flagd/blob/main/flagd/pkg/service/flag-evaluation/ofrep/handler.go):

- `NewOfrepHandler` (line 40) accepts a `metricsRecorder telemetry.IMetricsRecorder` parameter, but only passes it into `metricsmw.NewHTTPMetric(...)` (lines 58-63, 67-72) for generic HTTP request count/duration middleware.
- The `handler` struct (line 32) never stores the recorder as a field.
- `HandleFlagEvaluation` (line 78) and `HandleBulkEvaluation` (line 120) call `h.evaluator.ResolveAsAnyValue` / `h.evaluator.ResolveAllValues` directly and return, without ever calling `RecordEvaluation`.

Compare this to the Connect/gRPC evaluation path, which does call it after every resolve:

- `flagd/pkg/service/flag-evaluation/flag_evaluator_v1.go:94` — `s.metrics.RecordEvaluation(ctx, resolved.Error, resolved.Reason, resolved.Variant, resolved.FlagKey)`
- `flagd/pkg/service/flag-evaluation/flag_evaluator_v2.go:257` — same pattern
- `flagd/pkg/service/flag-evaluation/flag_evaluator.go:92,386` — same pattern

So any flag evaluated only through OFREP is invisible to the `impression`/`result.reason` metrics, while identical flags evaluated over gRPC record correctly. Since these two metrics are flagd's primary evaluation-rate/error-rate signal, this makes evaluation observability silently incomplete for any OFREP-only client — with no error or log to indicate it.

## Expected behavior

`HandleFlagEvaluation` should call `RecordEvaluation` once per evaluation, and `HandleBulkEvaluation` should call it once per flag in the batch — mirroring the gRPC handlers.

## Suggested fix

Store the `metricsRecorder` on the `handler` struct in `NewOfrepHandler`, then call `h.metricsRecorder.RecordEvaluation(ctx, evaluation.Error, evaluation.Reason, evaluation.Variant, flagKey)` after `ResolveAsAnyValue` in `HandleFlagEvaluation`, and loop over each result calling the same in `HandleBulkEvaluation`.

## How we found this

Confirmed by deploying a real OpenFeature-Operator-injected sidecar (flagd `v0.15.4-patched`) and comparing: a client evaluating the same three flags over OFREP showed zero `feature_flag.flagd.impression`/`.result.reason` samples, while switching the same client to the gRPC/Connect provider produced correct metrics immediately, with no other configuration change.

### Expected Behavior

_No response_

### Steps to reproduce

_No response_

Boulea7:
## This PR

- Records evaluation metrics for each flag resolved through the OFREP single and bulk endpoints.
- Uses each bulk result's own error, reason, variant, and key so per-flag metrics remain accurate.

### Related Issues

Fixes #2015

### Notes

The full flagd lint command reports four existing staticcheck findings in `flag_evaluator_v1.go` and `ofrep_service.go`; the same findings reproduce on `origin/main`. The changed code introduces no new lint findings.

### How to test

- `go test -race -count=1 ./flagd/pkg/service/flag-evaluation/ofrep`
- `make test`
- `make build-flagd`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.7.2 run --new-from-rev=origin/main ./flagd/...`

### DIFF
--- a/flagd/pkg/service/flag-evaluation/ofrep/handler.go
+++ b/flagd/pkg/service/flag-evaluation/ofrep/handler.go
@@ -34,6 +34,7 @@ type handler struct {
 	evaluator                  evaluator.IEvaluator
 	contextValues              map[string]any
 	headerToContextKeyMappings map[string]string
+	metricsRecorder            telemetry.IMetricsRecorder
 	tracer                     trace.Tracer
 }
 
@@ -50,6 +51,7 @@ func NewOfrepHandler(
 		evaluator:                  evaluator,
 		contextValues:              contextValues,
 		headerToContextKeyMappings: headerToContextKeyMappings,
+		metricsRecorder:            metricsRecorder,
 		tracer:                     otel.Tracer("flagd.ofrep.v1"),
 	}
 
@@ -109,6 +111,9 @@ func (h *handler) HandleFlagEvaluation(w http.ResponseWriter, r *http.Request) {
 	ctx := context.WithValue(r.Context(), store.SelectorContextKey{}, selector)
 
 	evaluation := h.evaluator.ResolveAsAnyValue(ctx, requestID, flagKey, evaluationContext)
+	if h.metricsRecorder != nil {
+		h.metricsRecorder.RecordEvaluation(ctx, evaluation.Error, evaluation.Reason, evaluation.Variant, evaluation.FlagKey)
+	}
 	if evaluation.Error != nil {
 		status, evaluationError := ofrep.EvaluationErrorResponseFrom(evaluation)
 		h.writeJSONToResponse(status, evaluationError, w)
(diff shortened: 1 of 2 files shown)


toddbaert (approved):
Clean simple fix. Thanks @Boulea7 !

PR #2017, opened 2026-08-12
feat: report STALE reason when a sync source is disconnected

### LINKED ISSUE #400: [FEATURE] Detect stale flags
### Requirements

**Background**

Consider this conversation [1]. Flagd can operate with multiple sync providers. These sync providers push flag configurations which get stored in the internal in-memory `store` [2].

There could be occasions where individual providers shutdown, and disconnect for various reasons. For example, for an HTTP flag source, the source can undergo maintenance causing downtime. When this happens, flags stored internally could go stale 

**Proposal**

As highlighted by the conversation [1], it could be beneficial to mark stored flag configurations to be `stale` and set state `Reason.STALE` for evaluations. This makes sure flag configurations served by flagd are consistent with their sources. 

**Update**
Given that we do not have a `stale` sate in existing evaluation reasons [3], a good alternative is to utilize `CACHED` state

[1] - https://github.com/open-feature/flagd/pull/297#discussion_r1101875472
[2] - https://github.com/open-feature/flagd/tree/main/pkg/store
[3] - https://docs.openfeature.dev/docs/specification/types#resolution-details

scottt732:
Closes #400.

## The problem

When a sync source goes offline, flagd keeps serving the flags it already holds. That is the right call — last-known-good data beats failing evaluations open. But until now nothing distinguished those values from live ones:

- evaluations keep returning `STATIC` / `TARGETING_MATCH` indefinitely
- `/readyz` stays `200`, [documented](https://flagd.dev/reference/monitoring/) to latch on the first successful sync and "not change from there on"
- no metric exposes the condition — `feature_flag.flagd.sync.active_streams` counts *inbound* subscribers, not the health of flagd's own sync client

Behind a load balancer this is genuinely hard to diagnose. We hit it in production: one replica of eight lost its stream, kept serving a flag document from hours earlier, and stayed in the Service endpoints because every health signal said it was fine. Successive requests to the same Service returned different values for the same flag, and nothing indicated which answer was current.

This implements the behaviour agreed in #400, where `STALE` was subsequently added to the [OpenFeature specification](https://openfeature.dev/specification/types#resolution-details) for exactly this case.

## The change

- `sync.DataSync` gains a `Stale` field, marking a connection-state notification rather than a flag payload. `FlagData` is ignored and the store is left untouched.
- The gRPC sync emits one when its stream drops — both on initial failure and on each failed re-establishment — and stops once a payload arrives. The send is best-effort against the context so a blocked runtime can never stall the reconnect loop.
- `store.SourceState` records which sources are disconnected. It is nil-safe: a nil value reports nothing stale, so embedders that never wire it up keep the previous behaviour exactly.
- The runtime marks a source stale on such a notification and clears it on the next successful payload. It deliberately does **not** `Emit` — no flag data changed, and the sync protocol has no way to convey staleness downstream.
- The evaluator reports `model.StaleReason` for flags resolved from a disconnected source, wired via a new `evaluator.WithSourceState` option.

`STALE` only ever replaces a *successful* resolution. `ERROR` is preserved — a failed evaluation has no value to be uncertain about — and `FALLBACK` is left alone because it carries internal meaning that is translated in the API response.

I implemented this only for the gRPC sync, since that is the push-based case where a silently dead stream is invisible. The same `DataSync.Stale` mechanism extends to the other syncs without further plumbing if maintainers would like that in this PR or a follow-up.

## Why this shape

Because reasons already surface on the `feature_flag.flagd.result.reason` metric, a disconnected replica becomes visible per pod with **no new telemetry** — just a new value on an existing attribute. It also reaches SDKs through a spec-defined reason rather than anything flagd-specific.

## Testing

New: `core/pkg/store/source_state_test.go` (including nil-receiver and concurrency) and `core/pkg/evaluator/stale_test.go` (reason replacement, error preservation, source isolation, recovery, and the unconfigured no-op case).

Two existing gRPC sync tests asserted the data channel was empty after a stream ended, which is no longer true now that a stale notification is emitted; they now assert no further *flag payloads* arrive, and `Test_SyncRetry` additionally asserts the stale notification is emitted. While there, I fixed a latent `err.Error()` on a nil `err` in that test which turned an assertion failure into a segfault.

Full `core` and `flagd` suites pass.

Verified end to end with two real flagd instances, one syncing from the other:

```
1. connected:          {"value":true,"reason":"STATIC"}
2. source killed:      {"value":true,"reason":"STALE"}    <- value still served
3. source restarted:   {"value":true,"reason":"STATIC"}   <- recovered
```

Happy to adjust the naming, the `FALLBACK` carve-out, or split the test fixes into their own PR if you'd prefer.

### DIFF
--- a/core/pkg/evaluator/json.go
+++ b/core/pkg/evaluator/json.go
@@ -140,6 +140,18 @@ type Resolver struct {
 	store  store.IStore
 	Logger *logger.Logger
 	tracer trace.Tracer
+	// sourceState is optional; when nil, no flag is ever reported stale and
+	// behaviour is identical to releases before stale reporting existed.
+	sourceState *store.SourceState
+}
+
+// WithSourceState wires per-source connection state into the evaluator, so that
+// flags resolved from a sync source which is currently disconnected are reported
+// with model.StaleReason instead of their usual reason.
+func WithSourceState(s *store.SourceState) JSONEvaluatorOption {
+	return func(je *JSON) {
+		je.Resolver.sourceState = s
+	}
 }
 
 func NewResolver(store store.IStore, logger *logger.Logger, jsonEvalTracer trace.Tracer) Resolver {
@@ -344,6 +356,20 @@ func (je *Resolver) evaluateVariant(ctx context.Context, reqID string, flagKey s
 		if value != nil {
 			metadata[key] = value
 		}
+	}
+
+	// A flag resolved from a disconnected sync source may no longer match the
+	// source of truth. flagd deliberately keeps serving last-known-good data
+	// rather than failing the evaluation, so the uncertainty is surfaced in the
+	// reason instead. STALE only ever replaces a successful resolution: errors
+	// keep their own reason, and FALLBACK is left alone because it carries
+	// internal meaning that is translated in the API response.
+	if je.sourceState.IsStale(flag.Source) {
+		defer func() {
+			if err == nil && reason != model.ErrorReason && reason != model.FallbackReason {
+				reason = model.StaleReason
+			}
+		}()
 	}
 
 	if flag.State == Disabled {
--- a/core/pkg/evaluator/stale_test.go
+++ b/core/pkg/evaluator/stale_test.go
@@ -1,0 +1,133 @@
+//nolint:wrapcheck
+package evaluator_test
+
+import (
+	"context"
+	"testing"
+
+	flagdEvaluator "github.com/open-feature/flagd/core/pkg/evaluator"
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/model"
+	"github.com/open-feature/flagd/core/pkg/store"
+	"github.com/open-feature/flagd/core/pkg/sync"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const staleSource = "testSource"
+
+func staleEvaluator(t *testing.T, state *store.SourceState) *flagdEvaluator.JSON {
+	t.Helper()
+
+	e := flagdEvaluator.NewJSON(
+		logger.NewLogger(nil, false),
+		store.NewFlags(),
+		flagdEvaluator.WithSourceState(state),
+	)
+	require.NoError(t, e.SetState(sync.DataSync{FlagData: flagConfig, Source: staleSource}))
+
+	return e
+}
+
+// Flags are still served while their source is disconnected -- serving
+// last-known-good data beats failing evaluations open -- but the reason tells
+// the caller the value may no longer match the source of truth.
+func TestStale_ReplacesSuccessfulReasons(t *testing.T) {
+	tests := []struct {
+		name        string
+		flagKey     string
+		evalCtx     map[string]interface{}
+		freshReason string
+		expectedVal bool
+	}{
+		{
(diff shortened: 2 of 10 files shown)


aepfli:
I am not sure that a stal reason is the right approach. I see the problem, and this is definitely something we should help/support/make easier to find. But the reason of the flag evaluation is not STALE. This is the data state and not the evaluation outcome, and i feel like we are mixing domains. Still I have no better solution, but I feel like we should maybe think more carefully.

scottt732:
Thanks, @aepfli. Happy to rework the PR if you have any alternative approaches. In our case, one replica of eight lost its stream, kept serving a flag document from hours earlier, and stayed in the Service endpoints because every health signal said it was fine.

I stumbled on #400 which pointed out that [STALE is now available](https://openfeature.dev/specification/types/#resolution-details) as a resolution reason in the openfeature spec. But there are two distinct staleness conditions:
1. Client lost its connection to flagd - the JS provider detects this and emits `STALE`.
2. flagd lost its connection to its own upstream - what my PR tries to address here.

An in-process client can see (1) but is completely blind to (2). If our flagd's upstream sync breaks, every pod using it keeps serving rules downloaded hours earlier, perfectly happily, with no signal available to it at any layer.

aepfli:
the thing, why i am against this is "The provider's cached state is no longer valid and may not be up-to-date with the source of truth. " - so the STALE reason currently is about the prociders cached state, and not about flagd. so this would be something the provider would return on a PROVIDER_STALE or PROVIDER_ERROR when doing evaluations. Hence i feel like this is not the right approch to solve this issue.

Maybe we need to step back and evaluate, what we know and how we could tackle this differently, and what would be ideal. Do we want to be seen as healthy, when we cant fetch/update for a while? Our flagd providers have a grace period, when we switch from STALE to ERROR - maybe we should also add this, and expect to be healthy again within a period of time?

- should a failed sync affect the healthiness of a  flagd instance?
- should all failed syncs affect the healthness?

I feel like there are other ways to handle this (and we should most likely think about what is the best approach. @toddbaert wdyt?

toddbaert:
> the thing, why i am against this is "The provider's cached state is no longer valid and may not be up-to-date with the source of truth. " - so the STALE reason currently is about the prociders cached state, and not about flagd. so this would be something the provider would return on a PROVIDER_STALE or PROVIDER_ERROR when doing evaluations. Hence i feel like this is not the right approch to solve this issue.
> 
> Maybe we need to step back and evaluate, what we know and how we could tackle this differently, and what would be ideal. Do we want to be seen as healthy, when we cant fetch/update for a while? Our flagd providers have a grace period, when we switch from STALE to ERROR - maybe we should also add this, and expect to be healthy again within a period of time?
> 
>     * should a failed sync affect the healthiness of a  flagd instance?
> 
>     * should all failed syncs affect the healthness?
> 
> 
> I feel like there are other ways to handle this (and we should most likely think about what is the best approach. @toddbaert wdyt?

@aepfli I agree overall - `PROVIDER_STALE` is the business of the provider, not flagd itself or it's sync sources. Some connected providers might not even care about some sync sources. I think we might want to brainstorm what to do if flagd looses connection to a sync source, but the options really just come down to dying, warning, emitting metrics, etc - the provider downstream can't "do anything" in this situation; it's an upstream infrastructure issue and I don't want to pollute the evaluations with such things.

PR #2022, opened 2026-08-14
fix(flagd-proxy): set gRPC keepalive enforcement policy on the sync server

### LINKED ISSUE #2021: [BUG] flagd-proxy: sync stream is severed every ~90s by GOAWAY ENHANCE_YOUR_CALM (missing KeepaliveEnforcementPolicy)
### Observed behavior

`flagd-proxy` serves the same `FlagSyncService` as `flagd`, but creates its gRPC
server without a `KeepaliveEnforcementPolicy`. gRPC's defaults therefore apply
(`MinTime: 5m`, `PermitWithoutStream: false`), while the official in-process
providers ping every 30s with `PermitWithoutStream: true`. The server counts
these as ping strikes and tears the connection down with
`GOAWAY ENHANCE_YOUR_CALM ("too_many_pings")` roughly every 90 seconds.

`flagd` itself already fixed exactly this — `flagd-proxy` was missed.

#### What we observe

Every client pod cycles through provider states roughly every 90–180 seconds:

```
15:43:15.626  PROVIDER_STALE   "connection error"
15:43:20.627  PROVIDER_ERROR   "provider error"
15:43:27.642  PROVIDER_READY
15:46:28.649  PROVIDER_STALE   "connection error"
15:46:33.649  PROVIDER_ERROR   "provider error"
15:46:40.663  PROVIDER_READY
```

The intervals are identical to the millisecond across cycles
(`STALE→ERROR` = 5.000s = `defaultGracePeriod`, `STALE→READY` = 12.01s =
`RetryBackOffMaxMs`), which is what first told us this was a configured timer
rather than network flakiness.

The provider's own log carries the real cause:

```
sync cycle failed: stream error: rpc error: code = Unavailable
desc = closing transport due to: connection error: desc = "error reading from server: EOF",
received prior goaway: code: ENHANCE_YOUR_CALM, debug data: "too_many_pings"
```

#### Root cause

**Server side** — `flagd-proxy/pkg/service/server.go:107`:

```go
s.grpcServer = grpc.NewServer()
rpc.RegisterFlagSyncServiceServer(s.grpcServer, s.oldHandler)
syncv1.RegisterFlagSyncServiceServer(s.grpcServer, s.handler)
```

No `ServerOption` at all, so gRPC's default enforcement policy applies:
`MinTime: 5 * time.Minute`, `PermitWithoutStream: false`. The 5-minute `MinTime`
against a 30-second client ping is the part that actually causes the disconnects.

**Client side** — `go-sdk-contrib/providers/flagd/pkg/service/in_process/grpc_sync.go:130`:

```go
keepaliveParams := keepalive.ClientParameters{
    Time:                30 * time.Second, // defaultKeepaliveTime
    Timeout:             5 * time.Second,  // defaultKeepaliveTimeout
    PermitWithoutStream: true,
}
```

These are package-level constants with **no provider option to override them**.

Walking `grpc-go`'s `http2_server.go` (`handlePing`): because the `SyncFlags`
stream is long-lived, `len(t.activeStreams) >= 1`, so the accounting takes the
`else` branch and compares against `kep.MinTime` — **`PermitWithoutStream` is not
what bites here, `MinTime` alone is**. Each ping arriving sooner than `MinTime`
increments `pingStrikes`, and `GOAWAY` is sent once `pingStrikes > maxPingStrikes`
(`maxPingStrikes = 2`), i.e. on the third strike: pings at t=30s/60s/90s → GOAWAY
at ~90s, which is what we measure.

`resetPingStrikes` clears the counter whenever the server sends data, which is why
the interval stretches to ~180s on a flag set that is being updated and stays near
90s on an idle one.

#### This was already fixed in flagd — flagd-proxy was left out

This is **not a new proposal**. The exact same problem was reported and fixed for
flagd's flag-sync server in
[#1998](https://github.com/open-feature/flagd/issues/1998) (opened 2026-07-14, closed
as completed 2026-07-16), and surfaced on the operator side in
[open-feature-operator#852](https://github.com/open-feature/open-feature-operator/pull/852)
(merged 2026-07-31). #852 states its scope explicitly — *"Both the injected sidecar
and the standalone `Flagd` ... pick this up through the shared injection path"* — and
never mentions flagd-proxy.

flagd-proxy serves the **same `FlagSyncService`** to the **same clients** over the same
long-lived streams, so it needs the same policy. As of `bbb05d4` (2026-07-31, i.e.
after both of the above landed), `flagd-proxy/pkg/service/server.go:107` and `:123` are
still bare `grpc.NewServer()` calls with no keepalive enforcement policy and no CLI
flags to set one.

`flagd/pkg/service/flag-sync/sync_service.go:71` — note the comment:

```go
// keepAliveEnforcementPolicy builds the gRPC keepalive enforcement policy so
// provider keepalive pings on the long-lived SyncFlags stream aren't rejected
// with GOAWAY ENHANCE_YOUR_CALM.
func keepAliveEnforcementPolicy(cfg SvcConfigurations) keepalive.EnforcementPolicy {
    return keepalive.EnforcementPolicy{
        MinTime:             cfg.KeepAliveMinTime,
        PermitWithoutStream: cfg.KeepAlivePermitWithoutStream,
    }
}
```

wired in via `grpc.KeepaliveEnforcementPolicy(...)`, and exposed as
`--keep-alive-min-time` (default `30s`) and
`--keep-alive-permit-without-stream` (default `true`) in `flagd/cmd/start.go:102`.

Those defaults match the client's behaviour exactly. `flagd-proxy` has neither the
server option nor the flags — its CLI only accepts `--port`, `--management-port`
and `--log-format`.

#### Impact

For in-process providers the flag rules stay in memory, so **evaluation keeps
working throughout** — this is not an availability problem. What it does cost:

- **Rule propagation stalls for up to ~12s** per disconnect. Note the 12s is not
  reconnect latency: after a stream-level failure the provider sleeps the full
  `RetryBackOffMaxMs` (`grpc_sync.go`) instead of retrying with the initial backoff.
- **Provider state alarms are unusable.** With one provider a pod sits in
  `stale`/`error` roughly 4–8% of the time. Anything alerting on provider state
  fires continuously. Consumers subscribing to two flag sets see double the events.
- **Log noise**: one ERROR-level provider event per stream per ~90–180s per pod.

#### Workarounds we evaluated (none are viable)

- **open-feature-operator**: `FeatureFlagSource.keepAliveMinTime` /
  `keepAlivePermitWithoutStream` exist, but `ToEnvVars()` emits
  `FLAGD_KEEP_ALIVE_*` for the **injected flagd sidecar** only. The flagd-proxy
  deployment never receives them, and OFO hardcodes its args to
  `start --management-port <n>` — which flagd-proxy would reject anyway.
- **Client-side tuning**: `defaultKeepaliveTime` is a package constant with no
  `ProviderOption`. `WithGrpcDialOptionsOverride` *fully replaces*
  `buildDialOptions()`, so using it means reimplementing transport credentials,
  max message size and the retry service config, and re-diverging on every
  provider upgrade.
- **`WithRetryBackoffMaxMs`**: shortens the degraded window but does not stop the
  disconnects.

### Expected Behavior

The sync stream should stay up. flagd-proxy serves the same `FlagSyncService` to
the same providers as flagd, so it should enforce the same keepalive policy —
`MinTime: 30s`, `PermitWithoutStream: true` — and expose the same two flags, so
that client and server agree without anyone having to configure either side.

#### Proposed fix

Mirror what `flagd` already does — add the enforcement policy to flagd-proxy's
gRPC server, with the same defaults so client and server agree out of the box.
Happy to open the PR if the approach looks right.

### Steps to reproduce

flagd-proxy needs no special configuration to reproduce this — the defaults are
enough. Any deployment serving at least one long-lived `SyncFlags` stream to an
official in-process provider will show it.

1. Run `flagd-proxy start` (no keepalive flags exist, so nothing to set).
2. Point a Go in-process flagd provider at it — `flagd.WithInProcessResolver()`
   against the proxy's sync port. The provider's defaults are what matter:
   it pings every 30s with `PermitWithoutStream: true`, and neither is
   overridable through a `ProviderOption`.
3. Keep the stream open and watch the provider's events and logs.

Within ~90 seconds the stream is torn down with
`GOAWAY ENHANCE_YOUR_CALM ("too_many_pings")` and the provider cycles
`STALE → ERROR → READY`. On a flag set that is being updated the interval
stretches to ~180s, because `resetPingStrikes` clears the counter whenever the
server sends data.

Reproducing it without Kubernetes is also possible: build any gRPC server with a
bare `grpc.NewServer()`, connect a client with `keepalive.ClientParameters{Time:
10 * time.Second, PermitWithoutStream: true}` over a long-lived stream, and the
server sends GOAWAY on the third ping.

##### Environment

| | Version |
|---|---|
| flagd-proxy | `v0.9.4` (the default pinned by open-feature-operator) |
| open-feature-operator | deploys flagd-proxy automatically, `FLAGD_PROXY_TAG` default `v0.9.4` |
| Client | `go-sdk-contrib/providers/flagd` `v0.6.0`, in-process resolver |
| OpenFeature Go SDK | `v1.17.2` |
| Platform | GKE, in-cluster plaintext gRPC to `flagd-proxy-svc:8015` |

yuchou87:
### What

Adds `grpc.KeepaliveEnforcementPolicy` to flagd-proxy's sync gRPC server, plus
`--keep-alive-min-time` / `--keep-alive-permit-without-stream` flags, mirroring
what `flagd` already does in `flagd/pkg/service/flag-sync/sync_service.go`.

Fixes #2021.

### Why

flagd-proxy serves the same `FlagSyncService` to the same providers as flagd, but
built its server with a bare `grpc.NewServer()`. gRPC's default enforcement policy
(`MinTime: 5m`, `PermitWithoutStream: false`) rejects the providers' 30s keepalive
pings, so the server sends `GOAWAY ENHANCE_YOUR_CALM ("too_many_pings")` and drops
the sync stream about every 90 seconds. flagd fixed this in #1998; flagd-proxy was
missed.

Defaults are chosen to match `flagd` — and therefore the providers — so the fix
applies without anyone having to set a flag.

### Change

Four files, +223 / −4 (186 of those lines are tests).

`core/pkg/service/iservice.go` — two fields on the shared `Configuration`,
mirroring flagd's `SvcConfigurations`:

```go
// KeepAliveMinTime is the minimum interval the gRPC server permits between
// client keepalive pings. Pings arriving more frequently are rejected with
// GOAWAY (ENHANCE_YOUR_CALM). Honoured by services exposing a gRPC sync
// server; ignored by the connect-based flag evaluation service.
KeepAliveMinTime time.Duration
// KeepAlivePermitWithoutStream allows clients to send keepalive pings even
// when there is no active stream. Honoured by the same services as
// KeepAliveMinTime.
KeepAlivePermitWithoutStream bool
```

`flagd-proxy/pkg/service/server.go` — a `keepAliveEnforcementPolicy` helper
matching flagd's, applied in `startServer`:

```go
s.grpcServer = grpc.NewServer(
	grpc.KeepaliveEnforcementPolicy(keepAliveEnforcementPolicy(s.config)),
)
```

`flagd-proxy/cmd/start.go` — the two flags, viper bindings and config assignment.
Flag names, defaults and help text are copied verbatim from `flagd/cmd/start.go`
so the two binaries stay consistent.

### Testing

Two tests in `flagd-proxy/pkg/service/server_test.go`. They cover the same ground
as the ones that shipped with #1998 in
`flagd/pkg/service/flag-sync/sync_service_test.go`, but are deliberately not
copies of them: written as mirrors, SonarCloud flagged 19.4% duplication on new
code against a 3% gate, so the case tables here are expressed as
`service.Configuration` values instead. Same assertions, 0.9% duplication.

- `TestKeepAliveEnforcementPolicy` — table-driven check that the config maps onto
  `keepalive.EnforcementPolicy`.
- `TestServerKeepAliveEnforcement` — drives the HTTP/2 connection with a raw
  framer (grpc-go's client clamps its keepalive interval to a 10s minimum) and
  floods keepalive pings, asserting that a permissive policy tolerates them while
  a strict one tears the connection down with `GOAWAY ENHANCE_YOUR_CALM`.

Verified against `main` (`5c74b83`): all three modules build; `go test -race`
passes for `./flagd-proxy/...` and `./flagd/...`; `golangci-lint` reports nothing
new (the two `SA1019` hits in `handler.go` are pre-existing). Reverting the
`KeepaliveEnforcementPolicy` option makes `TestServerKeepAliveEnforcement` fail
with exactly the reported symptom, and the test was run 10× under `-race` to
check it isn't timing-sensitive.

Those tests build `Configuration` directly, so they say nothing about the
flag → viper → `Configuration` path. That gap matters here, because if the
binding silently failed, `MinTime` would be `0`, grpc-go would substitute its
5-minute default (`internal/transport/http2_server.go:250`), and the bug would be
back with every test still green. So I also drove the built binary with a raw
HTTP/2 client, pinging on a provider-like cadence:

| flagd-proxy started with | ping interval | result |
|---|---|---|
| *(no flags — shipped defaults)* | 31s | alive for the full 110s window |
| `--keep-alive-min-time=5m` (= the pre-fix gRPC default) | 31s | `GOAWAY ENHANCE_YOUR_CALM` at 93.0s |

The second row reproduces this issue at real timescales; the first shows the
shipped defaults resolve it. Two faster runs pin the individual flags:
`--keep-alive-min-time=1ms --keep-alive-permit-without-stream=true` survives 150
pings at 20ms intervals, and flipping only the boolean to `false` produces
`GOAWAY` immediately.

### Notes for reviewers

- **`startMetricsServer` is deliberately left on a bare `grpc.NewServer()`, and
  setting the policy there would have no effect anyway.** That server is
  dispatched via `grpcServer.ServeHTTP(...)` under h2c, so it runs on grpc-go's
  `serverHandlerTransport` (`internal/transport/handler_server.go`), which
  implements no ping-strike enforcement at all — `pingStrikes` and the keepalive
  policy live only in `http2_server.go`. `grpc.KeepaliveEnforcementPolicy` would
  be inert on that path. It also only serves short-lived health-check RPCs and
  isn't reachable by providers' sync clients, and the flags' help text says "the
  flag sync gRPC server", which would stop being accurate. Happy to revisit if
  you see it differently.
- **These flags are CLI-only on flagd-proxy, unlike on flagd.** `flagd-proxy/cmd`
  never calls `viper.SetEnvPrefix` / `SetEnvKeyReplacer` (there's a dangling
  `// allows environment variables to use _ instead of -` comment in `start.go`
  with no code behind it), so `FLAGD_KEEP_ALIVE_MIN_TIME` will not be picked up.
  This is pre-existing and affects `log-format` and `management-port` identically,
  so fixing it here felt like scope creep — but the ADR advertises env-var
  configurability, so I wanted to flag it. Happy to open a follow-up issue, or to
  fix it in this PR if you'd rather.
- **The ADR** (`docs/architecture-decisions/sync-server-keepalive-enforcement.md`,
  still `status: proposed`) is written entirely in terms of flagd's flag-sync
  server. I didn't edit it since it's authored by someone else and under review —
  say the word and I'll add a note that flagd-proxy carries the same two flags.
- No new dependencies.

### DIFF
--- a/core/pkg/service/iservice.go
+++ b/core/pkg/service/iservice.go
@@ -39,14 +39,9 @@ type Configuration struct {
 	MaxRequestHeaderBytes      int64
 	MaxRequestBodyBytes        int64
 
-	// KeepAliveMinTime is the minimum interval the gRPC server permits between
-	// client keepalive pings. Pings arriving more frequently are rejected with
-	// GOAWAY (ENHANCE_YOUR_CALM). Honoured by services exposing a gRPC sync
-	// server; ignored by the connect-based flag evaluation service.
+	// min interval the gRPC sync server permits between client keepalive pings; ignored by the connect eval service
 	KeepAliveMinTime time.Duration
-	// KeepAlivePermitWithoutStream allows clients to send keepalive pings even
-	// when there is no active stream. Honoured by the same services as
-	// KeepAliveMinTime.
+	// permit keepalive pings with no active stream; honoured by the same services as KeepAliveMinTime
 	KeepAlivePermitWithoutStream bool
 }
 
--- a/core/pkg/service/iservice.go
+++ b/core/pkg/service/iservice.go
@@ -38,6 +38,16 @@ type Configuration struct {
 	StreamDeadline             time.Duration
 	MaxRequestHeaderBytes      int64
 	MaxRequestBodyBytes        int64
+
+	// KeepAliveMinTime is the minimum interval the gRPC server permits between
+	// client keepalive pings. Pings arriving more frequently are rejected with
+	// GOAWAY (ENHANCE_YOUR_CALM). Honoured by services exposing a gRPC sync
+	// server; ignored by the connect-based flag evaluation service.
+	KeepAliveMinTime time.Duration
+	// KeepAlivePermitWithoutStream allows clients to send keepalive pings even
+	// when there is no active stream. Honoured by the same services as
+	// KeepAliveMinTime.
+	KeepAlivePermitWithoutStream bool
 }
 
 /*
--- a/flagd-proxy/cmd/start.go
+++ b/flagd-proxy/cmd/start.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/open-feature/flagd/core/pkg/logger"
 	iService "github.com/open-feature/flagd/core/pkg/service"
@@ -22,9 +23,11 @@ import (
 // start
 
 const (
-	logFormatFlagName      = "log-format"
-	managementPortFlagName = "management-port"
-	portFlagName           = "port"
+	logFormatFlagName                    = "log-format"
+	managementPortFlagName               = "management-port"
+	portFlagName                         = "port"
+	keepAliveMinTimeFlagName             = "keep-alive-min-time"
+	keepAlivePermitWithoutStreamFlagName = "keep-alive-permit-without-stream"
 )
 
 func init() {
@@ -34,10 +37,14 @@ func init() {
 	flags.Int32P(portFlagName, "p", 8015, "Port to listen on")
 	flags.Int32P(managementPortFlagName, "m", 8016, "Management port")
 	flags.StringP(logFormatFlagName, "z", "console", "Set the logging format, e.g. console or json")
+	flags.Duration(keepAliveMinTimeFlagName, 30*time.Second, "Minimum interval the flag sync gRPC server permits between client keepalive pings. Pings arriving more frequently than this are rejected with GOAWAY (ENHANCE_YOUR_CALM). Defaults to 30s.")
+	flags.Bool(keepAlivePermitWithoutStreamFlagName, true, "Permit clients of the flag sync gRPC server to send keepalive pings even when there is no active stream. Defaults to true.")
 
 	_ = viper.BindPFlag(logFormatFlagName, flags.Lookup(logFormatFlagName))
 	_ = viper.BindPFlag(managementPortFlagName, flags.Lookup(managementPortFlagName))
 	_ = viper.BindPFlag(portFlagName, flags.Lookup(portFlagName))
+	_ = viper.BindPFlag(keepAliveMinTimeFlagName, flags.Lookup(keepAliveMinTimeFlagName))
+	_ = viper.BindPFlag(keepAlivePermitWithoutStreamFlagName, flags.Lookup(keepAlivePermitWithoutStreamFlagName))
 }
 
 // startCmd represents the start command
@@ -69,6 +76,9 @@ var startCmd = &cobra.Command{
 			ReadinessProbe: func() bool { return true },
 			Port:           viper.GetUint16(portFlagName),
 			ManagementPort: viper.GetUint16(managementPortFlagName),
+
+			KeepAliveMinTime:             viper.GetDuration(keepAliveMinTimeFlagName),
+			KeepAlivePermitWithoutStream: viper.GetBool(keepAlivePermitWithoutStreamFlagName),
 		}
 
 		errChan := make(chan error, 1)
--- a/flagd-proxy/pkg/service/server.go
+++ b/flagd-proxy/pkg/service/server.go
@@ -96,9 +96,7 @@ func (s *Server) Shutdown() {
 	s.grpcServer.GracefulStop()
 }
 
-// keepAliveEnforcementPolicy builds the gRPC keepalive enforcement policy so
-// provider keepalive pings on the long-lived SyncFlags stream aren't rejected
-// with GOAWAY ENHANCE_YOUR_CALM.
+// keepalive policy so provider pings on the long-lived SyncFlags stream aren't rejected with GOAWAY ENHANCE_YOUR_CALM
 func keepAliveEnforcementPolicy(cfg service.Configuration) keepalive.EnforcementPolicy {
 	return keepalive.EnforcementPolicy{
 		MinTime:             cfg.KeepAliveMinTime,
--- a/flagd-proxy/pkg/service/server.go
+++ b/flagd-proxy/pkg/service/server.go
@@ -21,6 +21,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/health"
 	"google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/keepalive"
 )
 
 type Server struct {
@@ -95,6 +96,16 @@ func (s *Server) Shutdown() {
 	s.grpcServer.GracefulStop()
 }
 
+// keepAliveEnforcementPolicy builds the gRPC keepalive enforcement policy so
+// provider keepalive pings on the long-lived SyncFlags stream aren't rejected
+// with GOAWAY ENHANCE_YOUR_CALM.
+func keepAliveEnforcementPolicy(cfg service.Configuration) keepalive.EnforcementPolicy {
+	return keepalive.EnforcementPolicy{
+		MinTime:             cfg.KeepAliveMinTime,
+		PermitWithoutStream: cfg.KeepAlivePermitWithoutStream,
+	}
+}
+
 func (s *Server) startServer() error {
 	var lis net.Listener
 	var err error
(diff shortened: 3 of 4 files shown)


toddbaert (approved):
Thanks @yuchou87  - looks good. I trimmed some comments and removed some non-ascii chars, but I'll merge now.

PR #2024, opened 2026-08-14
fix(core): register the OFO scheme at package init, not inside Sync.Init

### LINKED ISSUE #2023: [BUG] flagd-proxy crashes with "concurrent map read and map write" — Sync.Init mutates the global runtime.Scheme at runtime
### Observed behavior

`kubernetes.Sync.Init` calls `v1beta1.AddToScheme(scheme.Scheme)` on the **global**
`client-go` scheme. In flagd-proxy this runs concurrently, one goroutine per sync
target, at arbitrary times during the process lifetime. `runtime.Scheme` holds several
plain maps and carries no lock; apimachinery documents it as
*"only threadsafe after registration is complete."* Two concurrent `Init` calls
therefore race on those maps and the Go runtime kills the process with
`fatal error: concurrent map read and map write` (exit code 2, unrecoverable).

We see flagd-proxy restart roughly once per hour in a test cluster with 4 sync targets.

#### Stack trace

```
fatal error: concurrent map read and map write

goroutine 2695 [running]:
internal/runtime/maps.fatal(...)
        /usr/local/go/src/runtime/panic.go:1046
k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(0xc000552a10, ...)
        k8s.io/apimachinery@v0.33.2/pkg/runtime/scheme.go:180
k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypes(...)
        k8s.io/apimachinery@v0.33.2/pkg/runtime/scheme.go:158
k8s.io/apimachinery/pkg/runtime.(*Scheme).AddUnversionedTypes(...)
        k8s.io/apimachinery@v0.33.2/pkg/runtime/scheme.go:134
k8s.io/apimachinery/pkg/apis/meta/v1.AddToGroupVersion(...)
        k8s.io/apimachinery@v0.33.2/pkg/apis/meta/v1/register.go:75
.../open-feature-operator/apis/core/v1beta1.init.0.(*Builder).Register.func1(...)
        sigs.k8s.io/controller-runtime@v0.19.3/pkg/scheme/scheme.go:72
sigs.k8s.io/controller-runtime/pkg/scheme.(*Builder).AddToScheme(...)
        sigs.k8s.io/controller-runtime@v0.19.3/pkg/scheme/scheme.go:86
github.com/open-feature/flagd/core/pkg/sync/kubernetes.(*Sync).Init(...)
        /src/core/pkg/sync/kubernetes/kubernetes_sync.go:72
github.com/open-feature/flagd/flagd-proxy/pkg/service/subscriptions.(*Coordinator).watchResource(...)
        /src/flagd-proxy/pkg/service/subscriptions/manager.go:201
created by ...(*Coordinator).RegisterSubscription in goroutine 2694
        /src/flagd-proxy/pkg/service/subscriptions/manager.go:128
```

#### Root cause

`core/pkg/sync/kubernetes/kubernetes_sync.go:76` (main):

```go
func (k *Sync) Init(_ context.Context) error {
    ...
    if err := v1beta1.AddToScheme(scheme.Scheme); err != nil {   // global singleton
```

`k8s.io/apimachinery/pkg/runtime.Scheme` (v0.33.2, `pkg/runtime/scheme.go:47`):

```go
// Schemes are not expected to change at runtime and are only threadsafe after
// registration is complete.
type Scheme struct {
    gvkToType        map[schema.GroupVersionKind]reflect.Type
    typeToGVK        map[reflect.Type][]schema.GroupVersionKind
    ...
}
```

No mutex anywhere in the struct.

#### Why only flagd-proxy is affected

Both binaries consume this sync through `core/pkg/sync/builder`, but they call `Init`
very differently:

**flagd** — serial, at startup, single goroutine (`flagd/pkg/runtime/runtime.go:60`):

```go
for _, s := range r.Syncs {
    if err := s.Init(gCtx); err != nil {
        return fmt.Errorf("sync provider Init returned error: %w", err)
    }
}
```

**flagd-proxy** — one goroutine per target, at arbitrary runtime moments
(`flagd-proxy/pkg/service/subscriptions/manager.go:128`):

```go
go s.watchResource(target)      // → manager.go:201 syncSource.Init(ctx)
```

So the contract violation lives in `core`, and flagd-proxy's concurrency model is what
makes it fire. Two **distinct** targets registering close in time are enough — a second
subscriber to an *existing* target reuses the multiplexer and does not call `Init`.

In practice this happens on every flagd-proxy restart, when all client pods reconnect
at once. That makes it self-reinforcing: crash → reconnect storm → crash. One of our
crashed instances logged zero `starting kubernetes sync notifier` lines, i.e. it died
during `Init` before any `Sync()` ever started.

#### Impact

- flagd-proxy restarts repeatedly (~1/h for us) with exit code 2.
- Clients already holding flag state are unaffected (in-process evaluation continues
  against cached rules), so this is not an evaluation-availability issue.
- CR updates landing inside a crash window are delayed until the next resync
  (`resyncPeriod = 1 * time.Minute`).
- **Pods starting during a crash loop receive no rules at all** and fall back to
  built-in defaults — the practically dangerous case.

### Expected Behavior

Registering types into the global client-go scheme should happen once, on a
single goroutine, before anything can use the scheme concurrently — which is
exactly what `runtime.Scheme` documents as its contract. flagd-proxy should not
crash when two sync targets are registered at the same time.

#### Proposed fix

Move the registration to package `init()` so it happens once, single-goroutine, before
any concurrent use — matching what `Scheme` documents:

```go
func init() {
    utilruntime.Must(v1beta1.AddToScheme(scheme.Scheme))
}
```

and drop the `AddToScheme` block from `Sync.Init`. Happy to open a PR.

### Steps to reproduce

No special configuration is needed — the trigger is simply more than one sync
target being registered concurrently.

1. Run flagd-proxy against a cluster with **at least two distinct sync targets**
   (distinct namespace/FeatureFlag pairs). A second subscriber to an *existing*
   target reuses the multiplexer and never calls `Init`, so it does not count —
   the targets must be distinct.
2. Have clients subscribe to both targets close together in time.
   `Coordinator.RegisterSubscription` starts `go s.watchResource(target)` per
   target, and each of those goroutines calls `syncSource.Init(ctx)`, which
   writes to the global `scheme.Scheme`.
3. Two such writes overlapping is enough: the process dies with
   `fatal error: concurrent map read and map write` and exit code 2.

The easiest way to force step 2 is to restart flagd-proxy while several client
pods are connected — they all reconnect at once, which registers every target
within the same instant. That also makes the failure self-reinforcing in
practice: crash → reconnect storm → crash.

Because this is a data race on unguarded maps, it is timing-dependent rather
than deterministic. In a test cluster with 4 sync targets and several client
pods we see it roughly once per hour. One crashed instance logged zero
`starting kubernetes sync notifier` lines, i.e. it died inside `Init` before any
`Sync()` had started.

##### Environment

- flagd-proxy `v0.9.4` (the operator's default image) pinning `core v0.13.1` — the
  crash stack, the `Sync()` log line number and the `apis/` vs `api/` operator import
  path all match that build exactly
- **Still present on `flagd-proxy/v0.9.7` (`core v0.15.6`) and on `main`** — verified
  at `5c74b83`, where `AddToScheme` sits at `kubernetes_sync.go:76`, unchanged.
  Upgrading does not help.
- `k8s.io/apimachinery` v0.33.2
- `sigs.k8s.io/controller-runtime` v0.19.3
- 4 distinct sync targets (two namespaces, two FeatureFlag CRs each), several client pods

yuchou87:
### What

Moves `v1beta1.AddToScheme(scheme.Scheme)` out of `kubernetes.Sync.Init` and into
the package's `init()`, so the global client-go scheme is mutated once, on a single
goroutine, before anything can use it concurrently.

Fixes #2023.

### Why

`runtime.Scheme` holds plain maps and carries no lock. apimachinery states the
contract on the type itself:

```go
// Schemes are not expected to change at runtime and are only threadsafe after
// registration is complete.
type Scheme struct {
	gvkToType map[schema.GroupVersionKind]reflect.Type
	...
```

`Sync.Init` violates that by writing to the **global** `scheme.Scheme` at runtime.
That is harmless in flagd, which calls `Init` in a serial startup loop
(`flagd/pkg/runtime/runtime.go:60`), but flagd-proxy calls it from one goroutine
per sync target at arbitrary times (`flagd-proxy/pkg/service/subscriptions/manager.go:128`
→ `:201`), so two targets registering close together race on those maps and the Go
runtime kills the process:

```
fatal error: concurrent map read and map write

goroutine 2695 [running]:
k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName(...)
	k8s.io/apimachinery@v0.33.2/pkg/runtime/scheme.go:180
...
github.com/open-feature/flagd/core/pkg/sync/kubernetes.(*Sync).Init(...)
	/src/core/pkg/sync/kubernetes/kubernetes_sync.go:72
github.com/open-feature/flagd/flagd-proxy/pkg/service/subscriptions.(*Coordinator).watchResource(...)
	/src/flagd-proxy/pkg/service/subscriptions/manager.go:201
created by ...(*Coordinator).RegisterSubscription in goroutine 2694
	/src/flagd-proxy/pkg/service/subscriptions/manager.go:128
```

We see this roughly once an hour in a test cluster with four sync targets. Two
distinct targets registering close in time are enough — a second subscriber to an
*existing* target reuses the multiplexer and never calls `Init`. In practice it is
self-reinforcing: the crash makes every client reconnect at once, which is exactly
the condition that triggers it.

Registering at package init matches what the type documents and costs nothing: the
work was already idempotent, so doing it once instead of per-`Init` is strictly
less work. `utilruntime.Must` is the conventional Kubernetes idiom here — the call
only touches in-memory maps, so a failure would mean conflicting type definitions,
which is worth failing at startup for.

### Change

All of `core/pkg/sync/kubernetes/kubernetes_sync.go`:

```diff
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -31,6 +32,14 @@ var (
 
 const invalidAPIVersionMsg = "invalid api version %s, expected %s"
 
+// Registering into the global client-go scheme has to happen once, before
+// anything can use it concurrently: runtime.Scheme is documented as "only
+// threadsafe after registration is complete". Doing it from Sync.Init instead
+// races on its unguarded maps, see https://github.com/open-feature/flagd/issues/2023.
+func init() {
+	utilruntime.Must(v1beta1.AddToScheme(scheme.Scheme))
+}
+
 type SyncOption func(s *Sync)
 
 type Sync struct {
@@ -73,10 +82,6 @@ func (k *Sync) Init(_ context.Context) error {
 		return fmt.Errorf("unable to parse uri %s: %w", k.URI, err)
 	}
 
-	if err := v1beta1.AddToScheme(scheme.Scheme); err != nil {
-		return fmt.Errorf("unable to v1beta1 types to scheme: %w", err)
-	}
-
 	// The created informer will not do resyncs if the given defaultEventHandlerResyncPeriod is zero.
 	// For more details on resync implications refer to tools/cache/shared_informer.go
 	factory := dynamicinformer.NewFilteredDynamicSharedInformerFactory(k.dynamicClient, resyncPeriod, k.namespace, nil)
```

`Sync.Init`'s signature is unchanged, and neither caller distinguished that error
from the URI-parsing one, so this is not a breaking change.

No new dependency — `utilruntime` is already part of `k8s.io/apimachinery`. The
rest of the diff is the two tests described below.

### Testing

Two regression tests in `kubernetes_sync_test.go`:

- **`TestConcurrentInit`** — reproduces the reported crash directly: 16 goroutines
  calling `Sync.Init` concurrently. With the fix reverted this reports `DATA RACE`
  on 5 runs out of 5 (65–541 reports per run), pointing at `AddToScheme` ←
  `Sync.Init`; with the fix it passes. `make test` already runs `-race`, so this is
  caught in CI.
- **`TestSchemeRegisteredAtPackageInit`** — asserts the registration happened by
  package-init time, which holds without `-race` too. It captures the state from an
  `init()` rather than reading it in the test body, because `Sync.Init` registers
  the types as a side effect and the pre-existing `TestInit` calls it — read from a
  test function, the check passes whether or not the fix is present. I verified that
  false negative before settling on this shape.

Verified against `main` (`5c74b83`): all three modules build and vet clean;
`go test -race` passes for `./core/...` (19 packages), and for the touched package
under `-count=3` and three rounds of `-shuffle=on`; `golangci-lint` reports 0 issues
for it.

Also removed a now-redundant `v1beta1.AddToScheme` from `TestNotify` — package init
covers it, and it was a second global mutation from a test.

### One thing worth your call

While checking this I found that **nothing in-tree appears to consume the global
registration at all.** With the `init()` and the `k8s.io/client-go/kubernetes/scheme`
import deleted outright, all three modules build and the whole `kubernetes` package
test suite passes (only my new guard test fails, by construction). The real `dynamic`
client and `dynamicinformer` never touch `scheme.Scheme`; `toFFCfg` uses
`runtime.DefaultUnstructuredConverter`; only `dynamic/fake` consumes a scheme, and
the tests can pass their own.

So deleting the global mutation may be the better fix. I did not propose that here
because `core` is an importable module and dropping a global registration is an
observable change for downstream importers, which felt like your call rather than
mine. Happy to switch this PR to the deletion if you prefer it.

### DIFF
--- a/core/pkg/sync/kubernetes/kubernetes_sync.go
+++ b/core/pkg/sync/kubernetes/kubernetes_sync.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -30,6 +31,14 @@ var (
 )
 
 const invalidAPIVersionMsg = "invalid api version %s, expected %s"
+
+// Registering into the global client-go scheme has to happen once, before
+// anything can use it concurrently: runtime.Scheme is documented as "only
+// threadsafe after registration is complete". Doing it from Sync.Init instead
+// races on its unguarded maps, see https://github.com/open-feature/flagd/issues/2023.
+func init() {
+	utilruntime.Must(v1beta1.AddToScheme(scheme.Scheme))
+}
 
 type SyncOption func(s *Sync)
 
@@ -71,10 +80,6 @@ func (k *Sync) Init(_ context.Context) error {
 	k.namespace, k.crdName, err = parseURI(k.URI)
 	if err != nil {
 		return fmt.Errorf("unable to parse uri %s: %w", k.URI, err)
-	}
-
-	if err := v1beta1.AddToScheme(scheme.Scheme); err != nil {
-		return fmt.Errorf("unable to v1beta1 types to scheme: %w", err)
 	}
 
 	// The created informer will not do resyncs if the given defaultEventHandlerResyncPeriod is zero.
--- a/core/pkg/sync/kubernetes/kubernetes_sync.go
+++ b/core/pkg/sync/kubernetes/kubernetes_sync.go
@@ -32,10 +32,7 @@ var (
 
 const invalidAPIVersionMsg = "invalid api version %s, expected %s"
 
-// Registering into the global client-go scheme has to happen once, before
-// anything can use it concurrently: runtime.Scheme is documented as "only
-// threadsafe after registration is complete". Doing it from Sync.Init instead
-// races on its unguarded maps, see https://github.com/open-feature/flagd/issues/2023.
+// register into the global client-go scheme once at init; doing it in Sync.Init races the scheme's maps (#2023)
 func init() {
 	utilruntime.Must(v1beta1.AddToScheme(scheme.Scheme))
 }
--- a/core/pkg/sync/kubernetes/kubernetes_sync_test.go
+++ b/core/pkg/sync/kubernetes/kubernetes_sync_test.go
@@ -582,18 +582,14 @@ func TestSync_watcher(t *testing.T) {
 	}
 }
 
-// Captured from init() rather than read inside the test: Sync.Init registers
-// these types as a side effect, so once TestInit has run the scheme looks
-// populated either way. A package-level variable initialiser would be too early
-// — Go initialises all package variables before running any init function.
+// capture in init(), not a package var (too early) nor in-test (Sync.Init would mask it)
 var schemeRegisteredAtPackageInit bool
 
 func init() {
 	schemeRegisteredAtPackageInit = scheme.Scheme.Recognizes(v1beta1.GroupVersion.WithKind("FeatureFlag"))
 }
 
-// TestSchemeRegisteredAtPackageInit pins the fix without needing -race, which
-// complements TestConcurrentInit below.
+// pins the fix without needing -race
 func TestSchemeRegisteredAtPackageInit(t *testing.T) {
 	if !schemeRegisteredAtPackageInit {
 		t.Error("expected the FeatureFlag type to be registered in the global scheme by package init; " +
@@ -601,10 +597,7 @@ func TestSchemeRegisteredAtPackageInit(t *testing.T) {
 	}
 }
 
-// TestConcurrentInit reproduces the reported crash: flagd-proxy calls Init from
-// one goroutine per sync target, so registering into the global scheme there
-// races on its unguarded maps. Run under -race, which the Makefile's test target
-// uses.
+// reproduces #2023: concurrent Sync.Init races the global scheme (run under -race)
 func TestConcurrentInit(t *testing.T) {
 	var wg stdsync.WaitGroup
 	for i := 0; i < 16; i++ {
@@ -761,7 +754,7 @@ func TestNotify(t *testing.T) {
 	// only delivers events via the watch stream, not via the initial List/Replace.
 	// Pre-populating causes the object to silently land in the store without
 	// triggering AddFunc, so later UpdateStatus calls are seen as updates on an
-	// object the informer has never "added" — leaving the test waiting forever.
+	// object the informer has never "added", leaving the test waiting forever.
 	fc := fake.NewSimpleDynamicClient(scheme.Scheme)
 	l, err := logger.NewZapLogger(zapcore.FatalLevel, "console")
 	if err != nil {
--- a/core/pkg/sync/kubernetes/kubernetes_sync_test.go
+++ b/core/pkg/sync/kubernetes/kubernetes_sync_test.go
@@ -582,6 +582,47 @@ func TestSync_watcher(t *testing.T) {
 	}
 }
 
+// Captured from init() rather than read inside the test: Sync.Init registers
+// these types as a side effect, so once TestInit has run the scheme looks
+// populated either way. A package-level variable initialiser would be too early
+// — Go initialises all package variables before running any init function.
+var schemeRegisteredAtPackageInit bool
+
+func init() {
+	schemeRegisteredAtPackageInit = scheme.Scheme.Recognizes(v1beta1.GroupVersion.WithKind("FeatureFlag"))
+}
+
+// TestSchemeRegisteredAtPackageInit pins the fix without needing -race, which
+// complements TestConcurrentInit below.
+func TestSchemeRegisteredAtPackageInit(t *testing.T) {
+	if !schemeRegisteredAtPackageInit {
+		t.Error("expected the FeatureFlag type to be registered in the global scheme by package init; " +
+			"registering it from Sync.Init instead races on the scheme's unguarded maps")
+	}
+}
+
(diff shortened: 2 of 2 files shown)


toddbaert (approved):
Nice catch, thanks! @yuchou87 

Same addition as https://github.com/open-feature/flagd/pull/2022#pullrequestreview-4940474979.

Issue #1088, opened 2023-12-18
Remove deprecated proto compatibility


toddbaert:
Remove all the old protobuf implementations once the [new ones](https://github.com/open-feature/java-sdk-contrib/issues/621) have been available for a while :tm: (specifically, we should wait until every provider is updated to use the new RPCs).

:warning: BLOCKED until we have updated providers.

Kavindu-Dodan:
@toddbaert saw a todo in flagd and given we have upgraded all providers, we can work on this

toddbaert:
Did a sweep of the flagd providers to check the "every provider updated to the new RPCs" blocker:

| Provider | Evaluation RPC |
|---|---|
| Go, Java, JS, Python, .NET, PHP | `evaluation.v2` ✅ |
| **Ruby** | `evaluation.v1` ❌ |
| Rust | n/a (no flagd provider) |

**Ruby is the last `evaluation.v1` holdout** — its `client.rb` still requires `flagd/evaluation/v1/evaluation_services_pb` and its live docker-based tests exercise the v1 service. Removing `evaluation.v1` would break it. Tracked in open-feature/ruby-sdk-contrib#209.

Note: `schema.v1` (the older monolithic schema) looks safe to drop now — nobody calls it (providers moved to the split `evaluation.v1` then `v2`); it's only shipped as dead generated stubs in a couple of repos. It's specifically **`evaluation.v1`** that's still gated on the Ruby migration.

PR #2025, opened 2026-08-14
fix: possible race on eventstream connection

toddbaert:
Fixes a bug found while working on https://github.com/open-feature/flagd/pull/1993 - an extremely rare but real race (found running with `--race`) that can occur if a client disconnects right after connecting to the event stream.

The easiest fix is to just send the initial ready event from the handler, not the internal subscription channel as before, because then we don't have to enhance the concurrency stuff there.


core/pkg/sync/http/mock/http.go
```diff
@@ -53,64 +53,3 @@ func (mr *MockClientMockRecorder) Do(req any) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "Do", reflect.TypeOf((*MockClient)(nil).Do), req)
 }
-
```
toddbaert, line 56: Unrelated generated delta due to the mockgen change [here](https://github.com/open-feature/flagd/pull/2025/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52).

flagd/pkg/service/flag-evaluation/flag_evaluator_v1.go
```diff
@@ -173,8 +173,9 @@ func (s *FlagEvaluationService) EventStream(
 	s.eventingConfiguration.Subscribe(ctx, req, &selector, requestNotificationChan)
 	defer s.eventingConfiguration.Unsubscribe(req)
 
-	requestNotificationChan <- service.Notification{
-		Type: service.ProviderReady,
+	// send initial ready to the stream; the notifier channel may be closed by cleanup and this avoids racing with it
```
toddbaert, line 176: `v2` is the only handler version in use - v1 and the unversioned one are old handlers which are deprecated - only the PHP uses them; will be [removed soon](https://github.com/open-feature/flagd/issues/1088) but I made the fix in all 3 places for consistency.

jonathannorris (approved):

PR #2031, opened 2026-08-18
fix(flagd-proxy): rebuild a multiplexer whose watcher has stopped

### LINKED ISSUE #2030: [BUG] flagd-proxy: subscribing to a not-yet-existing FeatureFlag permanently wedges that target until restart
### Observed behavior

If a client subscribes to a `FeatureFlag` CR that does not exist yet, and the CR is
created afterwards, the client **never** receives the flag data. flagd-proxy stops
logging anything at all for that target — not even `starting sync` — even though the
client keeps reconnecting every 1–12s.

Restarting the **client** does not help. Only restarting **flagd-proxy** does, and
recovery is then immediate (the client does not even need to reconnect deliberately).

Since flagd-proxy is a cluster-wide singleton, the only known recovery affects every
service in the cluster.

#### Why this matters

This is the normal onboarding order, not an edge case:

1. A service integrates the SDK and ships. It is not using any flag yet, so **no CR
   exists**.
2. Later, someone enables the first flag → the control plane creates the CR.
3. Running pods are expected to pick it up within seconds.

Step 3 never happens. Dynamic flag delivery degrades into "restart the shared proxy".

This also explains why it stayed hidden for us: every service we had onboarded so far
was created **CR-first**, so none of them had ever taken the "service exists before the
CR" path.

#### Evidence

While the CR was missing (repeated every 1–12s, dozens of times over ~3 minutes):

```
starting sync from my-ns/my-flags
error with the initial fetch: unable to fetch FeatureFlag my-ns/my-flags:
  featureflags.core.openfeature.dev "my-flags" not found
manager.go:212 error from sync for target core.openfeature.dev/my-ns/my-flags: ...
```

The CR was then created, with complete content.

After that — **nothing** for that target, for 20+ minutes, while other targets kept
emitting `kube sync notifier event` every 30–60s. So the proxy process itself was
healthy.

A client pod started **9 minutes after** the CR existed still timed out:

```
Provider flagd transitioned from state NOT_READY to state ERROR
Initialization timeout exceeded; did not complete within the 10000 ms deadline
```

and produced **no** proxy-side log line at all — no `starting sync`, no
`unable to initiate sync`.

Client and network ruled out from inside that pod: DNS resolved, TCP to
`flagd-proxy-svc:8015` connected, and evaluation returned `FLAG_NOT_FOUND` rather than
`PROVIDER_NOT_READY` — i.e. the provider was serving, its store was simply empty.

After `rollout restart deploy/flagd-proxy`, **without touching the client**:

```
starting sync from my-ns/my-flags        # no error this time
kube sync notifier event: add: my-ns my-flags
```

The client had been retrying the whole time, so what was wedged was the proxy's
subscription state, not the client.

#### Root cause

`flagd-proxy/pkg/service/subscriptions/manager.go`.

`RegisterSubscription` decides purely on map membership and never checks whether the
multiplexer is still alive:

```go
sh, ok := s.multiplexers[target]
if !ok {
    s.multiplexers[target] = &multiplexer{...}
    go s.watchResource(target)                 // watcher started only when absent
} else {
    sh.subs[key] = storedChannels{...}         // attach only; no watcher restarted
    if sh.syncRef != nil {
        go func() {
            s.mu.RLock()
            defer s.mu.RUnlock()
            if _, ok := s.multiplexers[target]; ok {   // re-checks membership, not liveness
                ...sh.syncRef.ReSync(ctx, dataSync)...
            }
        }()
    }
}
```

The `else` branch does attempt a `ReSync`, which is what would otherwise rescue the new
subscriber. But that attempt is guarded by a second membership check, and it runs on its
own goroutine — so when the delete lands in between, the check fails and the `ReSync` is
skipped. Neither guard tests whether the multiplexer is still being watched.

Meanwhile `watchResource` removes the entry **asynchronously** on ctx cancellation:

```go
ctx, cancel := context.WithCancel(s.ctx)
defer cancel()
...
go func() {
    <-ctx.Done()
    s.mu.Lock()
    delete(s.multiplexers, target)       // deleted on another goroutine
    s.mu.Unlock()
}()
...
err = syncSource.Sync(ctx, sh.dataSync)
if err != nil {
    s.logger.Error(...)
    sh.broadcastError(s.logger, err)
}
// function returns → deferred cancel() → only then does the goroutine delete
```

A subscription arriving between `watchResource` returning and that goroutine deleting
the entry takes the `else` branch and attaches to a multiplexer whose context is
already cancelled and whose broadcast goroutine has exited. It can never receive data,
and nothing will restart a watcher for it.

With the client retrying every 1–12s and dozens of failures while the CR was missing,
landing inside that window is close to inevitable.

#### The trigger is probably not specific to a missing CR

A missing CR is how we hit this, but reading the code, the window opens whenever
`watchResource` returns — which is after *any* `Sync` error, including transient ones
such as an API server blip or a dropped watch connection. There is also a self-feeding
loop: the error is broadcast to the subscribers, the client reconnects promptly, and
that reconnect lands in the window that the same error just opened. A missing CR is
simply the variant that fails over and over, giving many chances to land there.

We have only observed the missing-CR case in production, so treat the wider framing as a
code-level inference — though the reproduction below does support it, since it wedges
using an ordinary error rather than anything Kubernetes-specific.

This distinction matters for the fix: if the trigger really is only the onboarding
order, closing the window (the first option below) is enough; if any transient sync error
can do it, the liveness check (the second option) is the one that actually holds.

#### Reproduced in a unit test

The interleaving above is not just inferred — it reproduces against `main` (`e045237`)
using the mocks already in `flagd-proxy/pkg/service/subscriptions/manager_test.go`, with
no cluster involved. `syncMock.errChanIn` stands in for "the FeatureFlag does not exist",
and `broadcastError` runs *before* `watchResource` returns, so observing the error puts
the second subscription inside the window by construction rather than by luck:

1. `RegisterSubscription(k1)` → `!ok` branch → `watchResource` starts; confirm it is live
   by pushing one `DataSync` through.
2. Send an error on `errChanIn` → `Sync` returns → `watchResource` broadcasts and is
   about to return.
3. `RegisterSubscription(k2)` → `else` branch.
4. Push another `DataSync`. **k2 never receives it.**

Debug logging confirms each step:

```
manager.go:112  sync handler does not exist for target ns/flags, registering multiplexer with sub k1
manager.go:163  watching resource ns/flags
manager.go:212  error from sync for target ns/flags: featureflags.core.openfeature.dev "flags" not found
                multiplexer still present in the map: true
manager.go:131  registering sync subscription k2          <- else branch, no watcher restarted
                k2 received nothing
```

Note what is *absent*: there is no `triggering a resync` line. The `ReSync` goroutine ran
after the delete had landed, so its membership re-check failed and it returned without
doing anything. Both recovery paths miss.

The timing is what decides it, and both outcomes reproduce:

| when the second subscription arrives | branch | result |
|---|---|---|
| entry still present (inside the window) | `else` | wedged — no watcher, and the `ReSync` is skipped |
| entry already deleted | `!ok` | recovers — a new multiplexer and watcher are built |

Run as a loop without deliberately timing step 3, 14 of 20 subscriptions were wedged.

Note that `manager.go:212` is the same line as in the production log quoted above.

We are happy to contribute this as a regression test alongside a fix.

#### What is and is not established

**Confirmed**: the symptom reproduces in a real cluster; only a proxy restart recovers
it; client and network are healthy; the CR content is complete; the proxy is silent for
the affected target while healthy for others. **The interleaving described above is also
confirmed** — directly observed in the unit-test reproduction, not inferred.

**Not explained**: why the window stayed open for so long in production. A client pod
starting 9 minutes after the CR existed still took the `else` branch, which means the
entry was still in the map at that point — long after the failing `watchResource` had
returned and its deferred `cancel()` should have triggered the delete. The unit test
reproduces the wedge within the normal, short window; it does not reproduce a window
that stays open for minutes. Something else may be holding the entry alive, and the fix
below should be judged on its own merits rather than on this being fully understood.

#### Impact

| | |
|---|---|
| Evaluation availability | No crash, but **every flag for that service resolves to its built-in default** until the proxy restarts |
| Direction of failure | Safe — flags stay off rather than on |
| Dynamic delivery | **Completely broken**, which is the point of a flag system |
| Blast radius of the workaround | Restarting a cluster-wide singleton affects every service in the cluster |

#### How to tell it is this

- Client side: initialization timeout, and evaluation returns `FLAG_NOT_FOUND` rather
  than `PROVIDER_NOT_READY`
- Proxy side: **no log lines at all** for that target, while other targets keep
  emitting `notifier event` normally

#### Workaround

Create the `FeatureFlag` CR **before** the service starts — even an empty one. That way
the resource already exists at first subscription and the bad state is never entered.

For an instance already wedged, restart flagd-proxy.

### Expected Behavior

A subscription to a `FeatureFlag` that does not exist yet should start delivering data
once the CR is created, without operator intervention. A target should never be left in
a state where no watcher is running and nothing will ever start one — and if it is,
recovering it should not require restarting a cluster-wide singleton.

#### Proposed fix

Either make the removal synchronous, closing the window:

```go
defer func() {
    s.mu.Lock()
    delete(s.multiplexers, target)
    s.mu.Unlock()
    cancel()
}()
```

or, more robustly, validate liveness before attaching in the `else` branch and rebuild
the multiplexer when it is dead:

```go
sh, ok := s.multiplexers[target]
if !ok || sh.isDead() {
    // rebuild + go s.watchResource(target)
}
```

The second also covers a stale entry left behind for any other reason, and does not
depend on pinning down the exact interleaving. The first is smaller and more direct.
Ideally both. Happy to open a PR.

### Steps to reproduce

There are two ways to reproduce. The first needs nothing but `go test`.

##### Without a cluster

Described in full under "Reproduced in a unit test" above: drive
`RegisterSubscription` directly with the mocks already in
`flagd-proxy/pkg/service/subscriptions/manager_test.go`, fail the sync via
`syncMock.errChanIn`, and register a second subscription while the entry is still in the
map. No Kubernetes, no client SDK, no provider. This is how we confirmed the
interleaving, and we are happy to contribute it as a regression test.

##### In a cluster (how we originally hit it)

1. Deploy flagd-proxy.
2. Point an in-process provider at `core.openfeature.dev/<ns>/<name>` where that
   `FeatureFlag` does **not** exist. Let it retry for a minute or two.
3. `kubectl apply` the `FeatureFlag`.
4. Observe: the client still times out; flagd-proxy logs nothing for that target.
5. `kubectl rollout restart deploy/flagd-proxy` → the target syncs immediately.

##### Environment for the cluster reproduction

The unit-test route above needs none of this; these are the versions we observed the
problem on.

- flagd-proxy `v0.9.8` — this is **not** fixed by the two defects resolved in that
  release; it was found while verifying those fixes
- open-feature-operator `v0.9.2` (its default proxy image is `v0.9.4`; tag overridden
  to `v0.9.8`)
- Kubernetes, managed cloud offering
- Client: OpenFeature Java SDK 1.22.0 + flagd provider 0.14.0, in-process resolver

yuchou87:
### What

Fixes #2030: a subscription arriving after a target's watcher has stopped attaches to a
multiplexer nothing is watching and never receives data, until flagd-proxy is restarted.

While writing the regression tests I found that the same code path also kills the
process outright, which is covered below and is arguably the more urgent half.

### The wedge

`RegisterSubscription` decided purely on map membership:

```go
sh, ok := s.multiplexers[target]
if !ok {
    s.multiplexers[target] = &multiplexer{...}
    go s.watchResource(target)
} else {
    sh.subs[key] = storedChannels{...}   // attach; nothing restarts a watcher
    ...
}
```

while `watchResource` removed its entry from a goroutine woken by ctx cancellation, so
between the function returning and that goroutine running, the entry is present but dead.

Two details make it easy to land in that window rather than hard:

- **Any** `Sync` error opens it, not just a missing resource.
- The error is broadcast to the subscribers, so the client's reconnect arrives in the
  window that same error just opened.

The `else` branch does attempt a `ReSync`, which would otherwise rescue the subscriber,
but it is guarded by a second membership check inside a goroutine and is skipped once the
delete has landed. Both recovery paths miss.

### The crash

Driving ordinary subscription churn against a resource that does not exist — the real
gRPC handler, coordinator and file sync, no mocks — kills flagd-proxy on `main`, **5 runs
out of 5**:

```
fatal error: concurrent map iteration and map write
  multiplexer.broadcastError  multiplexer.go:24
  Coordinator.watchResource   manager.go:204
```

`subs` is written under `Coordinator.mu` but read under `multiplexer.mu`, so a subscriber
leaving while a broadcast iterates tears the map. This is a `fatal error`, not a
recoverable panic: the process dies with exit code 2. `syncRef` had the same shape —
written with no lock, read under `Coordinator.mu`.

I did not go looking for this; it is what the churn test hit on the first run.

### Change: the wedge itself

- `RegisterSubscription` treats a multiplexer whose watcher context is cancelled as
  absent and rebuilds. This also covers the cleanup loop shutting an idle multiplexer
  down while its watcher is still inside `Sync`.
- `watchResource` removes its entry in a `defer` instead of from a goroutine, and only if
  the entry is still its own — the previous unconditional delete could remove a
  replacement that a later subscription had already built.

### Change: three locking problems in the same file

These are **not** caused by #2030 and are not a refactor I went looking for. All three are
present on `main` and were surfaced by the regression tests, which exercise concurrent
subscribe/unsubscribe against a failing sync for the first time. I have kept them here
rather than splitting them because they cannot be separated from the tests — see the last
point below.

**1. `subs` was guarded by two different locks.** Written under `Coordinator.mu`
(`RegisterSubscription`, and the cleanup goroutine that removes a departing subscriber),
read under `multiplexer.mu` (`broadcastData` / `broadcastError`). So a subscriber leaving
while a broadcast iterates tears the map, which Go turns into a `fatal error` — the crash
shown above, 5 runs out of 5. Writes now take both locks; the ordering is always
`Coordinator.mu` then `multiplexer.mu`, and the broadcasts take only `multiplexer.mu`, so
no cycle exists. The invariant is now recorded on the field.

**2. `syncRef` was written without a lock.** `watchResource` assigned it directly while
`RegisterSubscription` and `FetchAllFlags` read it under `Coordinator.mu`. The write now
takes that lock, and `FetchAllFlags` reads the value while it still holds the read lock
instead of dereferencing after releasing it.

**3. `ReSync` ran while holding `Coordinator.mu`.** `RegisterSubscription`'s `else` branch
held `s.mu.RLock()` across `sh.syncRef.ReSync(...)`. The handler's `dataSync` is
unbuffered and every core `ReSync` implementation ends in an uncancellable send, so a
single stalled subscriber parks that goroutine — and with it the read lock — indefinitely,
jamming the whole coordinator. `syncRef` is now snapshotted under the lock the caller
already holds and `ReSync` runs outside it. For the same reason `watchResource` broadcasts
the sync error *before* taking `Coordinator.mu` in its cleanup: a jammed lock must never
be able to stop an error reaching subscribers.

Problem 3 is the one I would most understand you wanting split out, since it is a liveness
hazard rather than a race. I found it because the first version of this PR put the error
broadcast *behind* that lock and reintroduced a wedge; the fix and its test are in here as
a result.

**Why these ship together with the fix:** `make test` runs `go test -race`, and every one
of the new tests fails on unmodified `main` — two of them by killing the test binary
outright. Landing the tests without these fixes leaves CI red. If you would rather have
them as separate PRs, say so and I will split them; the ordering would have to be locking
first, then the wedge.

### Testing

Six tests. `Test_SyncFlags_churnOnMissingResource` goes through the real gRPC service;
the rest drive the coordinator directly using the mocks already in `manager_test.go`.

Every one of them fails against `main` (`e045237`). Each cell is 5 runs of that test
against unmodified `main` with this branch's tests applied:

| test | plain | `-race` (what `make test` runs) |
|---|---|---|
| `Test_SyncFlags_churnOnMissingResource` | **fatal error 5/5** | fatal 1, race 4 |
| `Test_multiplexerSubsGuardedConsistently` | **fatal error 5/5** | fatal 4, race 1 |
| `Test_RegisterSubscription_afterIdleShutdown` | fails 5/5 | race 5/5 |
| `Test_watchResource_doesNotDeleteReplacement` | fails 5/5 | fails 5/5 |
| `Test_watchResource_broadcastsErrorWhileResyncStalls` | passes | race 5/5 |
| `Test_RegisterSubscription_afterWatcherStopped` | passes | race 5/5 |

"fatal error" is the map crash above: the test binary is killed, not failed.

Two notes on how to read this. The last two only fail under `-race`, so on a plain
`go test` they are documentation rather than detection — `make test` runs `-race`, so CI
catches them either way. And the crash is loud enough that it can mask an assertion
underneath: `afterIdleShutdown` fails on its own assertion in plain mode but only reports
the race under `-race`.

On the fix, all six pass, including `-race -count=2` and `-shuffle=on`.

Reverting any individual change from this branch also breaks a named test, except for the
two lock-liveness changes — broadcasting the error before taking the lock, and keeping
`ReSync` off the lock — which are complementary defences: reverting either alone leaves
the suite green, reverting both makes
`Test_watchResource_broadcastsErrorWhileResyncStalls` fail. Reverting the synchronous
delete additionally breaks four pre-existing tests.

Verified on `e045237`: all three modules build and vet clean; `go test -race -count=2`
and `-shuffle=on` pass for `./flagd-proxy/...`; `golangci-lint` reports nothing new (the
two `SA1019` hits are pre-existing in `handler.go`, untouched here).

### Notes

- The churn test asserts that the process survives, not a hang count. Some subscriptions
  still hang to their deadline for an unrelated pre-existing reason: `broadcastError`
  does a non-blocking send on the handler's unbuffered channel, so an error can be
  dropped if the receiver is not ready at that instant. Out of scope here.
- Still unexplained from the issue: why the window stayed open for minutes in production.
  These tests reproduce the short window; the fix does not depend on that being resolved,
  since it removes the stale entry as a class rather than narrowing the timing.
- A `ReSync` goroutine can still be parked forever if its subscriber departs, because the
  core `ReSync` implementations end in an uncancellable send. That is pre-existing and
  unchanged in kind by this PR — it previously held `Coordinator.mu` while parked, and no
  longer does. Happy to open a separate issue.


flagd-proxy/pkg/service/subscriptions/multiplexer.go
```diff
@@ -11,11 +11,20 @@ import (
...
 // multiplexer distributes updates for a target to all of its subscribers
 type multiplexer struct {
+	// subs is written holding both Coordinator.mu and mu, and read holding either
 	subs       map[interface{}]storedChannels
 	dataSync   chan sourceSync.DataSync
 	cancelFunc context.CancelFunc
 	syncRef    sourceSync.ISync
 	mu         *sync.RWMutex
+	// watcherCtx is the watchResource context, nil until it starts. Guarded by Coordinator.mu.
+	watcherCtx context.Context
```
JamieSinn: would a single depth channel be a better solution for this? you're not using a context for its intended purpose really here.

yuchou87: @JamieSinn Done — it is a channel now. I closed it rather than reading `len()` on a depth-1 buffer, since the marking happens outside `Coordinator.mu` and `-race` does not instrument `len`; say the word if you would rather have the literal version.

`kill()` also nil-checks `cancelFunc` — that is outside #2030, it fixes a reachable panic in `cleanup`. Happy to pull it out.

flagd-proxy/pkg/service/subscriptions/manager.go
```diff
@@ -127,56 +137,70 @@ func (s *Coordinator) RegisterSubscription(
...
 		sh.subs[key] = storedChannels{
 			errChan:  errChan,
 			dataSync: dataSync,
 		}
-		// access pointer + trigger resync passing the dataSync
-		if sh.syncRef != nil {
+		sh.mu.Unlock()
+		// read syncRef under the lock we hold, but never run ReSync under it: a stalled
+		// subscriber blocks it on an unbuffered send and would jam the coordinator
+		if syncRef := sh.syncRef; syncRef != nil {
```
JamieSinn, line 149: you don't have the lock you state; you're unlocking right before this. Do you mean to do a read lock?

yuchou87: > you don't have the lock you state; you're unlocking right before this. Do you mean to do a read lock?

Coordinator.mu — held from the top of the function; the Unlock just above is the multiplexer's own mu, which guards subs. A read lock on Coordinator.mu would not work, since the branch above inserts into s.multiplexers.

Comment fixed — the lock note now sits on the field in multiplexer.go, next to the ones for subs and done.

erka (changes requested):
Hey @yuchou87.

Thanks for raising this issue and for opening the PR.

flagd-proxy/pkg/service/subscriptions/manager.go
```diff
@@ -127,56 +137,69 @@ func (s *Coordinator) RegisterSubscription(
...
 	go func() {
 		<-ctx.Done()
 		s.mu.Lock()
 		defer s.mu.Unlock()
-		if s.multiplexers[target] != nil && s.multiplexers[target].subs != nil {
+		if sh := registered; sh != nil && sh.subs != nil {
 			s.logger.Debug(fmt.Sprintf("removing sync subscription due to context cancellation %p", key))
-			delete(s.multiplexers[target].subs, key)
+			sh.mu.Lock()
+			delete(sh.subs, key)
```
erka, line 168: Could we introduce the methods for this in mutliplexer?

flagd-proxy/pkg/service/subscriptions/manager.go
```diff
@@ -127,56 +137,69 @@ func (s *Coordinator) RegisterSubscription(
...
 		go s.watchResource(target)
 	} else {
-		// register our sub in the map
+		// register our sub in the map; subs is also read by the broadcasts under sh.mu
 		s.logger.Debug(fmt.Sprintf("registering sync subscription %p", key))
+		sh.mu.Lock()
 		sh.subs[key] = storedChannels{
 			errChan:  errChan,
 			dataSync: dataSync,
 		}
```
erka, line 146: could we introduce method for this in multiplexer? This will improve the reading of the code because `sh.mu.lock` and `s.mu.lock` are so similar

flagd-proxy/pkg/service/subscriptions/multiplexer.go
```diff
@@ -11,11 +11,40 @@ import (
...
 type multiplexer struct {
+	// subs is written holding both Coordinator.mu and mu, and read holding either
 	subs       map[interface{}]storedChannels
 	dataSync   chan sourceSync.DataSync
 	cancelFunc context.CancelFunc
-	syncRef    sourceSync.ISync
-	mu         *sync.RWMutex
+	// syncRef is written by watchResource and read by the resync paths, all under Coordinator.mu
+	syncRef sourceSync.ISync
+	mu      *sync.RWMutex
```
erka, line 20:
```suggestion
	mu      sync.RWMutex
```

flagd-proxy/pkg/service/subscriptions/multiplexer.go
```diff
@@ -11,11 +11,40 @@ import (
...
+	// done is closed by kill once the watcher is cancelled. Nil until the watcher starts.
+	// Written once by watchResource under Coordinator.mu; readers hold it too, except
+	// watchResource's own defer, which is the writer.
+	done    chan struct{}
+	dieOnce sync.Once
```
erka, line 25: Ideally, I would like not to add this. The manager should keep tracks and remove unused multiplexers as soon as possible from its collection. 

Please modify this to `done bool` and use mutex for changing and reading value.

flagd-proxy/pkg/service/subscriptions/manager.go
```diff
@@ -222,11 +248,11 @@ func (s *Coordinator) cleanup() {
 		case <-time.After(5 * time.Second):
 			s.mu.Lock()
 			for k, v := range s.multiplexers {
-				// delete any multiplexers with 0 active subscriptions through cancelling its context
+				// reap any multiplexer with 0 active subscriptions; kill, never a bare cancel (#2030)
 				s.logger.Debug(fmt.Sprintf("multiplexer for target %s has %d subscriptions", k, len(v.subs)))
 				if len(v.subs) == 0 {
```
erka, line 253: `len(v.subs)` could race as v.subs could be modified with additions or removals. We probably should move this to a method in multiplexer.

flagd-proxy/pkg/service/subscriptions/manager.go
erka, line 247: should s.multiplexers be stopped here as well?

flagd-proxy/pkg/service/subscriptions/stale_multiplexer_test.go
```diff
@@ -0,0 +1,514 @@
...
+	return dataChan
+}
+
+// Test_RegisterSubscription_afterWatcherStopped covers #2030 end to end: a client that
+// reconnects after its sync failed must receive data again. It usually gets there via the
+// removed map entry rather than isDead, so it does not by itself prove the dead-multiplexer
+// path; Test_RegisterSubscription_whileStoppingWatcherBroadcasts pins that one down.
+func Test_RegisterSubscription_afterWatcherStopped(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
erka, line 112:
```suggestion
	ctx := t.Context()
```

flagd-proxy/pkg/service/subscriptions/stale_multiplexer_test.go
```diff
@@ -0,0 +1,514 @@
...
+		for {
+			select {
+			case <-done:
+				return
+			case syncSrc.dataSyncChanIn <- isync.DataSync{FlagData: "update"}:
+			}
+		}
+	}()
+
+	for i := 0; i < 50; i++ {
```
erka, line 185:
```suggestion
	pushers.Go(func() {
		for {
			select {
			case <-done:
				return
			case syncSrc.dataSyncChanIn <- isync.DataSync{FlagData: "update"}:
			}
		}
	})

	for i := range 50 {
```

flagd-proxy/pkg/service/subscriptions/multiplexer.go
yuchou87: Done — `done` is a bool under `mu` now. `cancelFunc` moved under `mu` too, which the bool needs: a nil channel used to be how "this watcher has not started yet" was expressed, and without an equivalent `cleanup` would mark an unstarted multiplexer dead, `RegisterSubscription` would replace the map entry, and the still-queued `watchResource` would adopt the replacement and orphan its watcher. `kill()` now leaves a multiplexer with no cancel for a later tick.

On removing unused multiplexers as soon as possible — I would rather do that as its own change than fold it in here, since it rewrites the premise of `Test_RegisterSubscription_afterCleanupLoopCancelled`. Happy to open an issue.

flagd-proxy/pkg/service/subscriptions/manager.go
yuchou87: Not needed — each watcher's context derives from `s.ctx`, so cancelling it unwinds every `watchResource`, and their defers remove their own entries. Measured: the map drains to zero about a millisecond after the coordinator context is cancelled. Happy to make it explicit if you would rather not rely on that.

yuchou87:
All in `d4e01ec` — methods on the multiplexer, `mu` as a value, `done` as a bool under it, `t.Context()`, `WaitGroup.Go`. `manager.go` no longer mentions the multiplexer's lock at all, which was the point of the ask.

One change you did not ask for: `watchResource`'s defer now does the whole teardown — kill, unmap, broadcast — under one hold of `Coordinator.mu`, so a `RegisterSubscription` cannot land in between. Defence in depth rather than a second fix, but it retires an ordering that was awkward to test; `Test_watchResource_teardownIsAtomic` pins it.

Two pre-existing things I would rather file separately than grow this PR: `FetchAllFlags` skips the `isDead` check `RegisterSubscription` does, and `handler.go`'s subscriber `errChan` is unbuffered against a non-blocking `broadcastError`, which accounts for most of the residual hangs the churn test reports here. Tell me if you would rather either came in.

flagd-proxy/pkg/service/subscriptions/manager.go
```diff
@@ -222,11 +236,11 @@ func (s *Coordinator) cleanup() {
...
 			for k, v := range s.multiplexers {
-				// delete any multiplexers with 0 active subscriptions through cancelling its context
-				s.logger.Debug(fmt.Sprintf("multiplexer for target %s has %d subscriptions", k, len(v.subs)))
-				if len(v.subs) == 0 {
+				// reap any multiplexer with 0 active subscriptions; kill, never a bare cancel (#2030)
+				s.logger.Debug(fmt.Sprintf("multiplexer for target %s has %d subscriptions", k, v.subCount()))
+				if v.subCount() == 0 {
 					s.logger.Debug(fmt.Sprintf("shutting down multiplexer %s", k))
-					s.multiplexers[k].cancelFunc()
+					s.multiplexers[k].kill()
```
erka: We acquire the lock three time here and it isn't an atomic operation in general. If we could address it here it would be great.

flagd-proxy/pkg/service/churn_test.go
```diff
@@ -0,0 +1,106 @@
...
+	defer cancel()
+
+	flagFile := filepath.Join(t.TempDir(), "flags.json")
+	target := "file:" + flagFile
+
+	port := freePort(t)
+	log := logger.NewLogger(nil, false)
+	s := NewServer(ctx, log, subscriptions.NewManager(ctx, log))
+	s.config = service.Configuration{Port: port, ManagementPort: freePort(t), ReadinessProbe: func() bool { return true }}
+	go func() { _ = s.startServer() }()
```
erka: if we start it we need to stop it as well. Please handle the error, it will be helpful to debug possible issues.

flagd-proxy/pkg/service/churn_test.go
```diff
@@ -0,0 +1,106 @@
...
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				start := time.Now()
+				cfg, err := attempt()
+				results <- err != nil && cfg == "" && time.Since(start) > 1500*time.Millisecond
+			}()
+		}
+		wg.Wait()
+	}
```
erka: it would be nice to modernize it with `for round := range 20`, `for c := ...` and `wg.Go`

flagd-proxy/pkg/service/churn_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+// Test_SyncFlags_churnOnMissingResource drives the real gRPC handler, coordinator and
+// file sync with no mocks. Subscribing to a resource that does not exist makes every
+// watcher fail and return, and clients retry into that window. On an unguarded subs map
+// this kills the process with "fatal error: concurrent map iteration and map write" (#2030).
+func Test_SyncFlags_churnOnMissingResource(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
erka:
```suggestion
	ctx := t.Context()
```

erka (approved):
Thanks @yuchou87 

There are few nits which would be nice to address in this PR.

flagd-proxy/pkg/service/subscriptions/manager.go
yuchou87: Down to one acquisition per entry: `killIfIdle` reads the count and marks under the same hold, and returns the count it saw so the log line does not need a second read. The unstarted-watcher guard both paths need is a single `markDeadLocked`, rather than the one `kill` already had plus a second copy inside `killIfIdle`.

Three tests came out of it. One pins the count — a multiplexer with subscribers keeps its cancel and its flag, which nothing covered before, here or in the version this replaces. The other two cover `killIfIdle`'s way into the shared guard and its ordering on a successful kill; `kill` had tests for both on its own path, but a mutation reached through `killIfIdle` alone survived them.

flagd-proxy/pkg/service/churn_test.go
yuchou87: Done — it shuts down in `t.Cleanup` and reports anything other than `grpc.ErrServerStopped`, and the loops are `for range 20` with `wg.Go`. `t.Context()` too.

One thing worth flagging while I was in there: `startServer` returns before it assigns `grpcServer`, so if the listen fails `Shutdown` would nil-deref: the bind error does get printed, and then a SIGSEGV trace lands on top of it and takes the rest of the package with it. The cleanup checks whether the server already exited first, and there is a second check right after `waitForListener`, because `freePort` only reserves its port briefly and the listener we reach may belong to whoever took it from us.

flagd-proxy/pkg/service/churn_test.go
yuchou87: Done — `for range 20` in both loops with `wg.Go`; same 400 attempts and the same per-round barrier.

flagd-proxy/pkg/service/churn_test.go
yuchou87: Done.

yuchou87:
Four things I ran into while working on this that are on `main` already, none of which I have touched here. Happy to open issues or PRs for any of them — say which, if any, you would rather see.

**`Server.Shutdown` dereferences `grpcServer` unconditionally** (`flagd-proxy/pkg/service/server.go`). `startServer` assigns it only after `net.Listen` succeeds, and `cmd/start.go` defers `Shutdown` and then selects between `ctx.Done()` and the error channel — the error branch calls `logger.Fatal` and skips the defer, but the `ctx.Done()` branch returns into it, so a signal arriving before the listener is up panics during shutdown. A nil check fixes it.

**The same field is read and written without synchronisation** — assigned in `startServer` after the listener is up, read in `Shutdown`, which callers run from a different goroutine than the one serving. Nothing orders the two, so it is a race by the memory model even though `-race` will not usually surface it: in a real run seconds separate the accesses.

**A subscriber can lose the error that tells it to reconnect.** `handler.go` builds `errChan` unbuffered and `broadcastError` sends non-blocking, so if the handler goroutine is not sitting in its `select` at that instant the error is logged server-side and dropped. That is most of the residual hang count the churn test in this PR reports. The counts are load-dependent — on my machine three runs of 400 reconnects gave 5, 13 and 5 hanging to their own deadline — but the effect of buffering that channel to one is not subtle: the hangs go to zero or near it, and the test's wall time collapses from seconds to milliseconds.

**`FetchAllFlags` does not apply the `isDead` check `RegisterSubscription` does.** It takes the `ReSync` path on a dead-but-mapped multiplexer instead of rebuilding. Bounded in practice, and for a duller reason than I first thought: the file sync discards the context it is handed, and the grpc and kubernetes syncs hold clients built in `Init` that outlive the cancelled watcher, so the fetch generally just succeeds. The cost is that the dead entry is worked around rather than rebuilt, until some `RegisterSubscription` replaces it — an asymmetry between two entry points in the same file rather than a wedge.

flagd-proxy/pkg/service/churn_test.go
erka: It could be as simple as `if err := s.startServer(); err != nil { t.Error....}` but your version also works.

flagd-proxy/pkg/service/subscriptions/stale_multiplexer_test.go
yuchou87: Done — and the other five in this file, so there is no hand-rolled cancel left in it.

JamieSinn (approved):

PR #2036, opened 2026-08-26
chore: OFREP ETag response based on full response body.

JamieSinn:
## This PR
The current ETag response from ofrep is based on the config version/context targeting key. There's a whole bunch of other attributes that are possible to change in the OFREP response that could invalidate the configuration and need to be taken into account of the response here.

This allows namely, the SSE URL to update between requests even if no other response values change. Without this, the old connection would be maintained and not updated/refreshes.

I wrote the etag handling as a middleware as this should be discrete from any actual body writing/response editing. The middleware should act discrete from the actual response and be a mutator to the headers/response code alone.

### DIFF
--- a/flagd/pkg/service/flag-evaluation/ofrep/etag.go
+++ b/flagd/pkg/service/flag-evaluation/ofrep/etag.go
@@ -107,16 +107,10 @@ func (rec *responseRecorder) flush() {
 	}
 }
 
-// ifNoneMatch reports whether any of the client's If-None-Match field values selects the
-// representation tagged with etag. Per RFC 9110 §13.1.2 the field is either "*" or a list of
-// entity tags, repeated fields extend that list, and the comparison is weak.
-//
-// net/http parses this too, but scanETag and checkIfNoneMatch are unexported, and the one
-// exported way in, http.ServeContent, answers a failed precondition on a non-GET method with 412
-// rather than the 304 OFREP asks for on this POST route.
-//
-// The scan is more forgiving than net/http's, which abandons the whole field on the first entry
-// that is not a syntactically valid entity tag.
+// ifNoneMatch reports whether any If-None-Match value selects the representation tagged with etag.
+// Per RFC 9110 13.1.2 the field is "*" or a weakly-compared list of entity tags. 
+// Hand-rolled because net/http's parser is unexported and its only public path (ServeContent) answers 412,
+// not the 304 OFREP wants on this POST route.
 func ifNoneMatch(fields []string, etag string) bool {
 	for _, field := range fields {
 		for _, candidate := range splitETagList(field) {
--- a/flagd/pkg/service/flag-evaluation/ofrep/etag.go
+++ b/flagd/pkg/service/flag-evaluation/ofrep/etag.go
@@ -25,7 +25,7 @@ func conditionalETag(next http.Handler) http.Handler {
 		etag := bodyETag(rec.body.Bytes())
 		w.Header().Set("ETag", quoteETag(etag))
 
-		if clientETag := r.Header.Get("If-None-Match"); clientETag != "" && normalizeETag(clientETag) == etag {
+		if ifNoneMatch(r.Header.Values("If-None-Match"), etag) {
 			w.Header().Del("Content-Type")
 			w.Header().Del("Content-Length")
 			w.WriteHeader(http.StatusNotModified)
@@ -71,9 +71,67 @@ func bodyETag(body []byte) string {
 	return hex.EncodeToString(sum[:])
 }
 
-// normalizeETag lets quoted and unquoted forms compare equal.
+// ifNoneMatch reports whether any of the client's If-None-Match field values selects the
+// representation tagged with etag. Per RFC 9110 §13.1.2 the field is either "*" or a list of
+// entity tags, repeated fields extend that list, and the comparison is weak.
+//
+// net/http parses this too, but scanETag and checkIfNoneMatch are unexported, and the one
+// exported way in, http.ServeContent, answers a failed precondition on a non-GET method with 412
+// rather than the 304 OFREP asks for on this POST route.
+//
+// The scan is more forgiving than net/http's, which abandons the whole field on the first entry
+// that is not a syntactically valid entity tag. A malformed entry is compared as an opaque token
+// instead, so neither it nor an unquoted tag hides a valid tag listed beside it. Erring this way
+// only ever costs a client the bandwidth of a body it already holds.
+func ifNoneMatch(fields []string, etag string) bool {
+	for _, field := range fields {
+		for _, candidate := range splitETagList(field) {
+			if candidate == "*" || normalizeETag(candidate) == etag {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// splitETagList splits a list of entity tags on its commas. A comma inside a quoted tag belongs
+// to the tag, so the split tracks quoting instead of reaching for strings.Split.
+func splitETagList(list string) []string {
+	var (
+		tags    []string
+		current strings.Builder
+		quoted  bool
+	)
+
+	flush := func() {
+		if tag := strings.TrimSpace(current.String()); tag != "" {
+			tags = append(tags, tag)
+		}
+		current.Reset()
+	}
+
+	for i := 0; i < len(list); i++ {
+		switch c := list[i]; {
+		case c == '"':
+			quoted = !quoted
+			current.WriteByte(c)
+		case c == ',' && !quoted:
+			flush()
+		default:
+			current.WriteByte(c)
+		}
+	}
+	flush()
+
+	return tags
+}
+
+// normalizeETag reduces an entity tag to the opaque value the weak comparison function looks at:
+// neither the quotes nor a weakness prefix carries meaning there. Dropping the quotes also lets
+// the unquoted form a lenient client may send compare equal.
 func normalizeETag(etag string) string {
-	return strings.Trim(etag, `"`)
+	return strings.Trim(strings.TrimPrefix(etag, "W/"), `"`)
 }
 
 func quoteETag(etag string) string {
--- a/flagd/pkg/service/flag-evaluation/ofrep/etag.go
+++ b/flagd/pkg/service/flag-evaluation/ofrep/etag.go
@@ -1,0 +1,81 @@
+package ofrep
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"net/http"
+	"strings"
+)
+
+// conditionalETag tags a 200 with a strong ETag over the bytes the handler wrote, and answers 304
+// when the client's If-None-Match matches. Hashing the response rather than the flag configuration
+// keeps the validator correct for context-dependent values; the trade is that producing the tag
+// needs the response, so a 304 saves bandwidth, not work.
+func conditionalETag(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		rec := &responseRecorder{ResponseWriter: w, status: http.StatusOK}
+		next.ServeHTTP(rec, r)
+
+		if rec.status != http.StatusOK {
+			rec.flush()
+			return
+		}
+
+		etag := bodyETag(rec.body.Bytes())
+		w.Header().Set("ETag", quoteETag(etag))
+
+		if clientETag := r.Header.Get("If-None-Match"); clientETag != "" && normalizeETag(clientETag) == etag {
+			w.Header().Del("Content-Type")
+			w.Header().Del("Content-Length")
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+
+		rec.flush()
+	})
+}
+
+type responseRecorder struct {
+	http.ResponseWriter
+
+	status  int
+	body    bytes.Buffer
+	written bool
+}
+
+func (rec *responseRecorder) WriteHeader(status int) {
+	if rec.written {
+		return
+	}
+	rec.status = status
+	rec.written = true
+}
+
+func (rec *responseRecorder) Write(b []byte) (int, error) {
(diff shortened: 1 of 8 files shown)


flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,81 @@
...
+
+		if rec.status != http.StatusOK {
+			rec.flush()
+			return
+		}
+
+		etag := bodyETag(rec.body.Bytes())
+		w.Header().Set("ETag", quoteETag(etag))
+
+		if clientETag := r.Header.Get("If-None-Match"); clientETag != "" && normalizeETag(clientETag) == etag {
```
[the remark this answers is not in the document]
JamieSinn: We do not implement sending multiple entity tags as part of our response generation. This is not implemented from the etag generation.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,139 @@
...
+		current.Reset()
+	}
+
+	for i := 0; i < len(list); i++ {
+		switch c := list[i]; {
+		case c == '"':
+			quoted = !quoted
+			current.WriteByte(c)
+		case c == ',' && !quoted:
+			flush()
```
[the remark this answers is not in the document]
JamieSinn: this is a nit/edge case.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
```diff
@@ -172,12 +172,8 @@ func (h *handler) HandleBulkEvaluation(w http.ResponseWriter, r *http.Request) {
...
 	ctx := context.WithValue(r.Context(), store.SelectorContextKey{}, selector)
 
-	// Conditional evaluation (ADR-0008): short-circuit with 304 when the client already holds
-	// the current config version.
-	lastModified, notModified := h.applyConditionalETag(w, r, selectorExpression)
-	if notModified {
-		w.WriteHeader(http.StatusNotModified)
-		return
+	if trigger := r.URL.Query().Get(flagConfigEtagParam); trigger != "" {
+		h.Logger.Debug(fmt.Sprintf("bulk refetch triggered by %s=%s", flagConfigEtagParam, trigger))
```
erka: structured logging pls

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,81 @@
...
+		rec.WriteHeader(http.StatusOK)
+	}
+	return rec.body.Write(b)
+}
+
+func (rec *responseRecorder) flush() {
+	rec.ResponseWriter.WriteHeader(rec.status)
+	if rec.body.Len() > 0 {
+		_, _ = rec.ResponseWriter.Write(rec.body.Bytes())
+	}
```
erka, line 108: `io.Copy` could be a better option.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,81 @@
...
+	rec.status = status
+	rec.written = true
+}
+
+func (rec *responseRecorder) Write(b []byte) (int, error) {
+	if !rec.written {
+		rec.WriteHeader(http.StatusOK)
+	}
+	return rec.body.Write(b)
```
erka, line 97: It's possible to calculate hash as the bytes are written `return io.MultiWriter(&r.body, r.hash).Write(p)` plus ```responseRecorder{
	ResponseWriter: w,
	hash:            sha256.New(),...}```

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+func bodyETag(body []byte) string {
+	sum := sha256.Sum256(body)
+	return hex.EncodeToString(sum[:])
+}
+
+// normalizeETag lets quoted and unquoted forms compare equal.
+func normalizeETag(etag string) string {
```
erka: I would suggest to remove this func. When generating etag we could apply `quoteETag` and then just compare without extra memory allocations. 
```go
		etag := quoteETag(bodyETag(rec.body.Bytes()))
		w.Header().Set("ETag", etag)
```

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,81 @@
...
+
+		rec.flush()
+	})
+}
+
+type responseRecorder struct {
+	http.ResponseWriter
+
+	status  int
+	body    bytes.Buffer
```
erka: Should we use `sync.Pool` for it? It's very common to reuse `bytes.Buffer` instead of creating more work for garbage collector.

erka (commented):
Great work @JamieSinn!

Some nits from me.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: Coderabbit pointed (correctly) out that we may not control the etag throughout its lifecycle of the request path, and we need to ensure that it's in a consistent format. 

I thought it's a good catch, but also a bit extra to ensure/enforce a given format. 

Thoughts?

flagd/pkg/service/flag-evaluation/ofrep/handler.go
JamieSinn: yep this was a test log left in - my bad - I wasn't even sure if I wanted this log in the first place.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: interesting idea, I wasn't really thinking about the reuse pattern here but it makes a lot more sense given that the recorder parent could be kept active and just allocating only the growth between bodies.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: does that not double allocate the destinations bytes? I guess we already do that when calculating the hash and it's just a bit more clean.

JamieSinn: using a multiwriter from testing actually caused more allocs as a result. keeping the discrete writes was faster and cheaper apparently.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: refactored

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: apparently at runtime it distills down to this. `io.Copy` proxies to `WriteTo` which then calls `Write`

flagd/pkg/service/flag-evaluation/ofrep/etag.go
erka: This is a tough one. The RFC says that `*` could be used only with unsafe methods. Per 13.1.2 the safe method are HEAD and GET and only those could return Not Modified status. In all other cases should respond with 412 Precondition Failed.

In some way the OFREP spec with POST endpoints conflicts with the  RFC for ETag. I see OFREP POST as a safe method so we should not allow `"*"` and you are right about weak comparing. 

I still prefer etag generation with quotes and compare it to incoming values.

flagd/pkg/service/flag-evaluation/ofrep/handler.go
erka: This is a common issue in flagd. If log level is higher than debug the strings are generated for nothing.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
erka: I see some different results locally. Probably it  should be more researched and done in another PR if it makes sense.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: Sounds good - I can update to always quote and then we can skip normalizing.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: yeah - agreed

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: @erka you might be interested in this too - https://github.com/open-feature/protocol/pull/84

flagd/pkg/service/flag-evaluation/ofrep/handler.go
JamieSinn: removed

toddbaert:
I like the middleware approach. Good idea.

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,168 @@
...
+// ifNoneMatch reports whether any of the client's If-None-Match field values selects the
+// representation tagged with etag. Per RFC 9110 §13.1.2 the field is either "*" or a list of
+// entity tags, repeated fields extend that list, and the comparison is weak.
+//
+// net/http parses this too, but scanETag and checkIfNoneMatch are unexported, and the one
+// exported way in, http.ServeContent, answers a failed precondition on a non-GET method with 412
+// rather than the 304 OFREP asks for on this POST route.
+//
+// The scan is more forgiving than net/http's, which abandons the whole field on the first entry
+// that is not a syntactically valid entity tag.
```
toddbaert:
```suggestion
// ifNoneMatch reports whether any If-None-Match value selects the representation tagged with etag.
// Per RFC 9110 13.1.2 the field is "*" or a weakly-compared list of entity tags. 
// Hand-rolled because net/http's parser is unexported and its only public path (ServeContent) answers 412,
// not the 304 OFREP wants on this POST route.
```

toddbaert: Seems too verbose to me.

toddbaert (approved):
One nit: https://github.com/open-feature/flagd/pull/2036/changes#r3882646566

flagd/pkg/service/flag-evaluation/ofrep/etag.go
JamieSinn: yeah I rewrote this a few times - I like your rewrite much better

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,162 @@
...
+// bodyBuffers recycles response buffers: consecutive bulk evaluations are close in size, so a
+// recycled buffer already holds the capacity the next one needs.
+var bodyBuffers = sync.Pool{
+	New: func() any { return new(bytes.Buffer) },
+}
+
+// getBodyBuffer lends out an empty buffer.
+func getBodyBuffer() *bytes.Buffer {
+	buf := bodyBuffers.Get().(*bytes.Buffer)
+	buf.Reset()
```
erka: It would be nice to move `buf.Reset()` into `putBodyBuffer`

flagd/pkg/service/flag-evaluation/ofrep/etag.go
```diff
@@ -0,0 +1,162 @@
...
+		case c == ',' && !quoted:
+			flush()
+		default:
+			current.WriteByte(c)
+		}
+	}
+	flush()
+
+	return tags
+}
```
erka:
```suggestion
func splitETagList(list string) iter.Seq[string] {
	return func(yield func(string) bool) {
		var (
			start  int
			quoted bool
		)
		flush := func(end int) bool {
			if tag := strings.TrimSpace(list[start:end]); tag != "" {
				if !yield(tag) {
					return false
				}
			}
			return true
		}
		for i := 0; i < len(list); i++ {
			switch c := list[i]; {
			case c == '"':
				quoted = !quoted
			case c == ',' && !quoted:
				if !flush(i) {
					return
				}
				start = i + 1
			}
		}
		flush(len(list))
	}
}
```

Bench

```go
var benchInput = `"abc123", "def456", W/"ghi789", "jkl012", "mno345, still quoted", "pqr678"`

func BenchmarkSplitETagListOld(b *testing.B) {
	b.ReportAllocs()
	for b.Loop() {
		_ = splitETagListOld(benchInput)
	}
}

func BenchmarkSplitETagListNew_Range(b *testing.B) {
	b.ReportAllocs()
	for b.Loop() {
		for tag := range splitETagList(benchInput) {
			_ = tag
		}
	}
}
```

```sh
BenchmarkSplitETagListOld-8         	  538238	      2284 ns/op	     400 B/op	      16 allocs/op
BenchmarkSplitETagListNew_Range-8   	 2130223	       524.2 ns/op	      48 B/op	       3 allocs/op
```

erka (approved):
It looks good. One nit from me.

An observation not directly related to this work. If server is restarted the new etag is generated for the same unmodified file and browser gets 200. I think this is a side effect of 

```go		sub.ver.CompareAndSwap(prev, &version{configEtag: fp, lastModified: time.Now().Unix()})
```
I don't know if it's possible to get the file last modified time or something similar (for example, s3 LastModified object attribute)

JamieSinn:
> I don't know if it's possible to get the file last modified time or something similar (for example, s3 LastModified object attribute)

This was something I was debating on - but the implementation of such and the plumbing of the "source"'s LM/Etag ref I thought was a bit too much of a change scope on an already quite large PR.
I definitely agree though; having the underlying source value be the authoritative value is 100% something I want to improve on in the future. That reduces/removes the entire complexity of the timestamping.

PR #2037, opened 2026-08-27
feat: HTTP fetch equivalent of the gRPC FetchAllFlags

JamieSinn:
## This PR

HTTP was being a bit of a forgotten middle child here. This is to bring a bit more parity to HTTP vs gRPC.

This implements the same watcher/listener as the gRPC setup to prevent any issues and also handles for the ETag/LastModified handling so that this can be a valid downstream or upstream to other flagd instances.

A key slight difference is that this will return 404, not 400 on an invalid flagset selector. (eg. `potato=2`) specifically because of HTTP semantics around the route accessor `/v1/flags/potato%3D2`, nothing is especially incorrect in a user perspective from the route accessor, but instead the path is just "not found". 

A 400 is returned when there is an invalid charset/character set (eg - invalid UTF8 characters) as that is genuinely a "you did something wrong" situation. It will return a 200 when a valid selector string returns an empty flagset, along with an empty flags object.


docs/reference/grpc-sync-service.md
```diff
@@ -5,7 +5,8 @@ description: flagd as a gRPC sync service
 # Overview
 
 flagd can expose a gRPC sync service, allowing in-process providers to obtain their flag definitions.
-The gRPC sync stream contains flag definitions currently configured at flagd as [sync-configurations](./sync-configuration.md).
+The gRPC sync stream contains flag definitions currently configured at flagd
```
aepfli: we usually keep documentation in markdown on purpose on one line, so you can easily move the whole sentence. could you please revert the formatting changes.

erka:
I would like to understand why we need an additional net listener and port. Per flagd docs, we do this with connectrpc

```sh
curl -X POST \
-d '{"flagKey":"myBoolFlag","context":{}}' \
-H "Content-Type: application/json" \
"http://localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" 
```

What prevents us from using the same approach for the sync service?

docs/reference/grpc-sync-service.md
JamieSinn: yeah - this was one of the auto-formatter things that went through. Happy to coalesce to a one line.

JamieSinn:
> I would like to understand why we need an additional net listener and port. Per flagd docs, we do this with connectrpc
> 
> ```shell
> curl -X POST \
> -d '{"flagKey":"myBoolFlag","context":{}}' \
> -H "Content-Type: application/json" \
> "http://localhost:8013/flagd.evaluation.v1.Service/ResolveBoolean" 
> ```
> 
> What prevents us from using the same approach for the sync service?

Conceptually you want the resolution service to be disparate from any service that provides the full configuration to prevent inadvertently exposing your full configuration. This was from a conversation with @toddbaert when we were discussing this functionality. 

Leaking a full configuration for a flagset is much more damaging than leaking the result of a resolved value. 

Creating a new port/listener here is a way to prevent any existing deployments that are passing the port to something like nginx/etc do not then inadvertently expose their configuration publicly.

erka:
The feature of connectRPC over plain GRPC is that it could serve both grpc and http requests on the same port. And `flagd.evaluation.v1.Service` does it on `8013` port. 

I haven't researched the internals of flagd very well, but from my blackbox dummy tries I see that `flagd.sync.v1.FlagSyncService` on `8015` only responses to grpc requests. Is there any reason we don't want to support http requests on `8015`? @JamieSinn

JamieSinn:
Personally no objections on the shared port, as long as it doesn't share the same port as the the OFREP endpoint, or another endpoint that's not already enabling the full config fetch. 

The refactor to use the shared port doesn't seem too horrible either. I'll see how that looks

toddbaert:
> The feature of connectRPC over plain GRPC is that it could serve both grpc and http requests on the same port. And `flagd.evaluation.v1.Service` does it on `8013` port.
> 
> I haven't researched the internals of flagd very well, but from my blackbox dummy tries I see that `flagd.sync.v1.FlagSyncService` on `8015` only responses to grpc requests. Is there any reason we don't want to support http requests on `8015`? @JamieSinn

> Personally no objections on the shared port, as long as it doesn't share the same port as the the OFREP endpoint, or another endpoint that's not already enabling the full config fetch.
> 
> The refactor to use the shared port doesn't seem too horrible either. I'll see how that looks

@JamieSinn @erka I agree with @erka here... but for me it's a "nice to have" if the code isn't crazy, and there's no downside, I'm in favor of putting both on this port since they have the same exact security exposure.

docs/reference/grpc-sync-service.md
```diff
@@ -21,15 +22,19 @@ erDiagram
...
+standard merge strategy.
 You can read more about the merge strategy in our dedicated [concepts guide on syncs](../concepts/syncs.md).
 
-If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector value to a source, and stream just the flags identified in that source.
+If you specify a `selector` in the gRPC sync request, the gRPC service will attempt match the provided selector value to
+a source, and stream just the flags identified in that source.
 For example, if `selector` is set to `myFlags.json`, service will stream flags observed from `myFlags.json` file.
-Note that, to observe flags from `myFlags.json` file, you may use startup option `uri` like `--uri myFlags.json` or `source` option `--sources='[{"uri":"myFlags.json", provider":"file"}]`.
+Note that, to observe flags from `myFlags.json` file, you may use startup option `uri` like `--uri myFlags.json` or
+`source` option `--sources='[{"uri":"myFlags.json", provider":"file"}]`.
```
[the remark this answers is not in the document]
JamieSinn: you missed the second single quote too escaping the double quoted values inside.

JamieSinn:
Shared port it is! Merged the stacked pr into this branch for easier reviewing

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// get reports the zero time until the first snapshot lands, so ServeContent omits Last-Modified.
+func (m *modTime) get() time.Time {
+	sec := m.unix.Load()
+	if sec == 0 {
+		return time.Time{}
+	}
+	return time.Unix(sec, 0)
+}
```
erka: Just thoughts...

I think the store should track this as etag or version and return with get/getAll calls. Otherwise, there is always race and 1 of n calls will get stale evaluation or version results.

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,120 @@
...
+	flags, _, err := h.store.GetAll(r.Context(), &selector)
+	if err != nil {
+		h.log.Error(fmt.Sprintf("error retrieving flags from store: %v", err))
+		http.Error(w, "error retrieving flags from store", http.StatusInternalServerError)
+		return
+	}
+
+	body, err := generateResponse(flags)
+	if err != nil {
+		h.log.Error(fmt.Sprintf("error marshalling flags: %v", err))
```
erka:
```suggestion
		h.log.Error("error marshaling flags", zap.Error(err))
```

marshalling or marshaling? en_GB vs en_US

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,120 @@
...
+		return
+	}
+
+	// Sampled before the store read: a later stamp could outrun this body, stranding the client on
+	// stale flags behind a fresh Last-Modified.
+	lastModified := h.modTime.get()
+
+	flags, _, err := h.store.GetAll(r.Context(), &selector)
+	if err != nil {
+		h.log.Error(fmt.Sprintf("error retrieving flags from store: %v", err))
```
erka:
```suggestion
		h.log.Error("error retrieving flags from store", zap.Error(err))
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+			caPath:   "./test-cert/ca-cert.pem",
+			scheme:   "https",
+			// the test cert carries a single SAN of IP:0.0.0.0
+			host: "0.0.0.0",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.title, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion
			ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+			assert.Contains(t, string(body), "flagA")
+			assert.Contains(t, string(body), "flagB")
+		})
+	}
+}
+
+// The listener is bound eagerly so the TCP connect succeeds; it is the HTTP exchange that must not
+// complete until every source has reported in.
+func TestHTTPService_WaitsForInitialSync(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion	
	ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	emit()
+
+	resp := getWithRetry(t, &http.Client{Timeout: 2 * time.Second}, url)
+	defer resp.Body.Close()
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+}
+
+func TestHTTPService_ShutsDownWithContext(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+	require.NoError(t, err)
+
+	assert.Nil(t, svc.httpServer)
+	assert.Nil(t, svc.httpListener)
+	assert.NotNil(t, svc.grpcListener)
+}
+
+// Last-Modified is only advertised once a sync source has actually reported in.
+func TestHTTPService_ModTimeStampedOnEmit(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+	svc, err := NewSyncService(SvcConfigurations{
+		Logger:      logger.NewLogger(zap.New(core), false),
+		Port:        18030,
+		Sources:     sources,
+		Store:       flagStore,
+		HTTPEnabled: true,
+	})
+	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/http_service_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	time.Sleep(500 * time.Millisecond)
+	require.Equal(t, 1, logs.FilterMessageSnippet("timeout while waiting for all sync sources").Len(),
+		"initial-sync timeout warned more than once")
+}
+
+// The mux sends every h2 connection to gRPC, so ALPN ordering is what keeps the config endpoint
+// reachable over TLS: an h2-capable client must be steered to http/1.1 rather than into gRPC.
+func TestHTTPService_TLSNegotiatesHTTP11(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
```
erka:
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
```

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -173,11 +231,35 @@ func (s *Service) Start(ctx context.Context) error {
...
 }
 
 func (s *Service) shutdown() {
 	s.logger.Info("shutting down gRPC sync service")
 	s.server.Stop()
+
+	if s.httpServer != nil {
+		s.logger.Info("shutting down http sync service")
+		if err := s.httpServer.Close(); err != nil {
+			s.logger.Warn(fmt.Sprintf("error from sync http server shutdown: %v", err))
```
erka:
```suggestion
			s.logger.Warn("error from sync http server shutdown", zap.Error(err))
```

erka (commented):
I think we don't need `github.com/soheilhy/cmux`.  

We should focus on `buf.build/gen/go/open-feature/flagd/connectrpc/go/flagd/sync/v1/syncv1connect` and use it similar to the evaluation service.

flagd/pkg/service/flag-sync/http_handler.go
JamieSinn: ugh this is one of the more annoying "features" of my Dynatrace laptop. It auto uses en-US for any and all autocorrect/spell check. 

Will fix.

JamieSinn:
> I think we don't need `github.com/soheilhy/cmux`.
> 
> We should focus on `buf.build/gen/go/open-feature/flagd/connectrpc/go/flagd/sync/v1/syncv1connect` and use it similar to the evaluation service.

This would be a massive refactor. the cmux setup is/was the least impactful change I could reasonably setup to do this on the shared port. There's a whole ton of refactoring that needs to be done. 

i'd say that should be scoped to a different PR and it can be evaluated later as a result.

flagd/pkg/service/flag-sync/http_handler.go
JamieSinn: I would love it if we could source the config etag version from the source/store, but depending on the sync source that's not a guarantee, but would realistically be a net new PR from this one.

For the hyperscaler stores, they natively support both, but not every other source contains this info. HTTP/URI sourcing could plausibly have both.

Realistically outside the scope to do that, but I'm definitely open to other ideas on how to make this a bit more tolerant to race conditions. Maybe just extending the store and having this property tied to that?

toddbaert:
> > I think we don't need `github.com/soheilhy/cmux`.
> > We should focus on `buf.build/gen/go/open-feature/flagd/connectrpc/go/flagd/sync/v1/syncv1connect` and use it similar to the evaluation service.
> 
> This would be a massive refactor. the cmux setup is/was the least impactful change I could reasonably setup to do this on the shared port. There's a whole ton of refactoring that needs to be done.
> 
> i'd say that should be scoped to a different PR and it can be evaluated later as a result.

Ya I agree with this - I _think_ it should be possible but we'd be switching the implementation essentially I think there'd be some possible side effects, like the way errors are handled is slightly different in `connect` IIRC, and the code would be significantly different. I agree it's a better option overall, but I would make a new issue for it.

EDIT:

I guess the problem is if we decide to do this later, it will break the implementation.

JamieSinn:
> Ya I agree with this - I _think_ it should be possible but we'd be switching the implementation essentially I think there'd be some possible side effects, like the way errors are handled is slightly different in `connect` IIRC, and the code would be significantly different. I agree it's a better option overall, but I would make a new issue for it.
> 
> EDIT:
> 
> I guess the problem is if we decide to do this later, it will break the implementation.

I think the big question is more around whether or not this refactor should have already been done before now because regardless of when it happens, it will be a breaking change. 

Thoughts on that @erka ?

erka:
> I think the big question is more around whether or not this refactor should have already been done before now because regardless of when it happens, it will be a breaking change.
> 
> Thoughts on that @erka ?

Honestly, it's hard to give advice here, as I don't know how critical this is for the business. In an ideal world, I’d prefer not to introduce a known breaking change or future workarounds. But we don't live in an ideal world, so there needs to be a balance. If it needs to be shipped, I'm fine with it.

toddbaert:
@erka 

@JamieSinn and I were doing some research, and we think you are right that the best thing to do is implement this with connect-go. Connect has a specific guarantee:

> "you do not need to modify your service's clients: they can continue to use their current gRPC clients. Your current Protobuf schema will also work without modification."
"Handlers support the gRPC protocol by default: they work with grpc-go, grpcurl, and any other gRPC client using TLS without any special configuration."

So, basically connect should not break anything, and should give us what we want, without a new dep.

My recommendation is that @JamieSinn opens a new PR targeting this branch which does purely a connect migration without any other changes so we can review that separately, and then either merge that into this, or merge this into main and then fast-follow with the connect migration.

> Honestly, it's hard to give advice here, as I don't know how critical this is for the business. 

Internally we need an HTTP endpoint for getting all flags, and it's something I've heard requested from others as well at Kubecons for some basic CI integration and checks (you can imaging that calling an HTTP endpoint is much easier than gRPC from some automation).

But - again - I think my recommendation is not to implement this as is, but instead use connect, and @JamieSinn is working on that now - we'll just do it in a separate PR to keep the changes unmixed. That will bring flagd into further consistency anyway since the evaluation.proto implementation is also based on connect.

Thanks again @erka , it was a good callout.

toddbaert:
We should probably add a `Vary` header entry on this if the selector is used in a header to prevent caching that ignores it, IIUC.

JamieSinn:
@toddbaert / @erka merged #2042 into this branch; feel free to re-review

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,92 @@
...
+	http.ServeContent(w, r, contentName, lastModified, bytes.NewReader(body))
+}
+
+// writeError maps a fetch failure onto a status. No branch echoes the expression back: it is
+// unescaped client input.
+func (h httpHandler) writeError(w http.ResponseWriter, err error) {
+	var fetchErr fetchError
+	if !errors.As(err, &fetchErr) {
+		fetchErr = fetchError{kind: fetchStoreRead, cause: err}
+	}
```
erka:
```suggestion
	fetchErr, ok := errors.AsType[fetchError](err)
	if !ok {
		fetchErr = fetchError{kind: fetchStoreRead, cause: err}
	}
```

flagd/pkg/service/flag-sync/http_handler_test.go
```diff
@@ -0,0 +1,319 @@
...
+
+	grpcHandler := syncHandler{
+		store:           flagStore,
+		log:             logger.NewLogger(nil, false),
+		metricsRecorder: &telemetry.NoopMetricsRecorder{},
+	}
+
+	for _, selector := range []string{"", "source=" + testSource1, "source=" + testSource2} {
+		t.Run("selector="+selector, func(t *testing.T) {
+			grpcResp, err := grpcHandler.FetchAllFlags(context.Background(),
```
erka:
```suggestion
			grpcResp, err := grpcHandler.FetchAllFlags(t.Context(),
```

flagd/pkg/service/flag-sync/keepalive_test.go
```diff
@@ -0,0 +1,252 @@
...
+		})
+	}
+}
+
+// The other side of #1998: an idle stream must not be reaped either.
+func TestSyncService_StreamOutlivesReadTimeout(t *testing.T) {
+	const port = 18051
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
```
erka: I believe we could just use `ctx := t.Context()` here. There are tests with similar setup. Probably it was not modernized with recent additions to the go standard test library.

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -28,105 +27,103 @@ type ISyncService interface {
...
+	var lis net.Listener
+	if cfg.SocketPath != "" {
+		l.Info(fmt.Sprintf("starting flag sync service at %s", cfg.SocketPath))
+		lis, err = net.Listen("unix", cfg.SocketPath)
+	} else {
+		l.Info(fmt.Sprintf("starting flag sync service on port %d", cfg.Port))
+		lis, err = net.Listen("tcp", fmt.Sprintf(":%d", cfg.Port))
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error creating listener: %w", err)
```
erka: could we move this down after the cert condition (to the order it was originally where listener was created at the last step) ? I don't want to worry about that `lis.Close()` if there is an issue with tls certs.

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -171,13 +167,42 @@ func (s *Service) Start(ctx context.Context) error {
...
+	}
 }
 
 func (s *Service) shutdown() {
-	s.logger.Info("shutting down gRPC sync service")
-	s.server.Stop()
+	s.logger.Info("shutting down flag sync service")
+
+	// Close, not Shutdown: a long-lived SyncFlags stream would hold a graceful shutdown to its timeout.
+	if err := s.server.Close(); err != nil {
```
erka: I would like to see `Shutdown` with some context.Timeout(with 5 or 10 seconds) so the non-stream requests would have a chance to finish without interruption.

flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -171,13 +167,42 @@ func (s *Service) Start(ctx context.Context) error {
...
+		s.logger.Warn("error from sync server shutdown", zap.Error(err))
+	}
+
+	// Serve may never have started, in which case http.Server does not know the listener.
+	s.listener.Close()
```
erka: It would be nice to log this error if any. We may use debug level but I am sure that there will be a case when people would like see this error in future.

erka (approved):
nice. few nits.

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,92 @@
...
+	fetchErr, ok := errors.AsType[fetchError](err)
+	if !ok {
+		fetchErr = fetchError{kind: fetchStoreRead, cause: err}
+	}
+
+	switch fetchErr.kind {
+	case fetchSelectorMalformed:
+		http.Error(w, "malformed selector", http.StatusBadRequest)
+	case fetchSelectorInvalid:
+		http.Error(w, "no such selector", http.StatusNotFound)
```
toddbaert: This behavior is inconsistent. In all other cases, a selector that returns nothing simply filters the result-set, it doesn't return a 404.

Also, `fetchSelectorInvalid` doesn't seem to be to be a good name for `no such selector` - but I guess it should be removed entirely.

flagd/pkg/service/flag-sync/http_handler.go
```diff
@@ -0,0 +1,92 @@
...
+	// Sampled before the read: a later stamp would strand the client on stale flags.
+	lastModified := h.modTime.get()
+
+	body, err := fetchAllFlags(r.Context(), h.store, resolveSelector(r.Header, ""))
+	if err != nil {
+		h.writeError(w, err)
+		return
+	}
+
+	w.Header().Set("ETag", etagOf(body))
```
toddbaert, line 61:
```suggestion

	w.Header().Set("ETag", etagOf(body))
	w.Header().Set("Vary", flagdService.FLAGD_SELECTOR_HEADER)
```

This should prevent caching across different selectors which we do not want.

toddbaert: Caching by proxies, I mean.

flagd/pkg/service/flag-sync/http_handler.go
JamieSinn: removed

toddbaert (approved):

PR #2038, opened 2026-08-28
chore: Use shared port with gRPC for HTTP (stacked)

JamieSinn:
@erka this is the diff for sharing the gRPC setup. Thoughts?


flagd/pkg/service/flag-sync/sync_service.go
```diff
@@ -138,34 +127,43 @@ func NewSyncService(cfg SvcConfigurations) (*Service, error) {
...
+	// Split the one listener by protocol rather than serving gRPC through http.Handler, so gRPC
+	// keeps grpc-go's own HTTP/2 server and its keepalive enforcement.
+	svc.mux = cmux.New(lis)
+	if cfg.HTTPEnabled {
+		// Split on the HTTP/2 preface. Both content-type matchers are
+		// unusable here: the SendSettings one writes a SETTINGS frame while probing and corrupts
+		// non-gRPC h2 connections, and the read-only one waits for HEADERS that grpc-go will not
+		// send until it has read a SETTINGS frame, deadlocking every gRPC client.
+		svc.grpcListener = svc.mux.Match(cmux.HTTP2())
+		svc.httpListener = svc.mux.Match(cmux.Any())
```
JamieSinn, line 159: @erka this was the largest specific change for this. We now have to do special handling to allow http2 to properly terminate at the right endpoint. Not sure I'm in love with this; especially because it's an added layer of complexity.

PR #2045, opened 2026-09-09
feat!: make file poll interval configurable

leakonvalinka:
## This PR

- makes the poll interval for files configurable instead of the previously hard-coded 1s

### Related Issues

In scope of https://github.com/open-feature/go-sdk-contrib/issues/936 - needed in order to introduce `offlinePollIntervalMs` in the go-sdk-contrib flagd provider (as it wraps flagd core)

### DIFF
--- a/core/pkg/sync/file/fileinfo_watcher.go
+++ b/core/pkg/sync/file/fileinfo_watcher.go
@@ -38,7 +38,7 @@ type fileInfoWatcher struct {
 const defaultFileInfoPollIntervalMs = 1000
 
 // NewFileInfoWatcher returns a new fileInfoWatcher
-func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger, pollIntervalMs int) Watcher {
+func NewFileInfoWatcher(ctx context.Context, pollIntervalMs int, logger *logger.Logger) Watcher {
 	if pollIntervalMs <= 0 {
 		pollIntervalMs = defaultFileInfoPollIntervalMs
 	}
--- a/core/pkg/sync/file/fileinfo_watcher.go
+++ b/core/pkg/sync/file/fileinfo_watcher.go
@@ -35,8 +35,13 @@ type fileInfoWatcher struct {
 	wg sync.WaitGroup
 }
 
+const defaultFileInfoPollIntervalMs = 1000
+
 // NewFileInfoWatcher returns a new fileInfoWatcher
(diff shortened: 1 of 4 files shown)


core/pkg/sync/file/fileinfo_watcher.go
```diff
@@ -45,7 +50,7 @@ func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger) Watcher {
 		watches:  make(map[string]fs.FileInfo),
 		done:     make(chan struct{}),
 	}
-	fiw.run(ctx, (1 * time.Second))
+	fiw.run(ctx, time.Duration(pollIntervalMs)*time.Millisecond)
```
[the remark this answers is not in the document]
leakonvalinka: Technically true but practically only relevant if someone provides a duration in ms equivalent to ~290 years, so I don't think we necessarily need this

JamieSinn (approved):

core/pkg/sync/file/fileinfo_watcher.go
```diff
@@ -35,8 +35,13 @@ type fileInfoWatcher struct {
 	wg sync.WaitGroup
 }
 
+const defaultFileInfoPollIntervalMs = 1000
+
 // NewFileInfoWatcher returns a new fileInfoWatcher
-func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger) Watcher {
+func NewFileInfoWatcher(ctx context.Context, logger *logger.Logger, pollIntervalMs int) Watcher {
```
jsalaber: small nit: logger and the polling param are in different orders for `NewFileInfoWatcher` and `NewFileSync`, could be cleaner if they both have `logger` at the end for example. Not blocking, feel free to resolve this comment

toddbaert: Agree, if we are making breaking changes (we are here) we should fix that anyway.

toddbaert (approved):
LGTM, but I agree with @jsalaber on [this](https://github.com/open-feature/flagd/pull/2045#discussion_r3973871566), and I this is also a breaking change (fine since this is an internal lib, but I added `!` to the title).

Cleared the FOSSA alert - it was on an unshipped python dev dep.

core/pkg/sync/file/fileinfo_watcher.go
leakonvalinka: Thank you guys! Will update

PR #2050, opened 2026-09-11
fix: wait for flag state before reporting ready

### LINKED ISSUE #2047: `/readyz` reports ready before flags are evaluable
## Summary

flagd's readiness probe returns 200 while the flag store is still empty, so an evaluation made immediately after `/readyz` first succeeds gets `FLAG_NOT_FOUND` for a flag the configuration plainly defines. Measured at **100% of starts** with a realistic flag set, with a window of ~14ms (valid 30KB payload) to ~51ms (30KB payload containing entries that fail schema validation).

To be precise about what is and is not being claimed: the [monitoring docs](https://flagd.dev/reference/monitoring/) say 200 means "all sync providers at least have one successful data sync", and the code **literally satisfies that** — the sync provider really has fetched and delivered the data. So this is not code violating its documented contract. The claim is that those semantics are the wrong ones for a readiness probe, because "delivered" is not "applied".

## Mechanism

`core/pkg/sync/file/filepath_sync.go`:

```go
func (fs *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
	...
	fs.sendDataSync(ctx, dataSync)   // only pushes onto the channel
	fs.setReady(true)                // IsReady() is now true -> /readyz 200
```

`sendDataSync` ends in `dataSync <- sync.DataSync{...}`, and in `flagd/pkg/runtime/runtime.go` that channel is

```go
dataSync := make(chan sync.DataSync, len(r.Syncs))
```

buffered to exactly the number of sources. So every source can hand its payload to the buffer and call `setReady(true)` without a single payload having been received. `Runtime.isReady()` only asks each sync `IsReady()`, so `/readyz` flips to 200 at that point.

What actually makes a flag evaluable happens afterwards, on the separate goroutine draining the channel:

```go
case data := <-dataSync:
	r.updateAndEmit(data)     // -> r.Evaluator.SetState(payload) : parse, validate, swap store
```

The window is therefore the cost of parsing and validating the payload, which is why it scales with flag-set size and is invisible in a trivial test fixture.

## Reproduction

No provider and no test harness involved — plain flagd, one file source, tight-polling `/readyz` and evaluating over OFREP the instant it returns 200.

`flags.json` — 201 trivial, schema-valid flags (~30KB), generated with:

```python
import json
flags = {"flag-%03d" % i: {"state": "ENABLED",
                           "variants": {"on": True, "off": False},
                           "defaultVariant": "on"} for i in range(200)}
flags["boolean-flag"] = {"state": "ENABLED",
                         "variants": {"on": True, "off": False},
                         "defaultVariant": "on"}
json.dump({"flags": flags}, open("flags.json", "w"), indent=2)
```

```bash
docker run -d --name flagd-repro -p 8014:8014 -p 8016:8016 \
  -v "$PWD/flags.json:/flags.json:ro" \
  ghcr.io/open-feature/flagd:v0.16.0 start --uri file:/flags.json

# poll /readyz in a tight loop until 200, then immediately:
curl -s -X POST -H 'Content-Type: application/json' -d '{}' \
  http://localhost:8016/ofrep/v1/evaluate/flags/boolean-flag
# {"key":"boolean-flag","errorCode":"FLAG_NOT_FOUND","errorDetails":"flag `boolean-flag` does not exist","metadata":{}}
```

The `curl` above has to win a ~14ms race against a process start, so it is unreliable by hand. This runs the whole thing in a loop and reports a rate:

<details>
<summary>flagd_repro.go — stdlib only, <code>go run flagd_repro.go 15</code></summary>

```go
// Minimal reproduction: flagd's /readyz reports 200 before flags are evaluable.
package main

import (
	"bytes"
	"fmt"
	"io"
	"net/http"
	"os"
	"os/exec"
	"sort"
	"strconv"
	"time"
)

const (
	image     = "ghcr.io/open-feature/flagd:v0.16.0"
	container = "flagd-readyz-repro"
	flagKey   = "boolean-flag"

	readyzURL = "http://localhost:8014/readyz"
	ofrepURL  = "http://localhost:8016/ofrep/v1/evaluate/flags/" + flagKey
)

func main() {
	runs := 20
	if len(os.Args) > 1 {
		runs, _ = strconv.Atoi(os.Args[1])
	}

	client := &http.Client{Timeout: 2 * time.Second}
	var windows []time.Duration
	raced := 0

	for i := 1; i <= runs; i++ {
		stopContainer()
		if err := startContainer(); err != nil {
			fmt.Printf("%-4d could not start flagd: %v\n", i, err)
			continue
		}

		// Tight-poll /readyz so we observe the 200 the instant it happens.
		for {
			resp, err := client.Get(readyzURL)
			if err == nil {
				resp.Body.Close()
				if resp.StatusCode == http.StatusOK {
					break
				}
			}
		}
		ready := time.Now()

		notFound := 0
		var window time.Duration
		for {
			body, err := evaluate(client)
			elapsed := time.Since(ready)
			if err == nil && !bytes.Contains(body, []byte("FLAG_NOT_FOUND")) {
				window = elapsed
				break
			}
			notFound++
			if elapsed > 10*time.Second {
				window = elapsed
				break
			}
		}

		status := "ok"
		if notFound > 0 {
			status = "RACED"
			raced++
			windows = append(windows, window)
		}
		fmt.Printf("%-4d %-6s flag_not_found_responses=%-5d evaluable_after=%v\n", i, status, notFound, window)
	}
	stopContainer()

	fmt.Printf("\nflagd %s\n", image)
	fmt.Printf("runs=%d  runs_where_/readyz_200_preceded_evaluability=%d (%.0f%%)\n",
		runs, raced, 100*float64(raced)/float64(runs))
	if len(windows) > 0 {
		sort.Slice(windows, func(a, b int) bool { return windows[a] < windows[b] })
		fmt.Printf("window: min=%v median=%v max=%v\n",
			windows[0], windows[len(windows)/2], windows[len(windows)-1])
	}
}

func evaluate(c *http.Client) ([]byte, error) {
	resp, err := c.Post(ofrepURL, "application/json", bytes.NewReader([]byte("{}")))
	if err != nil {
		return nil, err
	}
	defer resp.Body.Close()
	return io.ReadAll(resp.Body)
}

func startContainer() error {
	wd, _ := os.Getwd()
	cmd := exec.Command("docker", "run", "-d", "--name", container,
		"-p", "8014:8014", "-p", "8016:8016",
		"-v", wd+"/flags.json:/flags.json:ro",
		image, "start", "--uri", "file:/flags.json")
	out, err := cmd.CombinedOutput()
	if err != nil {
		return fmt.Errorf("%v: %s", err, out)
	}
	return nil
}

func stopContainer() {
	_ = exec.Command("docker", "rm", "-f", container).Run()
}
```

</details>

### Results — flagd v0.16.0, OFREP on 8016, Docker on Linux

| payload | runs | runs where `/readyz` 200 preceded evaluability | window (min / median / max) |
| --- | --- | --- | --- |
| 1 flag, ~200 bytes | 20 | 0 (0%) | not observable |
| 201 valid flags, 30KB | 15 | **15 (100%)** | 11.9ms / 13.8ms / 23.0ms |
| flagd-testbed's own flag set, 30KB, contains entries that fail schema validation | 20 | **20 (100%)** | 43.5ms / 51.4ms / 85.5ms |

The single-flag row is the control: the race is structural, but the window is sub-millisecond when there is nothing to parse — which is why a minimal fixture will not show it, and probably why this has gone unnoticed.

It is not protocol-specific. The flag evaluation port (8013) races identically; OFREP is just the easiest way to observe it.

## Why it matters

**Kubernetes.** `/readyz` on the management port is the documented readiness probe. A pod therefore passes readiness and starts taking traffic before it can resolve a flag, and every request in that window gets `FLAG_NOT_FOUND` — which providers surface as the code default, silently. On a rollout that is a burst of wrong-but-plausible flag values rather than an error anyone notices. The window grows with flag-set size, so the larger the deployment, the worse it gets.

**Conformance testing.** This is how it was found. The OpenFeature provider conformance suite (open-feature/spec#417) drives flagd-testbed, whose `/start` waited on `/readyz`. Providers that block during their own initialisation absorb the window and never see it; a **stateless** provider such as OFREP evaluates the instant the backend reports ready, and fails essentially every scenario — which reads as a catastrophically broken provider rather than as a racing backend. Worked around on the testbed side in open-feature/flagd-testbed#394, but that fix is a workaround for this.

## Suggested direction

Gate readiness on the store having been populated rather than on the payload having been handed over — e.g. have the runtime mark readiness once `updateAndEmit` has applied at least one payload per source, rather than deriving it from `sync.IsReady()` alone.

If the current semantics are intentional, then the documentation is the thing to change: it should say plainly that 200 does not mean flags are servable, and that `/readyz` is consequently not suitable as a Kubernetes readiness probe. I would expect most users to read the current wording as the opposite.

## Version

- flagd `v0.16.0` (`ghcr.io/open-feature/flagd:v0.16.0`, built 2026-06-01)
- Mechanism also present on `main` at `6fdf338`

akashchamp:
## Summary

Make readiness wait until every configured sync source has successfully updated the evaluator. A sync provider can report that it is watching before the runtime has drained and applied its first payload; `/readyz` now stays unavailable during that gap and after a failed initial `SetState`.

Readiness is tracked per configured source URI. Once every source has applied state, an `initialized` flag is set and readiness only depends on the providers' watch state from then on.

The regression tests cover two configured sources, an initial evaluator error, and the runtime forwarding path, so a single applied payload cannot make readiness pass.

Fixes #2047

## Testing

- `go test -race -count=1 ./flagd/pkg/runtime/...`
- `go test -race -covermode=atomic -cover -short ./flagd/pkg/...`
- `go build ./...` and `go vet ./pkg/runtime/...` in `flagd/`, `gofmt -l` clean
- Manual runtime verification with a freshly built binary and the sample file source on isolated ports: across three startup cycles, I polled `/readyz` and immediately evaluated `myBoolFlag` through OFREP. Each cycle returned HTTP 200 with `value: true`, with no `FLAG_NOT_FOUND` response.

### DIFF
--- a/flagd/pkg/runtime/from_config.go
+++ b/flagd/pkg/runtime/from_config.go
@@ -117,6 +117,7 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
 	for _, provider := range config.SyncProviders {
 		sources = append(sources, provider.URI)
 	}
+	sourceReadiness := make([]bool, len(config.SyncProviders))
 
 	// build flag store, collect flag sources & fill sources details
 	store, err := store.NewStore(logger, sources)
@@ -214,7 +215,8 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
 			MaxRequestBodyBytes:        config.MaxRequestBodyBytes,
 			MaxRequestHeaderBytes:      config.MaxRequestHeaderBytes,
 		},
-		Syncs: iSyncs,
+		Syncs:           iSyncs,
+		sourceReadiness: sourceReadiness,
 	}, nil
 }
 
--- a/flagd/pkg/runtime/runtime.go
+++ b/flagd/pkg/runtime/runtime.go
@@ -27,7 +27,15 @@ type Runtime struct {
 	ServiceConfig     service.Configuration
 	Syncs             []sync.ISync
 
+	// sourceReadiness records whether each configured sync provider has successfully updated the evaluator.
+	sourceReadiness []bool
+
 	mu msync.Mutex
+}
+
+type providerDataSync struct {
+	payload       sync.DataSync
+	providerIndex int
 }
 
 //nolint:funlen
@@ -44,13 +52,13 @@ func (r *Runtime) Start() error {
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer cancel()
 	g, gCtx := errgroup.WithContext(ctx)
-	dataSync := make(chan sync.DataSync, len(r.Syncs))
+	dataSync := make(chan providerDataSync, len(r.Syncs))
 	// Initialize DataSync channel watcher
 	g.Go(func() error {
 		for {
 			select {
 			case data := <-dataSync:
-				r.updateAndEmit(data)
+				r.updateAndEmit(data.payload, data.providerIndex)
 			case <-gCtx.Done():
 				return nil
 			}
@@ -62,11 +70,30 @@ func (r *Runtime) Start() error {
 			return fmt.Errorf("sync provider Init returned error: %w", err)
 		}
 	}
-	// Start sync provider
-	for _, s := range r.Syncs {
-		p := s
+	// Start sync providers. Each provider has its own channel so the runtime can preserve
+	// provider identity even when two configured providers share a source URI.
+	for index, s := range r.Syncs {
+		p, providerIndex := s, index
+		providerData := make(chan sync.DataSync, 1)
 		g.Go(func() error {
-			if err := p.Sync(gCtx, dataSync); err != nil {
+			for {
+				select {
+				case data, ok := <-providerData:
+					if !ok {
+						return nil
+					}
+					select {
+					case dataSync <- providerDataSync{payload: data, providerIndex: providerIndex}:
+					case <-gCtx.Done():
+						return nil
+					}
+				case <-gCtx.Done():
+					return nil
+				}
+			}
+		})
+		g.Go(func() error {
+			if err := p.Sync(gCtx, providerData); err != nil {
 				return fmt.Errorf("sync provider returned error: %w", err)
 			}
 			return nil
@@ -119,11 +146,19 @@ func (r *Runtime) isReady() bool {
 			return false
 		}
 	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for _, ready := range r.sourceReadiness {
+		if !ready {
+			return false
+		}
+	}
 	return true
 }
 
 // updateAndEmit helps to update state, notify changes and trigger sync updates
-func (r *Runtime) updateAndEmit(payload sync.DataSync) {
+func (r *Runtime) updateAndEmit(payload sync.DataSync, providerIndex int) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
@@ -131,6 +166,10 @@ func (r *Runtime) updateAndEmit(payload sync.DataSync) {
 	if err != nil {
 		r.Logger.Error(fmt.Sprintf("error setting state: %v", err))
 		return
+	}
+
+	if providerIndex < len(r.sourceReadiness) {
+		r.sourceReadiness[providerIndex] = true
 	}
 	r.SyncService.Emit(payload.Source)
 }
--- a/flagd/pkg/runtime/runtime_test.go
+++ b/flagd/pkg/runtime/runtime_test.go
@@ -1,0 +1,211 @@
+package runtime
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	evalmock "github.com/open-feature/flagd/core/pkg/evaluator/mock"
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/service"
+	coresync "github.com/open-feature/flagd/core/pkg/sync"
+	flagsync "github.com/open-feature/flagd/flagd/pkg/service/flag-sync"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+type readySync struct {
+	ready bool
+}
+
+func (s readySync) Init(context.Context) error {
+	return nil
+}
+
+func (s readySync) Sync(context.Context, chan<- coresync.DataSync) error {
+	return nil
+}
+
+func (s readySync) ReSync(context.Context, chan<- coresync.DataSync) error {
+	return nil
+}
+
+func (s readySync) IsReady() bool {
+	return s.ready
+}
+
+type recordingSyncService struct {
+	emitted []string
+}
+
+var _ flagsync.ISyncService = (*recordingSyncService)(nil)
+
+func (s *recordingSyncService) Start(context.Context) error {
+	return nil
+}
+
+func (s *recordingSyncService) Emit(source string) {
+	s.emitted = append(s.emitted, source)
+}
+
+type controlledSync struct {
+	payloads <-chan coresync.DataSync
+	started  chan<- struct{}
+}
+
+func (s controlledSync) Init(context.Context) error {
+	return nil
+}
+
+func (s controlledSync) Sync(ctx context.Context, dataSync chan<- coresync.DataSync) error {
+	select {
+	case s.started <- struct{}{}:
+	case <-ctx.Done():
+		return nil
+	}
+
+	for {
+		select {
+		case payload := <-s.payloads:
+			select {
+			case dataSync <- payload:
+			case <-ctx.Done():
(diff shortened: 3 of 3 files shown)


flagd/pkg/runtime/runtime.go
```diff
@@ -27,9 +27,17 @@ type Runtime struct {
 	ServiceConfig     service.Configuration
 	Syncs             []sync.ISync
 
+	// sourceReadiness records whether each configured sync provider has successfully updated the evaluator.
+	sourceReadiness []bool
+
```
erka, line 32:
```suggestion
	initialized     atomic.Bool
```

flagd/pkg/runtime/runtime.go
```diff
@@ -119,11 +146,19 @@ func (r *Runtime) isReady() bool {
...
 		}
 	}
+
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for _, ready := range r.sourceReadiness {
+		if !ready {
+			return false
+		}
+	}
```
erka, line 156:
```suggestion
func (r *Runtime) isReady() bool {
	if r.initialized.Load() {
		// if all providers can watch for flag changes, we are ready.
		for _, p := range r.Syncs {
			if !p.IsReady() {
				return false
			}
		}
		return true
	}
	return false
}
````

File: flagd/pkg/runtime/runtime.go [hunk 3]
```diff
@@ -132,5 +167,9 @@ func (r *Runtime) updateAndEmit(payload sync.DataSync) {
 		r.Logger.Error(fmt.Sprintf("error setting state: %v", err))
 		return
 	}
+
+	if providerIndex < len(r.sourceReadiness) {
+		r.sourceReadiness[providerIndex] = true
+	}
```
erka, line 173:
```suggestion
	if !r.initialized.Load() && providerIndex < len(r.sourceReadiness) {
		r.sourceReadiness[providerIndex] = true

		allReady := false
		for _, ready := range r.sourceReadiness {
			allReady = allReady && ready
		}
		if allReady {
			r.initialized.Store(true)
		}
	}
```

flagd/pkg/runtime/runtime.go
```diff
@@ -62,11 +70,30 @@ func (r *Runtime) Start() error {
...
+					select {
+					case dataSync <- providerDataSync{payload: data, providerIndex: providerIndex}:
+					case <-gCtx.Done():
+						return nil
+					}
+				case <-gCtx.Done():
+					return nil
+				}
+			}
+		})
```
erka, line 94: Personally, I prefer the previous version. I don't think that the same source will cause any issue.

erka (commented):
Hey @akashchamp 

Thank you for taking care of this!

flagd/pkg/runtime/from_config.go
```diff
@@ -117,6 +117,10 @@ func FromConfig(logger *logger.Logger, version string, config Config) (*Runtime,
 	for _, provider := range config.SyncProviders {
 		sources = append(sources, provider.URI)
 	}
+	sourceReadiness := make(map[string]bool, len(sources))
+	for _, source := range sources {
+		sourceReadiness[source] = false
+	}
```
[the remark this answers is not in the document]
akashchamp: I went back to keying readiness by source URI here, since erka preferred the previous version and a repeated URI is not a problem in practice.

flagd/pkg/runtime/runtime.go
akashchamp: Added the initialized atomic.Bool next to the readiness map, and isReady and updateAndEmit now use it as you suggested.

flagd/pkg/runtime/runtime.go
akashchamp: Applied as suggested: isReady now only consults the providers' IsReady once initialized is set, so it no longer needs the mutex.

flagd/pkg/runtime/runtime.go
akashchamp: Took this over, adapted to the source-keyed map since I reverted the index version. One small change: allReady starts as true, because starting from false the loop could never become true and initialized would never be set.

flagd/pkg/runtime/runtime.go
akashchamp: Agreed, I went back to the previous version: Start uses the shared channel again and readiness is keyed by source URI. That should also clear the Sonar cognitive-complexity finding on Start.

erka (approved):

akashchamp:
@erka hi, could you please merge this pr, if nothing is left ?

erka:
> could you please merge this pr, if nothing is left ? 

Hey @akashchamp 

I'm only a reviewer in flagd repo. The maintainers will take care of their review and merge activity.

toddbaert (approved):
Thanks!

